PR #1, opened 2015-03-24
Removed parameters.yml and sqlite database files from repo

norzechowicz:


javiereguiluz:
@norzechowicz thanks for your pull request. I was expecting a PR like this, but I didn't expect it that soon!

Sadly I must close this PR without merging it. The reason? This demo application must work out-of-the-box without configuring anything or executing any command. That's why we link compiled CSS/JS instead of using Assetic, we provide a SQLite database with the sample data instead of forcing the user to create the database and load fixtures, etc.

This application is for the absolute Symfony beginner, so everything must work right without doing anything. I hope you understand it :)

norzechowicz:
Yea I see the point of keeping the database file but not the parameters.yml, even absolute Symfony beginner needs to install dependencies and this will trigger generating parameters.yml  but correct me if I'm wrong.

javiereguiluz:
@norzechowicz when installing this application via the `symfony demo` command, the installer will download a ZIP/TGZ package with all the vendors and everything included. Composer won't even be necessary to download, install and test the demo application.

michalpipa:
Why not configure database in `config.yml`? You don't have to store database connection parameters in `parameters.yml`.

It is a bad practice to keep `parameters.yml` in VCS repository.

norzechowicz:
It's also bad practice to keep whole database in repository.

javiereguiluz:
@michalpipa you are right (and @norzechowicz) is also right. It's a good practice to not put `parameters.yml` in the repository. That's [what we explain](http://symfony.com/doc/current/best_practices/configuration.html) in the Best Practices book.

But this application is a very special piece of software. It's designed as a learning tool for the absolute beginner. What we want is that you execute the "symfony demo" command in the Symfony Installer ... and 10 seconds later you have downloaded and installed a full Symfony application that works without executing any command or doing any change.

weaverryan:
Hi guys!

I totally agree with Javier :). But what we *can* do is add comments in the code to help guide the beginners as they're discovering things.

@javiereguiluz I do have one idea I'd like your thoughts on. What if we didn't commit `parameters.yml` nor the database. But we *did* have a `parameters.yml.dist` (we already do) and a `app/data/blog.sqlite.dist`. We could then have 1 very simple "prep" step that copied both of these to `parameters.yml` and `blog.sqlite` respectively right before creating the tar. I want to keep things simpler, but this might be a nice way to keep things very very easy, but include more examples of good practices.

javiereguiluz:
@weaverryan I propose you something: let me finish the `symfony demo` command and all the other things to make this a reality. And once we test the actual command and the experience, let's see how we can improve it, or how much can we change it without degrading it. We won't have to wait long, because everything should be finished very soon.

weaverryan:
:+1: Love that we'll be able to iterate and improve the demo very easily. So yea, no rush to do this

PR #2, opened 2015-03-24
Removed form->isSubmitted() checks

mickaelandrieu:
$form->isValid() already check if a form has been validated or not.

So I think it's not required to do this check (see: https://github.com/symfony/Form/blob/master/Form.php#L769)

### DIFF
--- a/src/AppBundle/Controller/Admin/BlogController.php
+++ b/src/AppBundle/Controller/Admin/BlogController.php
@@ -64,7 +64,7 @@ class BlogController extends Controller
 
         $form->handleRequest($request);
 
-        if ($form->isSubmitted() && $form->isValid()) {
+        if ($form->isValid()) {
             $post->setSlug($this->get('slugger')->slugify($post->getTitle()));
 
             $em = $this->getDoctrine()->getManager();
@@ -115,7 +115,7 @@ class BlogController extends Controller
 
         $editForm->handleRequest($request);
 
-        if ($editForm->isSubmitted() && $editForm->isValid()) {
+        if ($editForm->isValid()) {
             $post->setSlug($this->get('slugger')->slugify($post->getTitle()));
             $em->flush();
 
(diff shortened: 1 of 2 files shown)


weaverryan:
@mickaelandrieu You're right, but we actually recommend that people add the `isSubmitted()`. The reason is that `isValid()` doesn't *look* like it checked if the form was submitted. So if you're not familiar with Symfony forms, adding `isSubmitted()` reads more clearly. This detail is hiding inside of the best practices :).

javiereguiluz:
@mickaelandrieu thanks for your proposal. It's true that technically speaking this method is unneeded. But for the best practices book it was decided to use `->isSubmitted()` in the [form recomendations](http://symfony.com/doc/current/best_practices/forms.html) because it feels more natural.

Sadly, that's the reason why I must close this PR without merging it.

weaverryan:
@javiereguiluz it looks like we were typing at the same moment!

javiereguiluz:
@weaverryan we almost said the same thing at the very same moment in time in completely different parts of the world and without any previous coordination. How weird is that? :smile:

mickaelandrieu:
@javiereguiluz  @weaverryan ok, still sounds very weird to me :/ 

Why just not say clearly that a form is valid when:
* the form has been submitted
* the form has no validation errors
* this form is not disabled

This could be added directly on Best Practices part.

Anyway, I think symfony-demo is a very good initiative, thank you for this work :+1:

javiereguiluz:
@mickaelandrieu it may sound weird at first, but it you read the code as if it were English, you'll notice the following:

  * With `isSubmitted()`: *"if the form is submitted and valid, do this. Otherwise, do that."* OK, it makes sense.
  * Without `isSubmitted()`: *"if the form is valid, do this. Otherwise do that."* OK, but when do I send the form or process the sent form?

PR #4, opened 2015-03-24
Renamed Default to Static

WouterJ:
A little while ago, @weaverryan, @xabbuh and me decided Static was a much better name for this than Default.

Default doesn't really explain anything. Does it mean that every request will go to this controller *by default*? Nope...

Btw, this makes me wonder if we shouldn't use http://symfony.com/doc/current/cookbook/templating/render_without_controller.html instead of this controller?


javiereguiluz:
I fully agree with you: `Default` is not a great name ... but in my opinion, `Static` is much worse than `Default`. In the programming world, the `static` word implies a lot of things that have nothing to do with the purpose of this controller or the related templates. HTML pages are static, but Twig templates, PHP controllers, Symfony routing, etc. are not static.

However, I do like two things from your PR:

  * Rename `indexAction()` to `homepageAction()`
  * Render the `homepage.html.twig` without a controller. That would be a great example to showcase this feature.

WouterJ:
Well, the template is a static template. That's why we called it a StaticController, it just renders static pages (it could be a simple HTML template instead of twig). I'll update it with your suggestions.

WouterJ:
Updated to use the template controller. While I really don't like it, I've reverted the rename. :crying_cat_face:

javiereguiluz:
Thanks @WouterJ.

PR #11, opened 2015-03-25
Added ircmaxell/password-compat as a dependency

### LINKED ISSUE #10: Add ircmaxell/password-compat as a dependency
I'm running the demo in 5.4 and when i go to the login page i get an error because the password_hash function doesn't exists in 5.4, we can add ircmaxell/password-compat as a dependency to solve this. if you are ok with this i can send a PR :)

PD: Sorry for my english, i'm noob

dosten:
This PR fixes #10


javiereguiluz:
@dosten thanks for fixing this problem so quickly!

PR #14, opened 2015-03-26
Added a new command to create users

javiereguiluz:
The command it's almost finished. I'll add more help notes and the `interact()` method to show how to use it.

### DIFF
--- a/app/Resources/assets/scss/main.scss
+++ b/app/Resources/assets/scss/main.scss
@@ -144,6 +144,24 @@ body#homepage {
 // Page: 'login'
 // --------------------------------------------------
 body#login {
+    #login-users-help p {
+        font-size: $font-size-base;
+        line-height: $line-height-base;
+
+        &:last-child {
+            margin-bottom: 0;
+        }
+
+        .label {
+            margin-right: 5px;
+        }
+
+        .console {
+            display: block;
+            margin: 5px 0;
+            padding: 10px;
+        }
+    }
 }
 
 //
--- a/src/AppBundle/Command/AddUserCommand.php
+++ b/src/AppBundle/Command/AddUserCommand.php
@@ -44,8 +44,12 @@ class AddUserCommand extends ContainerAwareCommand
     protected function configure()
     {
         $this
+            // a good practice is to use the 'app:' prefix to group all your custom application commands
             ->setName('app:add-user')
             ->setDescription('Creates users and stores them in the database')
+            ->setHelp($this->getCommandHelp())
+            // commands can optionally define arguments and/or options (mandatory and optional)
+            // see http://symfony.com/doc/current/components/console/console_arguments.html
             ->addArgument('username', InputArgument::OPTIONAL, 'The username of the new user')
             ->addArgument('password', InputArgument::OPTIONAL, 'The plain password of the new user')
             ->addArgument('email', InputArgument::OPTIONAL, 'The email of the new user')
@@ -209,5 +213,37 @@ class AddUserCommand extends ContainerAwareCommand
         }
 
         return $email;
+    }
+
+    /**
+     * The command help is usually included in the configure() method, but when
+     * it's too long, it's better to define a separate method to maintain the
+     * code readability.
+     */
+    public function getCommandHelp()
+    {
+        return <<<HELP
+The <info>%command.name%</info> command creates new users and saves them in the database:
+
+  <info>php %command.full_name%</info> <comment>username password email</comment>
+
+By default the command creates regular users. To create administrator users,
+add the <comment>--is-admin</comment> option:
+
+  <info>php %command.full_name%</info> username password email <comment>--is-admin</comment>
+
+If you omit any of the three required arguments, the command will ask you to
+provide the missing values:
+
+  # command will ask you for the email
+  <info>php %command.full_name%</info> <comment>username password</comment>
+
+  # command will ask you for the email and password
+  <info>php %command.full_name%</info> <comment>username</comment>
+
+  # command will ask you for all arguments
+  <info>php %command.full_name%</info>
+
+HELP;
     }
 }
--- a/src/AppBundle/Command/AddUserCommand.php
+++ b/src/AppBundle/Command/AddUserCommand.php
@@ -1,0 +1,78 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace AppBundle\Command;
+
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use AppBundle\Entity\User;
+
+/**
+ * Creates users and stores them in the database
+ *
+ * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ */
+class AddUserCommand extends ContainerAwareCommand
+{
+    protected function configure()
+    {
+        $this
+            ->setName('app:add-user')
+            ->setDescription('Creates users and stores them in the database')
+            ->addArgument('username', InputArgument::REQUIRED, 'The username of the new user')
+            ->addArgument('password', InputArgument::REQUIRED, 'The plain password of the new user')
+            ->addArgument('email', InputArgument::REQUIRED, 'The email of the new user')
+            ->addOption('is-admin', null, InputOption::VALUE_NONE, 'If set, the user is created as an administrator')
+        ;
+    }
+
+    protected function initialize(InputInterface $input, OutputInterface $output)
+    {
+        $this->em = $this->getContainer()->get('doctrine')->getManager();
+
+        $this->username = $input->getArgument('username');
+        $this->plainPassword = $input->getArgument('password');
+        $this->email = $input->getArgument('email');
+        $this->isAdmin = $input->getOption('is-admin');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        // first check if a user with the same username already exists
+        $existingUser = $this->em->getRepository('AppBundle:User')->findOneBy(array(
+            'username' => $input->getArgument('username')
+        ));
+
+        if (null !== $existingUser) {
+            throw new \RuntimeException(sprintf('There is already a user registered with the "%s" username.', $this->username));
+
+        }
+
+        // create the user and encode its password
+        $user = new User();
+        $user->setUsername($this->username);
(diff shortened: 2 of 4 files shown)


mseshachalam:
Why not have the class level variables declared ?

stof:
They must be declared

javiereguiluz:
@mseshachalam good point. Done in 5f14ed7.

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    protected function initialize(InputInterface $input, OutputInterface $output)
+    {
+        $this->em = $this->getContainer()->get('doctrine')->getManager();
+
+        $this->username = $input->getArgument('username');
+        $this->plainPassword = $input->getArgument('password');
+        $this->email = $input->getArgument('email');
+        $this->isAdmin = $input->getOption('is-admin');
```
stof: assigning arguments and options should not be done here. ``initialize`` is called before the call to ``interact`` and the validation of the input (which happens after ``interact`` to give ``interact`` a chance to ask for missing mandatory arguments)

weaverryan: And I don't see the advantage of initializing the input to properties - we can just get the values right inside `execute()` :).

Also (feature creep!) it'd be cool to show off interact by asking them these questions.

javiereguiluz: @weaverryan interactive questions added in 9a48893 :)

javiereguiluz:
OK. The command is now finished and ready for the final review before merging it. Thanks.

app/Resources/assets/scss/main.scss
```diff
@@ -144,6 +144,24 @@ body#homepage {
 // Page: 'login'
 // --------------------------------------------------
 body#login {
+    #login-users-help p {
```
stof: an id selector inside an id selector ? Please no. We should try to follow CSS best practices too

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -0,0 +1,259 @@
...
+            'username' => $username = $input->getArgument('username'),
+        ));
+
+        if (null !== $existingUser) {
+            throw new \RuntimeException(sprintf('There is already a user registered with the "%s" username.', $username));
+        }
+
+        $this->username = $input->getArgument('username');
+        $this->plainPassword = $input->getArgument('password');
+        $this->email = $input->getArgument('email');
```
stof: I suggest using local variables for these instead of storing them in the command instance

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -0,0 +1,259 @@
...
+
+        return $email;
+    }
+
+    /**
+     * The command help is usually included in the configure() method, but when
+     * it's too long, it's better to define a separate method to maintain the
+     * code readability.
+     */
+    public function getCommandHelp()
```
stof: should be private

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -0,0 +1,259 @@
...
+
+    /**
+     * This method is executed before initialize() and execute(). Its purpose is
+     * to check if some of the options/arguments are missing and interactively
+     * ask the user for those values.
+     *
+     * This method is completely optional. If you are developing an internal console
+     * command, you probably should not implement this method because it requires
+     * quite a lot of work. However, if the command is meant to be used by external
+     * users, this method is a nice way to fall back an prevent errors.
```
stof: typo: ``and prevent`` (unless the issuer is bigger in the sentence)

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -0,0 +1,259 @@
...
+
+        if ($output->isVerbose()) {
+            $finishTime = microtime(true);
+            $elapsedTime = $finishTime - $startTime;
+
+            $output->writeln(sprintf('[INFO] New user database id: %d / Elapsed time: %.2f ms', $user->getId(), $elapsedTime*1000));
+        }
+    }
+
+    public function passwordValidator($plainPassword)
```
stof: I would tag them as ``@internal`` with a comment saying they are public for PHP 5.3 compat

app/Resources/assets/scss/main.scss
javiereguiluz: Generally speaking you're right. But in this case I'd like to maintain it to only apply these styles when we are in the login page (although I recognize that it looks difficult to imagine displaying this same message in other pages).

app/Resources/assets/scss/main.scss
```diff
@@ -144,6 +144,24 @@ body#homepage {
...
+
+        &:last-child {
+            margin-bottom: 0;
+        }
+
+        .label {
+            margin-right: 5px;
+        }
+
+        .console {
```
stof: should this style be restricted to the login page ? We might want to display some help with commands in other pages too

app/Resources/assets/scss/main.scss
stof: well, if you use the id ``login-users-help`` outside the login page, it would look weird regarding the naming convention for ids...

app/Resources/assets/scss/main.scss
javiereguiluz: The thing is that this "console" is very special in this page because I want to display it as a "soft console" not a "strong console" with a dark background. Here it is in action:

login-console

app/Resources/assets/scss/main.scss
javiereguiluz: It's true ... but we could use it for example in the 403 error page, where the `id` of the body is `error`.

stof: well, then the id ``login-users-*`` would be weird (a 403 page would not display a login form btw)

app/Resources/assets/scss/main.scss
stof: maybe we should have a ``.console-soft`` class to be used in place you want a soft console, and a ``.console-strong`` for places using a strong console. We could probably want to reuse the soft console on other pages too

javiereguiluz:
@stof I've decided to merge this PR as it was, without doing any additional CSS optimization. The reason is that I prefer to think about CSS optimization when we have more pages and we can find more "synergies". However, you can be certain your comments have been duly noted as always. Thanks for helping me fix/improve this command.

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -0,0 +1,265 @@
...
+use AppBundle\Entity\User;
+
+/**
+ * A command console that creates users and stores them in the database.
+ * To use this command, open a terminal window, enter into your project
+ * directory and execute the following:
+ *   $ php app/console app:add-user
+ *
+ * To output detailed information, increase the command verbosity:
+ *   $ php app/console app:add-user -vv
```
WouterJ: wouldn't it be more readable if there was an empty line before code blocks?

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -0,0 +1,265 @@
...
+{
+    const MAX_ATTEMPTS = 5;
+
+    /** @var ObjectManager */
+    private $em;
+
+    protected function configure()
+    {
+        $this
+            // a good practice is to use the 'app:' prefix to group all your custom application commands
```
WouterJ: what's the difference between a good practice or a best practice?

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -0,0 +1,265 @@
...
+     */
+    protected function interact(InputInterface $input, OutputInterface $output)
+    {
+        if (null !== $input->getArgument('username') && null !== $input->getArgument('password') && null !== $input->getArgument('email')) {
+            return;
+        }
+
+        $output->writeln('');
+        $output->writeln('Add User Command Interactive Wizard');
+        $output->writeln('-----------------------------------');
```
WouterJ: why not use an array here? (btw, you can gropu line 75 up to 92 into one writeln call)

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -0,0 +1,265 @@
...
+            'Now we\'ll ask you for the value of all the missing command arguments.',
+            '',
+        ));
+
+        // See http://symfony.com/doc/current/components/console/helpers/questionhelper.html
+        $console = $this->getHelper('question');
+
+        // Ask for the username if it's not defined
+        if (null === $username = $input->getArgument('username')) {
+            $question = new Question(' > <info>Username</>: ');
```
WouterJ: why not use `</info>`? It's much better to be explicit imo (and you know the start tag, as it's in the same string)

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -0,0 +1,265 @@
...
+            $question->setMaxAttempts(self::MAX_ATTEMPTS);
+
+            $username = $console->ask($input, $output, $question);
+            $input->setArgument('username', $username);
+        } else {
+            $output->writeln(' > <info>Username</>: '.$username);
+        }
+
+        // Ask for the password if it's not defined
+        if (null === $password = $input->getArgument('password')) {
```
WouterJ: I know this is Symfony style, but don't you think it's more readable to do this:

```php
$password = ...;
if (null === $password) {
```

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -0,0 +1,265 @@
...
+
+            $email = $console->ask($input, $output, $question);
+            $input->setArgument('email', $email);
+        } else {
+            $output->writeln(' > <info>Email</>: '.$email);
+        }
+    }
+
+    /**
+     * This method is executed after the interact() and before the execute()
```
WouterJ: it's executed before interact.

src/AppBundle/Command/AddUserCommand.php
javiereguiluz: To show the different ways of outputting contents. This demo application should be a "showcase" of most available possibilities.

src/AppBundle/Command/AddUserCommand.php
javiereguiluz: I agree. Done in 4f65c26.

src/AppBundle/Command/AddUserCommand.php
javiereguiluz: Some people get less nervous when you say "good practice" instead of "best practice".

src/AppBundle/Command/AddUserCommand.php
javiereguiluz: I agree. Done in 4f65c26.

src/AppBundle/Command/AddUserCommand.php
javiereguiluz: I agree again. Let's keep things simple. Done in 4f65c26.

src/AppBundle/Command/AddUserCommand.php
javiereguiluz: Ops! Thanks for reporting this error. Fixed in 4f65c26.

PR #15, opened 2015-03-28
Optimized slugger: trimming dash at the beginning & end and removed doubles dashes

sebastianblum:
Hello,

I found the symfony-demo repository and the idea behind is great.

also in the symfony best practices, the slugger has 2 problems in my opinion.
* several dashes like lorem--ipsum
* dashes at the begin or end like -lorem-ipsum-

here is my litte pull request.
sebastian


javiereguiluz:
@sebastianblum thanks for creating this PR. I love it when improvements are short and simple!

Just a note for future improvements: in this particular application we don't care about the actual slugger logic. We just use it as an example of how to define a service for classes that implement a simple feature. However, since your PR doesn't complicate anything and it includes a much needed improvement, it makes a lot of sense to merge it. Thanks.

PR #16, opened 2015-03-28
Remove redundant function call

WouterJ:
After https://github.com/symfony/symfony-demo/pull/15, the inner `trim()` call is no longer required as the outer one already trims all dashes.


javiereguiluz:
@WouterJ thanks for this quick win!

PR #17, opened 2015-03-29
Changed the README a bit

WouterJ:
* A bit personally, but an ordered list using numbers seem to indicate a list of steps, while an ordered list using characters indicates different options;
* The demo command was added to the symfony installer (and is available if you download it from symfony.com);
* Some code blocks used the bash highlighting, while others didn't;
* Link to symfony's installing composer guide, as it explains how to install globally (composer still doesn't document this on its download page).

Btw, why does the guide use `server:run` and not `server:start`? (maybe to be compatible with Windows users?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,36 +10,36 @@ Installation
 This is a fully-functional Symfony project. If you want to get it running,
 you have two alternatives:
 
-### 1) Use the Symfony Installer
+### A) Using the Symfony Installer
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks @WouterJ. This is really a nice set of improvements! And you are right, I used server:run because I think it's "more compatible" than server:start

PR #20, opened 2015-03-30
Added a new app:list-users command to show BufferedOutput + Table + Emails

javiereguiluz:


app/config/services.yml
```diff
@@ -1,5 +1,5 @@
 parameters:
-    # ...
+    app.notifications.email_sender: 'anonymous@symfony.com'
```
xabbuh: What about using example.com as the domain part as it is reserved for such things?

javiereguiluz: I think your proposal makes sense. Done. Thanks.

app/config/services.yml
```diff
@@ -1,5 +1,5 @@
 parameters:
-    # ...
+    app.notifications.email_sender: 'anonymous@example.com'
```
WouterJ: no need to quote

src/AppBundle/Command/ListUsersCommand.php
```diff
@@ -0,0 +1,133 @@
...
+ *
+ *     $ php app/console app:list-users
+ *
+ * See http://symfony.com/doc/current/cookbook/console/console_command.html
+ *
+ * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ */
+class ListUsersCommand extends ContainerAwareCommand
+{
+    /** @var ObjectManager */
```
WouterJ: Symfony CS never uses one-line docblocks

src/AppBundle/Command/ListUsersCommand.php
```diff
@@ -0,0 +1,133 @@
...
+    private function sendReport($contents, $recipient)
+    {
+        // See http://symfony.com/doc/current/cookbook/email/email.html
+        $mailer = $this->getContainer()->get('mailer');
+
+        $message = $mailer->createMessage()
+            ->setSubject(sprintf('app:list-users report (%s)', date('Y-m-d H:i:s')))
+            ->setFrom($this->getContainer()->getParameter('app.notifications.email_sender'))
+            ->setTo($recipient)
+            ->setBody(sprintf('<pre>%s</pre>', $contents), 'text/html')
```
stof: you are sending a plain-text table inside a ``<pre>`` tag in HTML part. Why not sending a plain text part instead ?

src/AppBundle/Command/ListUsersCommand.php
```diff
@@ -0,0 +1,133 @@
...
+
+        // Doctrine query returns an array of objects and we need an array of plain arrays
+        $usersAsPlainArrays = array_map(function ($user) {
+            return array($user->getId(), $user->getUsername(), $user->getEmail(), implode(', ', $user->getRoles()));
+        }, $users);
+
+        // Instead of the common output type, this command uses the BufferedOutput
+        // type, which allows to get the contents before displaying them in the
+        // console. You probably should never use this kind of output, because it's
+        // only suited for very special use cases, like the one proposed by this command
```
stof: should we explain a bit more what the special case is ?

PR #22, opened 2015-04-02
Added @Method constraints where appropriate and added more help notes

javiereguiluz:


94noni:
:+1: for the comments (and the commands you've created)

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -153,7 +168,11 @@ public function deleteAction(Request $request, Post $post)
     }
 
     /**
-     * Creates a form to delete a Post entity by id.
+     * Creates a form to delete a Post entity by id. This is necessary because
```
WouterJ: the explanation should be in the long descrpition, not the short one.

PR #23, opened 2015-04-02
Added some functional tests for the backend

javiereguiluz:
This allows to explain how to test resources protected by a firewall and the trick to change the firewall configuration just for the test environment.


src/AppBundle/Tests/Controller/Admin/BlogControllerTest.php
```diff
@@ -0,0 +1,82 @@
...
+            'PHP_AUTH_USER' => 'john_user',
+            'PHP_AUTH_PW'   => 'kitten',
+        ));
+
+        $client->request('GET', '/admin/post/');
+
+        $this->assertEquals(
+            Response::HTTP_FORBIDDEN,
+            $client->getResponse()->getStatusCode(),
+            'Regular users cannot access to the backend.'
```
WouterJ: is it really needed to repeat the method name in the error message? You can use `phpunit --testdox` to see these messages nicely already.

src/AppBundle/Tests/Controller/Admin/BlogControllerTest.php
```diff
@@ -0,0 +1,82 @@
...
+            'PHP_AUTH_USER' => 'anna_admin',
+            'PHP_AUTH_PW'   => 'kitten',
+        ));
+
+        $client->request('GET', '/admin/post/');
+
+        $this->assertEquals(
+            Response::HTTP_OK,
+            $client->getResponse()->getStatusCode(),
+            'Regular users cannot access to the backend.'
```
WouterJ: wrong error message

src/AppBundle/Tests/Controller/Admin/BlogControllerTest.php
javiereguiluz: I agree with you. I've just removed the redundant messages. Thanks.

app/config/config_test.yml
```diff
@@ -14,3 +14,19 @@ web_profiler:
 
 swiftmailer:
     disable_delivery: true
+
+# It's recommended to use a separate database for tests. This allows to have a
+# fixed and known set of data fixtures, it simplifies the code of tests and it
+# makes them more robust.
+# In this case we just define a different path for the application database,
+# but you can even use an entirely database engine for tests.
```
stof: using a different database engine should not be recommended IMO: some of your queries might depend on the engine being used, and would not work on the other.
Even if you rely on Doctrine only and never write SQL yourselves, it may have platform-specific bugs (especially on platforms being less tested like SqlServer or Oracle), for instance when paginating complex queries (there is a huge work in the ORM currently to make the pagination actually work on complex queries). And some features are simply not supported on some platforms

stof:
should we really commit the test DB ? The day you write a functional test for non-safe requests, this will hurt you

javiereguiluz:
Yes, I think we should commit the test DB to meet our promise: *"everything works perfectly out-of-the-box"*. ("everything" means frontend, backend and tests ... and "out-of-the-box" means: 0 questions to answer, 0 options to provide, 0 commands to execute and 0 config changes).

stof:
@javiereguiluz we also recommend following best practices. and commiting a DB does not look like a best practice. It can hurt your workflow quickly (you will also need to update it when changing your doctrine schema btw, which may create wunky git conflicts because of being a binary file).

A much better setup IMO is making the test self-contained by making it load the necessary content in the test DB.

javiereguiluz:
I understand you now. You are right. We'll load the fixtures dynamically in the test in a future pull request. But before I'm waiting for this old Symfony doc issue to be completed: https://github.com/symfony/symfony-docs/issues/480  but it seems nobody wants to work on it :(

PR #25, opened 2015-04-07
Use DIRECTORY_SEPARATOR constant to avoid Windows issues

javiereguiluz:
This PR fixes the issue reported at https://github.com/symfony/symfony/issues/14239

### DIFF
--- a/src/AppBundle/Twig/SourceCodeExtension.php
+++ b/src/AppBundle/Twig/SourceCodeExtension.php
@@ -82,8 +82,8 @@ class SourceCodeExtension extends \Twig_Extension
         $class = new \ReflectionClass($className);
 
         $absolutePath = $class->getFilename();
-        $pathParts = explode('/src/', $absolutePath);
-        $relativePath = 'src/'.$pathParts[1];
+        $pathParts = explode(DIRECTORY_SEPARATOR.'src'.DIRECTORY_SEPARATOR, $absolutePath);
+        $relativePath = 'src'.DIRECTORY_SEPARATOR.$pathParts[1];
 
         return $relativePath;
     }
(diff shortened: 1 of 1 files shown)


src/AppBundle/Twig/SourceCodeExtension.php
```diff
@@ -95,7 +95,7 @@ private function getTemplateCode()
 
     private function getTemplateRelativePath()
     {
-        return 'app/Resources/views/'.$this->template->getTemplateName();
+        return 'app'.DIRECTORY_SEPARATOR.'Resources'.DIRECTORY_SEPARATOR.'views'.DIRECTORY_SEPARATOR.$this->template->getTemplateName();
```
stof: This is not needed here, given that ``Twig_Template::getTemplateName()`` does not use ``DIRECTORY_SEPARATOR`` anyway.

On a side note, this implementation is fragile: if your template uses a namespaced path (``@WebProfiler/Collector/time.html.twig`` for instance) or any other loader than the Twig_Loader_Filesystem (the ``TwigBundle:Exception:exception.txt.twig`` notation for instance, or an anonymous template created by the ``{% embed %}`` tag for instance), the template name will not be a relative path at all.

javiereguiluz: Thanks for this information. I agree completely with you. However, we're lucky because this `show_source_code()` function is included only in the templates that we decide, so it doesn't have to work right for all the possible cases (at least for now).

stof: I suggest adding a comment in the method phpdoc warning about this though, in case people read the code of the demo and want to do the same in their own place

javiereguiluz: Good idea! Done in a4c538d. Thanks.

PR #26, opened 2015-04-07
Fix the escaping of the source code when displaying it

stof:
The source code of the controller cannot be considered as being safe HTML. It needs to be escaped.

This also moves the escaping of the template source code from the extension to the template rendering it, where it belongs (which also avoid using ``|raw``, which is flagged by Insight as being dangerous because it expects the escaping to happen elsewhere, see the first point about the controller source code)


javiereguiluz:
@stof thanks! A lot of improvements with such a small change. I love it!

PR #32, opened 2015-04-09
Updated Composer requirements

javiereguiluz:
First of all, **we can't merge this pull request**. I've just opened to get your feedback and comments about it.

In my computer, the demo application works perfectly and I see no errors or warnings. However, if I set a PHP extension constraint in PHP, I get the following Composer errors:

### ext-pdo-sqlite

ext-pdo-sqlite

### ext-sqlite

ext-sqlite

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     },
     "require": {
         "php"                                  : ">=5.3.3",
-        "ext-sqlite"                           : "*",
+        "ext-pdo_sqlite"                       : "*",
         "doctrine/doctrine-bundle"             : "~1.2",
         "doctrine/doctrine-fixtures-bundle"    : "~2.2",
         "doctrine/orm"                         : "~2.2,>=2.2.3",
--- a/composer.json
+++ b/composer.json
@@ -8,6 +8,7 @@
     },
     "require": {
         "php"                                  : ">=5.3.3",
+        "ext-sqlite"                           : "*",
         "doctrine/doctrine-bundle"             : "~1.2",
         "doctrine/doctrine-fixtures-bundle"    : "~2.2",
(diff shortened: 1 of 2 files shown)


weaverryan:
@javiereguiluz should it be ext-pdo_sqlite (since the extension is called `pdo_sqlite`)? I haven't needed to add an ext- before, so I thought I'd make the easy suggestion first :).

javiereguiluz:
@weaverryan that was exactly the problem! Everything works now. Thanks!

JavierFuentes:
Hi @javiereguiluz, I am a rookie in Symfony and I am making my first deploy with a brand new copy of Symfony Demo's project.

Locally, all is running OK...

I have not too many permissions in my Server and I can't access via SSH to execute `$ composer install` command so I only can copy all my project via FTP to the Server and expect it'll work.

When I go to **~/web/config.php** page in my Server, all is OK, but when I navigate inside de Demo site, I have **Error 500** because SQLite PDO is not activated in my Server (I will need an Administrator do it)...

Perhaps, **config.php** could advertised me about this to know what was the problem.

I have a bit experience in other technologies like Angular2+ where Angular CLI's projects are prepared to make a **build --prod** out of the box... I am surprised how Symfony have not any automatic scripts to do the same in **composer.json** to generate a **/build copy ready for production**, and how I must copy all the files and folders to my Server to make it run (I don't know if there is another way of do it)...

I have been investigating a little about [Robo](http://robo.li/) **Task Runner** and I think it is very cool... perhaps you could use this tool in this Demo and all new templates generated by `$ php symfony new <my-project>` command...

PS.- I have seen that 2 Test are failing now in last version of this Demo too.
Congratulations for your good work.

PR #35, opened 2015-04-20
Remove composer.lock

RobinDev:


stof:
-1. The demo is a project, so it should lock its dependencies.

RobinDev:
OK !

Sorry for the loss of time. Thanks for the learning.

PR #36, opened 2015-04-20
[config] use form_themes instead of deprecated form.resources.

aitboudad:


javiereguiluz:
Thanks @aitboudad for this quick win!

PR #39, opened 2015-04-20
[Backend] remove unused admin.css.

aitboudad:


javiereguiluz:
@aitboudad thanks for this fix. I'm sure that in the future we'll need to add some admin.css to aovid messing the other CSS files. But for now we don't have that admin.css file, so it's better to entirely remove this block.

PR #33, opened 2015-04-10
Moved web assets to web/ and implemented a pure PHP compression solution for CSS and JS files

javiereguiluz:
This fixes #5

PR #40, opened 2015-04-20
Refers to nonexistent stylesheet admin.css

bocharsky-bw:
Refers to nonexistent stylesheet **admin.css** in `app/Resources/views/admin/layout.html.twig` template.


javiereguiluz:
@bocharsky-bw thanks for reporting and fixing this issue.

In #39 we had a previous and different solution to this problem. Your solution is simpler (and I love that) but this time I merged the other solution. The main reason is that for now we don't have any content for the `admin.css`. The other reason is that in #33 we have a pending PR related to web assets and we need to make a decission about how to organize those assets.

PR #41, opened 2015-04-20
Translate to Russian and Ukrainian 

bocharsky-bw:


javiereguiluz:
@bocharsky-bw as you can see, the current i18n is in its early stages and we need to work on the full i18n of the application (there is a pending issue to do that). But for now, it's OK to start adding translations. Thank you very much!

PR #43, opened 2015-04-20
Fixed backend design for the listing actions

javiereguiluz:
I was playing with the Russian translation kindly provided by @bocharsky-bw and I noticed the following issue:

before

After this PR, the actions should always be correctly displayed:

after


stof:
:+1:

bocharsky-bw:
I probably missed it. Thank you, @javiereguiluz

PR #45, opened 2015-04-20
Minor improvment 

JulienItard:
- Add french translation
- Fix comments typo
- Add isSubmitted call + remove useless method attribute


javiereguiluz:
Nice improvements! Thank you very much @JulienItard

PR #46, opened 2015-04-21
Update BlogController.php

ZCJ:
Does't form->isValid already check id if the form has been submitted? Why do we need to check $form->isSubmitted()


javiereguiluz:
@ZCJ thanks for preparing this pull request!

Technically, you are 100% right and the `isSubmitted()` method is not necessary. But we include it to improce code readability. In this closed pull request we explain the reasons: #2

To avoid any further issues with this line of code, we've added a help note in 3b1446008770f00b972c93e9205065b5d32184a1

PR #48, opened 2015-04-21
Removed unneeded Request argument

frastel:
The assigned Request instance is currently not needed in the login action and thus should be removed from the methods signature.


javiereguiluz:
@frastel nice catch! There's no need to declare unused arguments. Thanks!

PR #49, opened 2015-04-21
Remove unnecessary call of isSubmitted method on forms

bocharsky-bw:
The `isSubmitted` already check in `isValid` method (https://github.com/symfony/Form/blob/master/Form.php#L771) and not need in this statement.


javiereguiluz:
@bocharsky-bw thanks for this pull request.

You are technically right, but there are some other reasons to include that method. Check out #2 and #46 for the detailed discussion. Thanks!

PR #50, opened 2015-04-21
Improved the help notes of the production front controller

### LINKED ISSUE #47: Duplicate instantiation of AppKernel in app.php
Hello.

According to the comments in app.php, in order to enable the Symfony Reverse Proxy, you must uncomment the lines:

```
// require_once __DIR__.'/../app/AppCache.php';
// $kernel = new AppKernel('prod', false);
// $kernel->loadClassCache();
// $kernel = new AppCache($kernel);
// Request::enableHttpMethodParameterOverride();
```

However, the AppKernel is instantied and the ClassCache is loaded right before this block. 

Is there a reason for re-constructing the kernel and loading the ClassCache only after
`require_once __DIR__.'/../app/AppCache.php';`?

Thank you!

javiereguiluz:
This tries to fix #47

### DIFF
--- a/web/app.php
+++ b/web/app.php
@@ -15,8 +15,8 @@ $loader = require_once __DIR__.'/../app/bootstrap.php.cache';
 // lines to use APC for class autoloading. This can improve application performance
 // very significantly. See http://symfony.com/doc/current/components/class_loader/cache_class_loader.html#apcclassloader
 //
-// NOTE: The first argument of ApcClassLoader() is a prefix that will be used to
-// prevent cache key conflicts. In a real Symfony application, make sure to change
+// NOTE: The first argument of ApcClassLoader() is the prefix used to prevent
+// cache key conflicts. In a real Symfony application, make sure to change
 // it to a value that it's unique in the web server. A common practice is to use
 // the domain name associated to the Symfony application (e.g. 'example_com').
 //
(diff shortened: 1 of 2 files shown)


stof:
Given that enablign the HttpCache only adds a few more lines but does not remove any of the previous lines, and that the new lines can be grouped together (it is totally fine to require the AppCache class only after instantiating the AppKernel), what about removing the need to comment all previous lines and uncomment the same lines later ? Using the HttpCache only requires adding 2 or 3 lines in the front controller (depending on whether you want to enable the method override or no).

IMO, the current way only makes things harder for the user.

javiereguiluz:
@stof you are right. I've simplified everything in e03d177 following your comments. Thanks.

web/app.php
```diff
@@ -27,12 +33,11 @@
 // See http://symfony.com/doc/current/book/http_cache.html#symfony-reverse-proxy
 //
 // require_once __DIR__.'/../app/AppCache.php';
-// $kernel = new AppKernel('prod', false);
-// $kernel->loadClassCache();
 // $kernel = new AppCache($kernel);
 // Request::enableHttpMethodParameterOverride();
```
stof: this line should be separate from others IMO. It is not always necessary to uncomment it (it depends of whether you expect this to be enabled or no, and my advice would be to keep it disabled for websites and avoiding to expect a DELETE or PUT request, which would need to be emulated all the time. It simplifies the need to protect against CSRF in places where you would not expect to need this. Websites like this demo application are targetting browsers, so it is fine to take browser limitations into account when designing them IMO)

javiereguiluz: But in the backend we use the `DELETE` method (https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Controller/Admin/BlogController.php#L186)  Can we still ignore this line?

stof: @javiereguiluz this is what I suggest changing actually. Making the site expect a delete request which would always be emulated using a way requiring a CSRF token to be safe is not worth it IMO. It still does not allow to perform a real delete request for supported clients (because you would still need the CSRF token) and browsers will all require using POST. I suggest acknowledging it fully and aviding the need to allow parameter-based method overriding

javiereguiluz: OK then. In the future we'll refactor the application to stop faking the HTTP method with the `_method` hack.

For now I've improved the note about this line as follows:

```
// If you use HTTP Cache and your application relies on the _method request parameter
// to get the intended HTTP method, uncomment this line.
// See http://symfony.com/doc/current/reference/configuration/framework.html#http-method-override
// Request::enableHttpMethodParameterOverride();
```

stof: looks like this improved note was not added in the PR before merging

javiereguiluz: Thanks for noticing! I've readded it in 0a1c24f

Issue #51, opened 2015-04-21
Add SQLite requirement


JulienItard:
We should add a Requirements section on the README and add PHP SQLite extentsion.

javiereguiluz:
We recently added a Composer dependency for this: https://github.com/symfony/symfony-demo/blob/master/composer.json#L11

But I think you are right and we should add a note about this in the README.

JulienItard:
Ok i'll do it

javiereguiluz:
Fixed in #52

PR #52, opened 2015-04-21
Add SQLite requirements #51

JulienItard:


javiereguiluz:
@JulienItard thanks for proposing this improvement in the first place and then creating a fix for it.

I've merged it as is to give you the credit of the pull request and then I've tweaked it a bit in https://github.com/symfony/symfony-demo/commit/7242ce6c3b0845247ca25eda1590a3d24132f49e

I've added the PHP 5.3 requirement, which I guess is important to show it for people running old PHP versions and a link to the Symfony requirements page, which explains things nicely and it's a page that usually doesn't receive much attention.

Lastly, I removed the specific instructions of the SQLite extension because we should provide the instructions for all the different Operating Systems and types of PHP installations.

xabbuh:
@javiereguiluz We should make clear that you don't have to use PHP 5.3, but PHP 5.3 **or** higher.

javiereguiluz:
@xabbuh thanks for the heads up. You are completely right. Fixed. Thanks!

JulienItard:
+1

PR #53, opened 2015-04-21
Translate to Romanian

cristi-contiu:
Romanian translations of initial i18n


stof:
@javiereguiluz  I think #8 should become higher priority now that the community starts translating the app, to avoid having to redo the work

javiereguiluz:
@stof you are right ... but that task is so boring to me that I'm waiting to see if any community member offers himself/herself to do it :smile:

javiereguiluz:
Great! Another translation! Thanks @cristi-contiu

PR #55, opened 2015-04-23
Stresses the best practice Constants vs Configuration Options

simohammedhttp:


javiereguiluz:
Nice addition. Thanks @simohammedhttp

PR #58, opened 2015-04-23
Change URL to login form documentation

xelaris:
Although `http://symfony.com/doc/current/cookbook/security/form_login.html` is a valid url, I think `http://symfony.com/doc/current/cookbook/security/form_login_setup.html` is more applicable here.


javiereguiluz:
Much better link! Thanks @xelaris

PR #60, opened 2015-04-24
Add show code in view post

mashware:


javiereguiluz:
Nice catch and a quick win! Thanks @mashware

PR #62, opened 2015-04-27
The controller and template file paths are now clickable links

### LINKED ISSUE #30: In the "view source code" button, it would be nice to be able to click on file name
(A mini reminder for myself ; But if anyone want to implement it before me ...)

javiereguiluz:
This fixes #30

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -13,7 +13,8 @@ imports:
 framework:
     # Uncomment the 'ide' option to turn file paths into clickable links that
     # open the given file path and line number using your favorite IDE.
-    # Supported values are 'textmate', 'macvim', 'emacs' and 'sublime'
+    # Supported values are 'textmate', 'macvim', 'emacs' and 'sublime' shortcuts
+    # and any custom configuration string, such as: "phpstorm://open?file=%%f&line=%%l"
     # See http://symfony.com/doc/current/reference/configuration/framework.html#ide
     #
     # ide: sublime
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -17,7 +17,7 @@ framework:
     # and any custom configuration string, such as: "phpstorm://open?file=%%f&line=%%l"
     # See http://symfony.com/doc/current/reference/configuration/framework.html#ide
     #
-    # ide: sublime
+    ide: sublime
 
     # esi:             ~
     translator:      { fallback: "%locale%" }
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -11,10 +11,13 @@ imports:
 
 # Basic configuration for the Symfony framework features
 framework:
-    # uncomment the 'ide' option to turn file paths into clickable links that
+    # Uncomment the 'ide' option to turn file paths into clickable links that
     # open the given file path and line number using your favorite IDE.
     # Supported values are 'textmate', 'macvim', 'emacs' and 'sublime'
+    # See http://symfony.com/doc/current/reference/configuration/framework.html#ide
+    #
     # ide: sublime
+
     # esi:             ~
     translator:      { fallback: "%locale%" }
     secret:          "%secret%"
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -11,12 +11,11 @@ imports:
 
 # Basic configuration for the Symfony framework features
 framework:
-    # Uncomment the 'ide' option to turn file paths into clickable links that
-    # open the given file path and line number using your favorite IDE.
+    # Uncomment the 'ide' option to turn all of the file paths in an exception
+    # page into clickable links that open the given file using your favorite IDE.
     # Supported values are 'textmate', 'macvim', 'emacs' and 'sublime' shortcuts
     # and any custom configuration string, such as: "phpstorm://open?file=%%f&line=%%l"
     # See http://symfony.com/doc/current/reference/configuration/framework.html#ide
-    #
     ide: sublime
 
     # esi:             ~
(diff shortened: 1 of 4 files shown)


app/config/config.yml
```diff
@@ -11,6 +11,13 @@ imports:
 
 # Basic configuration for the Symfony framework features
 framework:
+    # Uncomment the 'ide' option to turn file paths into clickable links that
+    # open the given file path and line number using your favorite IDE.
+    # Supported values are 'textmate', 'macvim', 'emacs' and 'sublime'
+    # See http://symfony.com/doc/current/reference/configuration/framework.html#ide
+    #
+    # ide: sublime
```
aitboudad: IMO It's better to move `ide` into `config_dev.yml`

aitboudad: and also it become useless when `xdebug.file_link_format` is set,

stof: @aitboudad why in config_dev ? It could be set for any environment without issue.

However, I would indeed recomment setting ``xdebug.file_link_format`` whenever possible. this allows configuring it only once for any project on your machine, and it makes things easier if people in your team are not using the same IDE

app/config/config.yml
```diff
@@ -11,6 +11,13 @@ imports:
 
 # Basic configuration for the Symfony framework features
 framework:
+    # Uncomment the 'ide' option to turn file paths into clickable links that
+    # open the given file path and line number using your favorite IDE.
+    # Supported values are 'textmate', 'macvim', 'emacs' and 'sublime'
```
stof: it also supports any custom pattern for IDEs not known by Symfony

src/AppBundle/Twig/SourceCodeExtension.php
```diff
@@ -82,12 +86,15 @@ private function getControllerRelativePath()
...
+        $projectBaseDir = realpath($this->kernelRootDir.'/..');
         $absolutePath = $class->getFilename();
-        $pathParts = explode(DIRECTORY_SEPARATOR.'src'.DIRECTORY_SEPARATOR, $absolutePath);
-        $relativePath = 'src'.DIRECTORY_SEPARATOR.$pathParts[1];
+        $relativePath = ltrim(str_replace($projectBaseDir, '', $absolutePath), DIRECTORY_SEPARATOR);
+
+        $fileLink = strtr($this->fileLinkFormat, array('%f' => $absolutePath, '%l' => $method->getStartline()));
 
-        return $relativePath;
+        return empty($fileLink) ? $relativePath : new \Twig_Markup(sprintf('<a href="%s">%s</a>', $fileLink, $relativePath), 'UTF-8');
```
stof: you need to escape the link and the path. Otherwise, your Twig_Markup is not safe

src/AppBundle/Twig/SourceCodeExtension.php
```diff
@@ -103,9 +110,15 @@ private function getTemplateCode()
...
+    private function getTemplateFilePath()
     {
-        return 'app/Resources/views/'.$this->template->getTemplateName();
+        $projectBaseDir = realpath($this->kernelRootDir.'/..');
+        $absolutePath = $this->kernelRootDir.'/Resources/views/'.$this->template->getTemplateName();
+        $relativePath = ltrim(str_replace($projectBaseDir, '', $absolutePath), DIRECTORY_SEPARATOR);
+
+        $fileLink = strtr($this->fileLinkFormat, array('%f' => $absolutePath, '%l' => 1));
+
+        return empty($fileLink) ? $relativePath : new \Twig_Markup(sprintf('<a href="%s">%s</a>', $fileLink, $relativePath), 'UTF-8');
```
stof: I suggest extracting the logic building a relative path into a separate private method to avoid duplicating it

stof:
Btw, you could use the ``|format_file`` filter available in Symfony, which handles building the link whenever possible. It would be better than reimplementing the logic yourselves

app/config/config.yml
aitboudad: @stof why we need to set this value into all environment as I think it's useful only in dev.

stof: @aitboudad it is useful for any environment you could run locally. Even though you generally run the dev env locally, it is not the only one. I regularly switch to app.php when I want to perform some profiling for instance, to avoid the overhead of all the symfony debugging tools, which is not relevant for the performance of my app.

aitboudad: @stof I understand, but IMO this isn't the common case anyway let's keep it in `config.yml`

stof:
@javiereguiluz ping

javiereguiluz:
I've updated this PR to use the `format_file` filter and to simplify everything else.

src/AppBundle/Twig/SourceCodeExtension.php
```diff
@@ -25,10 +25,14 @@ class SourceCodeExtension extends \Twig_Extension
...
     protected $template;
+    protected $kernelRootDir;
+    protected $fileLinkFormat;
 
-    public function __construct(\Twig_LoaderInterface $loader)
+    public function __construct(\Twig_LoaderInterface $loader, $kernelRootDir, $fileLinkFormat)
     {
+        $this->kernelRootDir = $kernelRootDir;
         $this->loader = $loader;
+        $this->fileLinkFormat = $fileLinkFormat ?: ini_get('xdebug.file_link_format') ?: get_cfg_var('xdebug.file_link_format');
```
stof: this seems unused now

javiereguiluz: Good catch! Fixed in 6297ae3 Thanks.

PR #64, opened 2015-04-29
Fixed the help notes of the AddUser command

javiereguiluz:
In the [run() method of the Command class](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Console/Command/Command.php#L220-263) you can see the following:

```php
public function run(InputInterface $input, OutputInterface $output)
{
    // ...

    $this->initialize($input, $output);

    // ...

    if ($input->isInteractive()) {
        $this->interact($input, $output);
    }

    // ...
    
    $statusCode = $this->execute($input, $output);

    return is_numeric($statusCode) ? (int) $statusCode : 0;
}
```

Therefore, our help note was wrong and the method order is ` initialize()` -> `interact()` -> `execute()`

### DIFF
--- a/src/AppBundle/Command/AddUserCommand.php
+++ b/src/AppBundle/Command/AddUserCommand.php
@@ -61,8 +61,18 @@ class AddUserCommand extends ContainerAwareCommand
     }
 
     /**
-     * This method is executed before initialize() and execute(). Its purpose is
-     * to check if some of the options/arguments are missing and interactively
+     * This method is executed before the interact() and the execute() methods.
+     * It's main purpose is to initialize the variables used in the rest of the
+     * command methods.
+     */
+    protected function initialize(InputInterface $input, OutputInterface $output)
+    {
+        $this->em = $this->getContainer()->get('doctrine')->getManager();
+    }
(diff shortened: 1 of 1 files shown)


src/AppBundle/Command/AddUserCommand.php
```diff
@@ -61,8 +61,18 @@ protected function configure()
     }
 
     /**
-     * This method is executed before initialize() and execute(). Its purpose is
-     * to check if some of the options/arguments are missing and interactively
+     * This method is executed before the interact() and the execute() methods.
+     * It's main purpose is to initialize the variables used in the rest of the
+     * command methods.
```
stof: I would add a note saying that the input has not been validated yet here (as this is done only after ``interact``, so the input should be used with great caution here (or not used at all)

javiereguiluz: Good point! I've added the following:

```
    /**
     * // ...
     * Beware that the input options and arguments are validated after executing
     * the interact() method, so you can't blindly trust their values in this method.
     */
```

PR #65, opened 2015-04-29
Update jQuery to v2.1.4

tronsha:
http://blog.jquery.com/2015/04/28/jquery-1-11-3-and-2-1-4-released-ios-fail-safe-edition/


javiereguiluz:
@tronsha thanks for sending this PR.

Because of the way this demo application works, it's needed to also adde the source jQuery files and to update the link in the base template. I've picked your work in this pull request and finish it in #66 to do that. Thanks!

PR #70, opened 2015-05-23
Improve the checking if the current request is a master request in the ControllerListener.

voronkovich:
There is a more convenient way to check whether the current request is a master request. See http://api.symfony.com/2.7/Symfony/Component/HttpKernel/Event/KernelEvent.html#method_isMasterRequest

### DIFF
--- a/src/AppBundle/EventListener/ControllerListener.php
+++ b/src/AppBundle/EventListener/ControllerListener.php
@@ -11,7 +11,6 @@
 
 namespace AppBundle\EventListener;
 
-use Symfony\Component\HttpKernel\HttpKernelInterface;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@voronkovich thanks for this improvement. I do agree with you that this proposal is much better. You don't need to know the internal Symfony details when you are learning Symfony.

In any case, before merging it, I'll wait a bit so others can comment if they don't agree or see some drawback. Cheers!

JulienItard:
I agree with it because it's more readable and less "frigthening" for beginners.

weaverryan:
:+1:

javiereguiluz:
@voronkovich thanks for helping us improve the code of the Symfony Demo application!:wq!

PR #74, opened 2015-06-30
Add check.php file for access from web

bocharsky-bw:
I propose to get Symfony's `web/config.php` configurator file and rename it to `web/check.php` with some minor changes in order to get ability to check required and optional environment configuration parameters directly from web browser for success run this demo project. It's like `app/check.php` file for console, but allow users to check environment directly from web browser. I think it will be useful for beginners.
What do you think, guys?

### DIFF
--- a/web/check.php
+++ b/web/check.php
@@ -1,0 +1,122 @@
+<?php
+
+if (!isset($_SERVER['HTTP_HOST'])) {
+    exit('This script cannot be run from the CLI. Run it from a browser.');
+}
+
+if (!in_array(@$_SERVER['REMOTE_ADDR'], array(
+    '127.0.0.1',
+    '::1',
+))) {
+    header('HTTP/1.0 403 Forbidden');
+    exit('This script is only accessible from localhost.');
+}
+
+require_once dirname(__FILE__).'/../app/SymfonyRequirements.php';
+
+$symfonyRequirements = new SymfonyRequirements();
+
+$majorProblems = $symfonyRequirements->getFailedRequirements();
+$minorProblems = $symfonyRequirements->getFailedRecommendations();
+
+?>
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
ping @javiereguiluz

javiereguiluz:
I'm not strongly against this, but I have some concerns:

1. If you are an experienced dev, you don't need it.
2. If you are a newcomer dev, you always install the demo with the Symfony Installer. This installer already checks the requirements and displays the list of errors to fix.

So my question is: do you think that newcomers have different `php.ini` configurations for the CLI and the web? If yes, I agree on adding this `check.php` file. If not, it could be unnecessary to do that.

bocharsky-bw:
@javiereguiluz Yes, I considered different `php.ini` for CLI and web too. I also think not many newcomers are know how to use web `php.ini` config file for console or simply forgot to do this when checking.

One more reason is to use familiar behavior to check current environment for **Symfony Demo Application** and **Symfony SE**, so I don't sure yet left this file name `config.php` for consistency with **Symfony SE** or rename it to `check.php`.

stof:
@javiereguiluz All Windows installer for PHP I know of are using different php.ini.
and Ubuntu packages are also using different php.ini files (even though enabling an extension with ``php5enmod`` will enable it for all SAPIs)

xabbuh:
:+1: It's not only Ubuntu. Many other Linux distributions follow this approach too

javiereguiluz:
@bocharsky-bw thanks for this pull request. Before merging it, I've decided to update its look-and-feel to better integrate it in the current demo application. To do that I've created a new PR (#81) and I'm closing this PR. But don't worry because I've reused your commit so you'll get the full credit for your work. Thanks!

PR #75, opened 2015-06-30
Add the app_ prefix to  form field types to avoid collisions

mantulo:
Best Practice
http://symfony.com/doc/current/best_practices/forms.html#custom-form-field-types


stof:
:+1:

javiereguiluz:
@mantulo thanks for this quick win! This is something we should have done from the beginning.

PR #78, opened 2015-07-03
Add missing translations

b-durand:
#8


JulienItard:
We should use yml or xlif translation file, not both.

javiereguiluz:
@b-durand thanks a lot for your amazing work in this PR!!

Before merging it we need some minor changes, such as not using YAML translation files. I've made those changes in #80. Of course I've taken all your commits from this PR so you get the full credit of the changes when that PR is merged. Thanks!

PR #79, opened 2015-07-06
Add docker support

dosten:
Hi!

This PR adds Docker support, so, anyone with a Docker installation can run `$ docker-compose up` to run the application :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM php:5.6
+
+RUN echo "date.timezone = UTC" > /usr/local/etc/php/conf.d/date.ini
+
+VOLUME /app
+WORKDIR /app
+
+ENTRYPOINT ["./app/console"]
+CMD ["server:run", "0.0.0.0:8000"]
+
+EXPOSE 8000
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,6 @@
+web:
+    build: .
+    volumes:
+        - .:/app
+    ports:
+        - "8000:8000"
--- a/README.md
+++ b/README.md
@@ -48,13 +48,24 @@ $ cd symfony-demo/
 $ php app/console server:run
 ```
 
-This command will start a web server for the Symfony application. Now you can
-access the application in your browser at <http://localhost:8000>. You can
-stop the built-in web server by pressing `Ctrl + C` while you're in the
-terminal.
-
 > **NOTE**
 >
 > If you're using PHP 5.3, configure your web server to point at the `web/`
 > directory of the project. For more details, see:
 > http://symfony.com/doc/current/cookbook/configuration/web_server_configuration.html
+
+You can also run the application using Docker, first, install [`docker`](https://docs.docker.com/installation/) (or [`boot2docker`](http://boot2docker.io/)) and
+[`docker-compose`](https://docs.docker.com/compose/install/) then run the following command:
+
+```bash
+$ docker-compose up
+```
+
+> **NOTE**
+>
+> If you are using [`boot2docker`](http://boot2docker.io/), you will have to replace `localhost` with the IP of your
+> boot2docker VM (you can get it by running `boot2docker ip`).
+
+Now you can access the application in your browser at <http://localhost:8000>. You can
+stop the built-in web server or the Docker container by pressing `Ctrl + C` while you're in the
+terminal.


javiereguiluz:
@dosten thanks for sending this improvement. Before merging it, I'll wait for someone with Docker experience to review it. Thanks.

romqin:
I'm agree to add docker.

I think it will be nicer if Dockerfiles files were in a `docker` (or `.docker`) directory.

I use a different Dockerfile for have no warning/error while running `app/check.php`, and also use the port 80:
```
FROM php:5.6

RUN apt-get update && \
    apt-get install -y libicu-dev g++ && \
    docker-php-ext-install intl mbstring opcache && \
    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
    echo "date.timezone = UTC\nshort_open_tag=Off" >> /usr/local/etc/php/conf.d/symfony.ini

VOLUME  /srv
WORKDIR /srv

EXPOSE 80

ENTRYPOINT ["app/console"]
CMD ["server:run", "0.0.0.0:80"]
```

javiereguiluz:
@romqin thanks for proposing this alternative! Now I'll wait for other Docker users to comment about which is the best file to use and about @romqin's idea to create a `docker` directory. Thanks.

bocharsky-bw:
The `Mac` users will have problems to use it (there are [permissions problem](https://github.com/boot2docker/boot2docker/issues/581#issuecomment-116359787) for cache dir that not fixed yet, only workaround exists)

romqin:
@bocharsky-bw I'm on Mac and I don't have this problems. You can set the option `-u` with the user uid.

bocharsky-bw:
@romqin Do you think it will be a right solution? It all will be need to add to docs, it's too specific for separate OS. And what about `Windows` users?

Anyway I vote to move this `Docker` solution to the separate repository as it too specific as for me.

Nicofuma:
For the cache and logs, what we can do is to add a volume container to hold them. This way we will not have any perm issue (but if you  want to see the logs you will have to use docker exec)

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM php:5.6
+
+RUN echo "date.timezone = UTC" > /usr/local/etc/php/conf.d/date.ini
+
+VOLUME /app
+WORKDIR /app
+
+ENTRYPOINT ["./app/console"]
+CMD ["server:run", "0.0.0.0:8000"]
```
jderusse: Please, use the default http port 80, if the end user want to map it with something else, he can do it with `-p 8000:80`

docker-compose.yml
```diff
@@ -0,0 +1,6 @@
+web:
+    build: .
```
jderusse: wath's about moving the Dockerfile in a subdirectory to keep the root folder clean?

docker-compose.yml
```diff
@@ -0,0 +1,6 @@
...
+    volumes:
+        - .:/app
```
jderusse: to improve performances on MacOS and Windows (and fix permission issues), you can define volumes for 
```
 volumes:
    - .:/app
    - /app/app/cache
    - /app/app/logs
```

docker-compose.yml
```diff
@@ -0,0 +1,6 @@
+web:
```
jderusse: you can force the user to the current one by adding the following line, and fixing permission issue for files generated by the container (uploaded data, local session, cache, logs, ...)
```
    user: 1000
```
But it may be worse, when the logged user is not the user 1000 (multiple users sharing the same machine)...
What's about commenting the line and explaining it in the documentation?..

killerwolf:
Is the Dockerfile for demo or dev purposes ?

javiereguiluz:
Do you think it's possible to agree on some common Docker configuration to merge this? Thanks!

Dockerfile
bocharsky-bw: I vote for `8000` port because it used as a default port for `app/console server:run` command.

docker-compose.yml
bocharsky-bw: :+1:

bocharsky-bw:
I plan to use Dockerfile for dev purpose too.

Dockerfile
jderusse: `server:run` is designed to run on the developer's machine which probably have an other webserver listening on port 80.
This issue does not exists in a containers.

docker-compose.yml
jderusse: I usely let the docker-compose.yml file in the root folder (with composer.json, travis.yml, bower.json, etc...)

Dockerfile
bocharsky-bw: Yes, it makes sense. However, I think almost every developer should change `80` to something different because he most probably also has an installed server which already listening `80` on host machine.

romqin: I also think that the server should be run on port 80.

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM php:5.6
+
+RUN echo "date.timezone = UTC" > /usr/local/etc/php/conf.d/date.ini
```
romqin: You should really add this:
```
apt-get update && \
apt-get install -y libicu-dev g++ && \
docker-php-ext-install intl mbstring opcache && \
apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
```

It's for the `app/check.php` and also to add `opcache`

docker-compose.yml
romqin: :+1: for move the `Dockerfile` in a subdirectory (`docker/php/Dockerfile` for example). And keep the `docker-compose.yml` here.

Dockerfile
jderusse: Some developpers don't use the port forwarding to the Host.
But use DNS to directly access the container. It will be great to use the default port.

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+VOLUME /app
```
romqin: I would prefer use `/srv` instead of `/app` because the `/srv` was created for :
> Site-specific data which are served by the system.
[source](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard)

Dockerfile
bocharsky-bw: I suppose that `libicu-dev` needs for `intl`, right? But why we need `g++` here?

romqin: Indeed. `g++` is not required, I removed it.

bocharsky-bw: What about `xdebug`? Could we install it here?

jderusse: :-1:

Dockerfile
bocharsky-bw: I like it :+1:

docker-compose.yml
bocharsky-bw: Can you explain a bit how a `/app/app/cache` and `/app/app/logs` volumes creation could improve performance on MacOS and Windows, please?

jderusse: Docker use feature only avialable on linux (cgroups, namespace, etc...). MacOs and Windows users have to run a VirtualMachine and launch docker inside it (it's Boot2Docker)
When you mount a volume `.:/app`, boot2docker create a sharedFolder by using vboxfs (or NFS with some tricks) wich is really slow.
But you have no choice, you have to share the sources between the Host and the container. Except for files you don't care (.gitignore is a good entrypoint to find this files).

By adding new volume `/app/app/cache` (notice there isn't any `:` => it's not a folder shared between the host and the container), docker will create a new volume for the container (it's the same thing when you use volume's container and `volumes_from`). This volume is not shared with the host, it does not use vboxsf (or NFS) and is much more faster.

bocharsky-bw: @jderusse Wow, thanks for this detailed explanation! It's looks very useful for performance boost :+1:

Dockerfile
bocharsky-bw: What about to create `docker/php/symfony.ini` file and [COPY](https://docs.docker.com/reference/builder/#copy) it to the container instead of echoing with RUN command?

javiereguiluz:
Should we close this issue as "won't fix for now" ... or do you think we are close to some agreement? Thanks.

frastel:
Also I am a big fan of Docker (I hosted a Docker Workshop on the latest Symfony Live in Berlin where I created different Docker environments for the Symfony Demo App), I am of the mind that it's the wrong approach for providing a Docker setup within this project. 
When we start to add the Docker solution than we also need the Vagrant, Puppet, Ansible way or whatsoever. The project would get bound and messed up with provisioning solutions. In my opinion the project should be focused on Symfony and nothing more.
Besides: The proposed Docker solution is too limited. It only creates an environment for PHP 5.6. Other possible and compatible environments are ignored.

theofidry:
@frastel :+1: 

Providing a VM, Docker, Vagrant or whatever will just add complexity to the project, is not really Symfony focused (although it makes sense of course, but it's out of the scope of this project IMO) and make it harder for newbies.

javiereguiluz:
Just asking: would it make sense to write some cookbook about "Symfony + Docker" and use the "Symfony Demo" as an example?

bocharsky-bw:
I'm :+1: for this cookbook! It'll be very interesting, IMO.

theofidry:
> Just asking: would it make sense to write some cookbook about "Symfony + Docker" and use the "Symfony Demo" as an example?

IMO it would be better to take a brand new Symfony standard rather. It would be easier to maintain.

bocharsky-bw:
@frastel It's something that I meant here, but you describe this much better :+1:

dosten:
@frastel :+1: I'm in favor of a cookbook for "Symfony + Docker" so closing..

ivoba:
where does this discussion continue now exactly? can you post a link to the "cookbook", pls.

bocharsky-bw:
@ivoba I think this in the docs repo now https://github.com/symfony/symfony-docs/issues/4055

PR #81, opened 2015-07-06
Add check.php file for access from web

javiereguiluz:
This PR takes the work made by @bocharsky-bw in #74 and applies the look-and-feel of the Symfony Demo application to better integrate it.

### No errors - Before

before_ok

### No errors - After

after-ok

### Errors - Before

before-ko

### Errors - After

after-ko


OskarStark:
i think the symfony logo should be integrated

bocharsky-bw:
@javiereguiluz Thanks a lot! I like it :+1:

bocharsky-bw:
@OskarStark There is no Symfony's logo in `Symfony Demo Application`, so I think it will be unnecessary for this PR

javiereguiluz:
@OskarStark I think the Symfony logo should be reserved for the Symfony installation. This demo is independent from the framework, so I think we can change its design to match our styles.

OskarStark:
yes but the name ```Symfony```is used everywhere...

javiereguiluz:
@OskarStark let me put my reasoning in other words: this app is "made with Symfony" but "it's not Symfony". But let's see if other developers think we should add the logo.

OskarStark:
@javiereguiluz i understand your reasoning, but i think this is an app which shows best practices by the symfony community with standard symfony edition... so a logo makes it more clear.... before this pr there was a symfony logo, or?

javiereguiluz:
@bocharsky-bw I agree. Done. Thanks.

bocharsky-bw:
@OskarStark I think better to create a new issue for this discussion about Symfony's logo in order to other developers could also tell their opinion.

bocharsky-bw:
@javiereguiluz What about to add some instructions to `README.md` about usage of this `check.php` at the beginning?

javiereguiluz:
@bocharsky-bw I like the idea a lot. Done in d88ef7b

OskarStark:
:+1:  for the note

xabbuh:
What about renaming the file to `config.php`? Then its name is consistent with the one from the Standard Edition.

bocharsky-bw:
@xabbuh I think about it too, but there are no any config operations, only environment checks, so it seems to confused a bit as for me.

javiereguiluz:
This feature is now merged. Thank you @bocharsky-bw for starting it and thank you all for the reviews and the discussion!

PR #85, opened 2015-07-07
Make the application name translatable

davialexandre:
While I was doing the Brazilian Portuguese translation I realized that the application name on the page title and on the navbar couldn't be translated. Since it is a descriptive name, I think it would be nice for it to be translatable.

### DIFF
--- a/app/Resources/translations/messages.en.xliff
+++ b/app/Resources/translations/messages.en.xliff
@@ -19,6 +19,11 @@
                 <target>MIT License</target>
             </trans-unit>
 
+            <trans-unit id="app.name">
+                <source>app.name</source>
+                <target>Symfony Demo Application</target>
+            </trans-unit>
+
             <trans-unit id="title.homepage">
                 <source>title.homepage</source>
                 <target><![CDATA[Welcome to the <strong>Symfony Demo Application</strong>]]></target>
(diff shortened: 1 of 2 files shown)


bocharsky-bw:
I don't think it's a good idea. I prefer to left it untranslatable as static project name.

davialexandre:
To me it sounds more like a descriptive name than a brand. Also, it has already been translated to French and Spanish (in the title.homepage message).

javiereguiluz:
I'm also divided about this. Using "Symfony Demo Application" for all languages is better for our branding. But displaying "Symfony Demo Application" in any language different from English may seem weird.

I definitely need to read more opinions from other developers. Let's see what does the community think about this.

javiereguiluz:
After musing about this issue, I think it's better to maintain the "Symfony Demo" name for all languages. However, I also think we can improve the current situation. Right now we call this app the "Symfony Demo Application". We could rename it just to "Symfony Demo" application.

### Before

before_1

### After

after_1

### Before

before_2

### After

after_2

bocharsky-bw:
Great compromise! It's much more clearer. I like it :+1:

javiereguiluz:
The changes have been done in #101.

davialexandre:
Nice! 👍

PR #87, opened 2015-07-08
bump PHP version

xabbuh:
see @Tobion's comment in https://github.com/symfony/symfony-demo/pull/80#discussion_r34105116

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "psr-0": { "": "src/", "SymfonyStandard": "app/" }
     },
     "require": {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I prefer to change the code which is non PHP 5-3 compatible. Keep in mind that the demo is mostly used by newcomers. And those developers usually have legacy PHP versions installed. If you use the Symfony Installer to install the demo, you have 5.4 for sure. But since is possible to install the demo via Composer, I think that the demo should have the same requirements as Symfony itself (if not, we should also change the newly introduced `check.php` script).

PR #91, opened 2015-07-08
Disable html5 validation

hason:


javiereguiluz:
I'm trying to understand the reasons of this change. Could you please elaborate a bit? Thanks.

bocharsky-bw:
I guess hason want to disable html5 validation to show users server validation messages.

bocharsky-bw:
But it will be confused developers because main purpose of this project is show best practices way, so they should not disable html5 validation in his projects.

bocharsky-bw:
What about to add comment at the beginning of form open tag (or I think better in form type) with tip that developers could use  `'novalidate' => 'novalidate' } }` or `'required' => false` attributes if they want to see and check server validation messages?
@hason What do you think?

javiereguiluz:
@bocharsky-bw I like your proposal to add a help note about this. I've done that in #100. Thanks.

javiereguiluz:
Closing it in favor of #100

PR #94, opened 2015-07-08
Pt_BR translation review

davialexandre:
Found a couple of problems while reviewing the translation


javiereguiluz:
These things happen :)  @davialexandre thanks for your quick response

PR #95, opened 2015-07-09
Make missing title on comment error page translatable

hason:

PR #96, opened 2015-07-09
Add missing translations according to PR #95

bocharsky-bw:


javiereguiluz:
@bocharsky-bw that was fast! Thanks :)

PR #98, opened 2015-07-10
Sync Brazilian Portuguese translation

davialexandre:


javiereguiluz:
Thanks @davialexandre.

PR #100, opened 2015-07-13
Added a note about client-side and server-side form validation

javiereguiluz:
This fixes #91.


bocharsky-bw:
I like it :+1:

bocharsky-bw:
But what about also to add this note to the `PostType` and `CommentType`?

javiereguiluz:
Done in c6c9a03

Issue #83, opened 2015-07-07
Update application translations


javiereguiluz:
In #80 @b-durand finished translating the entire Symfony Demo application. Now we need to update all the existing translation files:

* [x] English (in #80)
* [x] French (in #80)
* [x] Spanish (in #82)
* [x] Russian (in #86)
* [x] Ukrainian (in #93)
* [ ] Romanian
* [x] Brazilian Portuguese (in #84)
* [x] German (in #89)
* [x] Czech (in #90)

If you plan to update any of the existing translations, please add a comment to let others know that you are working on that.

And if you want to translate the application into other languages, you are more than welcome to do that. Thanks!

bocharsky-bw:
I could help with `Russian` and `Ukrainian` translations

javiereguiluz:
@bocharsky-bw thanks! I've "assigned" those translations to you so others don't start working on them. But there is no hurry to finish them :)

patricksebastien:
It is nice to have translations but I don't see where we can change it on the demo app. It would be nice to add a dropdown with the supported languages (of course the post will still be in English).

How to implement that?

bocharsky-bw:
I like this idea.

bocharsky-bw:
We need to add a dropdown list with available langs and add locale to each route.
@javiereguiluz what do you think?

JulienItard:
i could work on it if everyone is ok.

javiereguiluz:
I'm OK, but please, don't modify the routes to include the locale. Let's keep it as a query parameter to not complicate things much.

Anyway, let's wait for other people's opinions about this.

bocharsky-bw:
:+1: for query parameter

JulienItard:
Ok, i'll do it. Any more opinions ?

patricksebastien:
Since many people will use this demo app as a learning tool (or even as a template), please be sure to include comments & try to use the best practice for making a i18n website. Merci beaucoup Julien.

JulienItard:
Yep, i'll do my best.

b-durand:
Best pratices is to have locale in the route :wink: [Check out the doc.](http://symfony.com/doc/current/book/translation.html#the-locale-and-the-url)

javiereguiluz:
@b-durand strictly speaking, best practices are explained in the "Symfony Best Practices Book", not in the regular Symfony book ;)

[...]

PR #102, opened 2015-07-15
Dropdown to change locale

JulienItard:
see #83

Someone should add comments in routing.yml because my english it not good enough :)

capture du 2015-07-15 19 52 19


app/Resources/views/base.html.twig
```diff
@@ -50,6 +50,20 @@
...
+                                        <ul class="dropdown-menu locales" role="menu">
+                                            <li><a href="{{ path('blog_index', {_locale:'de'}) }}"><img src="{{ asset('images/flags/de.png') }}" alt="Deutsch"> Deutsch</a></li>
+                                            <li><a href="{{ path('blog_index', {_locale:'en'}) }}"><img src="{{ asset('images/flags/en.png') }}" alt="English"> English</a></li>
+                                            <li><a href="{{ path('blog_index', {_locale:'es'}) }}"><img src="{{ asset('images/flags/es.png') }}" alt="Español"> Español</a></li>
+                                            <li><a href="{{ path('blog_index', {_locale:'fr'}) }}"><img src="{{ asset('images/flags/fr.png') }}" alt="Français"> Français</a></li>
+                                            <li><a href="{{ path('blog_index', {_locale:'ru'}) }}"><img src="{{ asset('images/flags/ru.png') }}" alt="русский"> русский</a></li>
+                                            <li><a href="{{ path('blog_index', {_locale:'uk'}) }}"><img src="{{ asset('images/flags/uk.png') }}" alt="український"> український</a></li>
```
bocharsky-bw: Can you change "український" to "українська", please?

app/config/routing.yml
```diff
@@ -7,6 +7,11 @@
 app:
     resource: @AppBundle/Controller/
     type:     annotation
+    prefix:   /{_locale}
+    requirements: 
+        _locale: en|fr|de|es|cs|ru|uk| 
+    defaults:
+        _locale: en
```
bocharsky-bw: I think we should use `_locale: "%locale%"` in order to use `locale` parameter from `parameters.yml`

app/Resources/views/base.html.twig
```diff
@@ -49,7 +49,22 @@
...
+                                    <ul class="dropdown-menu locales" role="menu">
+                                        <li><a href="{{ path('blog_index', {_locale:'de'}) }}"><img src="{{ asset('images/flags/de.png') }}" alt="Deutsch"> Deutsch</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'en'}) }}"><img src="{{ asset('images/flags/en.png') }}" alt="English"> English</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'es'}) }}"><img src="{{ asset('images/flags/es.png') }}" alt="Español"> Español</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'fr'}) }}"><img src="{{ asset('images/flags/fr.png') }}" alt="Français"> Français</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'ru'}) }}"><img src="{{ asset('images/flags/ru.png') }}" alt="русский"> русский</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'uk'}) }}"><img src="{{ asset('images/flags/uk.png') }}" alt="український"> український</a></li>
```
bocharsky-bw: Can you change "український" to "українська", please?

JulienItard: Done. Google Translate :/

app/config/routing.yml
JulienItard: good idea

weaverryan:
@JulienItard I like the flags, but where are they from? Also, I think we should only add the flags that we're using from now (especially if we know the source, and can find other flags if we need them)

app/config/routing.yml
```diff
@@ -7,6 +7,11 @@
 app:
     resource: @AppBundle/Controller/
     type:     annotation
+    prefix:   /{_locale}
+    requirements: 
+        _locale: en|fr|de|es|cs|ru|uk|
```
weaverryan: One too many `|` at the end

javiereguiluz:
@JulienItard thanks for this. I like it.

My only concern are the flags. I think we should get rid of them. This is one of the classic usability problems, [as explained here](http://www.nngroup.com/articles/flag-problems/). Take for the example the Spanish language: it's official in 21 different countries, but we only show the flag for one specific country.

app/config/routing.yml
bocharsky-bw: I think we should move accepted locales to `parameters.yml` too, for example:
``` yaml
# parameters.yml
parameters:
    locales: en|fr|de|es|cs|ru|uk|
```
So we could easily refers to it in other routes in future.
What do you think?

weaverryan: Actually, it should go into a `parameters` key in `config.yml` in that case (since it doesn't change on a machine-by-machine basis). I think that's a good idea (we'll add a comment explaining it)

bocharsky-bw: @weaverryan The last `|` means that locale could be an empty string

JulienItard:
@weaverryan I use this flag https://dribbble.com/shots/1211759-Free-195-Flat-Flags

app/config/routing.yml
JulienItard: Done.

javiereguiluz:
@JulienItard I'm afraid that's not a good enough reason. Flags cannot be shown in a language selector because countries and languages have nothing to do. In you look around, you'll only see flags when a website lets you select a country, not a language.

app/Resources/views/base.html.twig
chapay: All language names start with upper case, but russian and ukrainian with lower, should not we change it to "Русский" and "Українська"?

JulienItard: Are you sure that cyrillic alphabet have capital letter ?

bocharsky-bw: Yes, it should be changed to  `Русский` and `Українська` :+1:

stof:
please revert the permission changes. Templates and config files have no reason to be executable

JulienItard:
Everything is ok now?

app/Resources/views/base.html.twig
```diff
@@ -49,7 +49,23 @@
...
+                                    <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false"><i class="fa fa-flag"></i> <span class="caret"></span></a>
+                                    <ul class="dropdown-menu locales" role="menu">
+                                        <li><a href="{{ path('blog_index', {_locale:'de'}) }}">Deutsch</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'en'}) }}">English</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'es'}) }}">Español</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'fr'}) }}">Français</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'ro'}) }}">Român</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'ru'}) }}">Pусский</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'uk'}) }}">Yкраїнська</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'cs'}) }}">česká republika</a></li>
```
javiereguiluz: I guess `česká republika` means `Czech Republic` (the country) instead of `Czech` (the language). Is it `Česko` more appropriate?

TomasVotruba: `Česky` would be the correct one.

davialexandre: Could you please add Português Brasileiro (pt_BR) to the dropdown?

javiereguiluz:
I've tested it and from a code point of view, everything looks correct to me.

However, I'd like to suggest some design changes. This is the current design:

before

The first issue I see is that the icon represents a flag, which has nothing to do with languages ;)  The second issue, is that I think it's better to display the current language name. The last issue is that this selector is usually dispalyed at the rightmost part of the page. Something like this:

after-1

It would be nice too if the current language is displayed in bold letters in the language list:

after-2

Lastly, if you don't like the `fa-language` icon, another common alternative is to use `fa-globe`, which FontAwesome also relates to "translation":

after-3

app/Resources/views/base.html.twig
```diff
@@ -49,7 +49,23 @@
...
+                                {# Generate a dropdown with all our locales #}
+                                <li class="dropdown">
+                                    <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false"><i class="fa fa-flag"></i> <span class="caret"></span></a>
+                                    <ul class="dropdown-menu locales" role="menu">
+                                        <li><a href="{{ path('blog_index', {_locale:'de'}) }}">Deutsch</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'en'}) }}">English</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'es'}) }}">Español</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'fr'}) }}">Français</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'ro'}) }}">Român</a></li>
+                                        <li><a href="{{ path('blog_index', {_locale:'ru'}) }}">Pусский</a></li>
```
bocharsky-bw: The first letter in this word is from English alphabet now. Could you replace (copy/paste) it with "Русский"?

app/Resources/views/base.html.twig
```diff
@@ -49,7 +49,23 @@
...
+                                        <li><a href="{{ path('blog_index', {_locale:'uk'}) }}">Yкраїнська</a></li>
```
bocharsky-bw: The first letter in this word is from English alphabet now. Could you replace (copy/paste) it with "Українська"?

bocharsky-bw:
I vote for `fa-language` icon :+1:

JulienItard:
I vote for fa-globe :)

app/Resources/views/base.html.twig
JulienItard: done

app/Resources/views/base.html.twig
JulienItard: Done.

app/Resources/views/base.html.twig
JulienItard: @javiereguiluz @TomasVotruba  Done.
@davialexandre My bad. Done.

JulienItard:
Up

bocharsky-bw:
Yeah, it's a bit overhead. What about to add `Twig` function to return  `locale_code => locale_name` array of all supported locales.

javiereguiluz:
Yes, it would be better to create a Twig global variable from the `locales` parameter and use something like: `for locale in locales|split('|')`

voronkovich:
Parameters do not need to be flat strings, they can also contain array values: http://symfony.com/doc/current/components/dependency_injection/parameters.html#array-parameters. You don't need to use the "split" function.

javiereguiluz:
@voronkovich yes, but we want to define the locales in one single parameter and then, define the global Twig variable based on that parameter:

```yaml
parameters:
    locales: en|fr|de|es|cs|ru|uk|ro|pt_BR

# ...

twig:
    globals:
        locales: %locales%
```

bocharsky-bw:
What about locale human readable names? Maybe somethink like:

``` yaml
twig:
    globals:
        languages:
            en: English
            ru: Russian
            # ... other
```

voronkovich:
@bocharsky-bw, I agree with you. In the template we will need languages names, not only codes.

javiereguiluz:
I was hoping we can define just the locale codes and Symfony will do the magic of displaying the name of each language in its own language (English, Français, Español, etc.)

voronkovich:
@javiereguiluz, are you meaning this? http://symfony.com/doc/current/components/intl.html#country-names

bocharsky-bw:
Yep, `Intl` component could help us, but I think we need [locales](http://symfony.com/doc/current/components/intl.html#locales)

JulienItard:
Good ideas, i'll do it tomorrow.

JulienItard:
What do you think about it ?

bocharsky-bw:
I think locale menu relates to whole project, not only to Blog, so this method in BlogController a bit confused me. Maybe better to use custom Twig function in order to render locale menu?

JulienItard:
Better ?

bocharsky-bw:
I like it more :+1:

src/AppBundle/Twig/LocaleExtension.php
```diff
@@ -0,0 +1,52 @@
...
+            new \Twig_SimpleFunction('locales', array($this, 'getLocales')),
+        );
+    }
+
+    public function getLocales()
+    {
+        $array = explode('|', $this->container->getParameter('locales'));
+
+        foreach ($array as $locale) {
+            $locales[] = ['code' => $locale, 'name' => Intl::getLocaleBundle()->getLocaleName($locale, $locale)];
```
javiereguiluz: We need to use the long array syntax because we still target PHP 5.3. Sorry.

app/config/services.yml
```diff
@@ -18,6 +18,12 @@ services:
         tags:
             - { name: twig.extension }
 
+    app.twig.locale_extension:
+        class:     AppBundle\Twig\LocaleExtension
+        arguments: ["@service_container"]
```
javiereguiluz: We can't inject the entire service container because that's a practice we always discourage. Injecting just the `%locales%` parameters will be enough.

app/config/config.yml
```diff
@@ -9,6 +9,9 @@ imports:
     - { resource: security.yml }
     - { resource: services.yml }
 
+parameters:
+    locales: en|fr|de|es|cs|ru|uk|ro|pt_BR
```
javiereguiluz: I wonder if `locales` is a too generic name. Should we rename it to `enabled_locales`, `active_locales`, `app_locales`, etc.?

bocharsky-bw: You're right, I like `app_locales`

src/AppBundle/Twig/LocaleExtension.php
JulienItard: Done.

app/config/services.yml
JulienItard: Done.

app/config/config.yml
JulienItard: Done.

javiereguiluz:
@JulienItard thanks a lot for implementing this feature, for your patience during the review process and for removing the flags even if you didn't want to do it at first :)

bocharsky-bw:
@JulienItard Thanks a lot!

PR #104, opened 2015-07-20
Display the comment's datetime localized

### LINKED ISSUE #99: Make the comment published date translatable
Right now, the date format is 'M jS \\a\\t g:i a', which doesn't work very well for every language.

I'm not sure about what would be the best approach for this, but I really like the idea of adding a new message to store the date format. With that it would be possible for every language to have its own date format and we could also have distinct date formats for different parts of the application.

I actually tried this, but I found a problem with the approach. Because the application is using the fallback translator option to set the actual locale, it may (if there's not date format to the given locale) end up returning the the message key as the date format. Maybe we could change fallback to something like: 
```yaml
translator:      { fallback: ["%locale%", "en"] }
```
What do you think?

javiereguiluz:
This fixes #99

### DIFF
--- a/app/Resources/views/admin/blog/index.html.twig
+++ b/app/Resources/views/admin/blog/index.html.twig
@@ -19,7 +19,7 @@
             <tr>
                 <td>{{ post.title }}</td>
                 <td>{{ post.authorEmail }}</td>
-                <td>{% if post.publishedAt %}{{ post.publishedAt|date('Y-m-d H:i') }}{% endif %}</td>
+                <td>{% if post.publishedAt %}{{ post.publishedAt|localizeddate('short', 'short', app.request.get('_locale')|default('en')) }}{% endif %}</td>
                 <td>
                     <div class="item-actions">
                         <a href="{{ path('admin_post_show', { id: post.id }) }}" class="btn btn-sm btn-default">
--- a/app/Resources/views/admin/blog/index.html.twig
+++ b/app/Resources/views/admin/blog/index.html.twig
@@ -19,7 +19,7 @@
             <tr>
                 <td>{{ post.title }}</td>
                 <td>{{ post.authorEmail }}</td>
-                <td>{% if post.publishedAt %}{{ post.publishedAt|localizeddate('short', 'short', app.request.get('_locale')|default('en')) }}{% endif %}</td>
+                <td>{% if post.publishedAt %}{{ post.publishedAt|localizeddate('short', 'short') }}{% endif %}</td>
                 <td>
                     <div class="item-actions">
                         <a href="{{ path('admin_post_show', { id: post.id }) }}" class="btn btn-sm btn-default">
--- a/app/Resources/views/blog/post_show.html.twig
+++ b/app/Resources/views/blog/post_show.html.twig
@@ -32,7 +32,7 @@
         <div class="row post-comment">
             <h4 class="col-sm-3">
                 <strong>{{ comment.authorEmail }}</strong> {{ 'post.commented_on'|trans }}
-                <strong>{{ comment.publishedAt|date('M jS \\a\\t g:i a') }}</strong>
+                <strong>{{ comment.publishedAt|localizeddate('medium', 'short', app.request.get('_locale')|default('en')) }}</strong>
             </h4>
             <div class="col-sm-9">
                 {{ comment.content|md2html }}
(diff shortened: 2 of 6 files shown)


app/Resources/views/blog/post_show.html.twig
```diff
@@ -32,7 +32,7 @@
         <div class="row post-comment">
             <h4 class="col-sm-3">
                 <strong>{{ comment.authorEmail }}</strong> {{ 'post.commented_on'|trans }}
-                <strong>{{ comment.publishedAt|date('M jS \\a\\t g:i a') }}</strong>
+                <strong>{{ comment.publishedAt|localizeddate('medium', 'short', app.request.get('_locale')|default('en')) }}</strong>
```
bocharsky-bw: What about filter name with underscore `localized_date`?

bocharsky-bw: Oh, sorry, I think it custom filter.

javiereguiluz: We can't change it :(  We're using the Twig extension filter and it's called like that:

twig_filter

bocharsky-bw:
Should not we add `"ext-intl": "*"` to `composer.json` requirements section?

composer.json
```diff
@@ -23,7 +23,7 @@
         "symfony/monolog-bundle"               : "~2.7",
         "symfony/swiftmailer-bundle"           : "~2.3",
         "symfony/symfony"                      : "~2.7",
-        "twig/extensions"                      : "~1.2"
+        "twig/extensions"                      : "^1.2"
```
bocharsky-bw: Why did you use `^` sign instead of `~`?

javiereguiluz: Because I executed `composer require twig/extensions` without looking first at `composer.json`. We already had that dependency, and Composer changed when executing that `require` command. Should I revert this change?

bocharsky-bw: I think better to revert it, but I don't think that much necessary, the behavior is the same as [~1.2](https://getcomposer.org/doc/articles/versions.md#caret). I only want to be consistent with other dependencies versions definition via tilde `~`.

bocharsky-bw:
@javiereguiluz Could you also localize date on some admin pages:
[blog index page](https://github.com/symfony/symfony-demo/blob/master/app/Resources/views/admin/blog/index.html.twig#L22)
[post show page](https://github.com/symfony/symfony-demo/blob/master/app/Resources/views/admin/blog/show.html.twig#L24)

javiereguiluz:
@bocharsky-bw good idea! I've done it in bb9d1a7. Thanks!

stof:
> Should not we add ``"ext-intl": "*"`` to composer.json requirements section?

not needed. If you don't have the intl extension, the Symfony stub will be used. It will always format it in the en format, but it will work

app/Resources/views/admin/blog/index.html.twig
```diff
@@ -19,7 +19,7 @@
             <tr>
                 <td>{{ post.title }}</td>
                 <td>{{ post.authorEmail }}</td>
-                <td>{% if post.publishedAt %}{{ post.publishedAt|date('Y-m-d H:i') }}{% endif %}</td>
+                <td>{% if post.publishedAt %}{{ post.publishedAt|localizeddate('short', 'short', app.request.get('_locale')|default('en')) }}{% endif %}</td>
```
stof: no need to pass the locale explicitly. The Request object sets the Intl locale properly already

app/config/services.yml
```diff
@@ -24,6 +24,11 @@ services:
         tags:
             - { name: twig.extension }
 
+    app.twig.extension_intl:
+        class: Twig_Extensions_Extension_Intl
```
stof: you should make the service private so that it gets inline when optimizing the container

stof: btw, other Twig extensions could also be marked as private

bocharsky-bw: @stof Does [app.controller_listener](https://github.com/symfony/symfony-demo/blob/master/app/config/services.yml#L33-L37) should be private too?

bocharsky-bw: I opened PR https://github.com/symfony/symfony-demo/pull/108 for other services

xabbuh:
Though we could add the intl extension to the `suggest` section.

PR #106, opened 2015-07-21
Fix homepage locale resetting

bocharsky-bw:
Add locale to `homepage` route to avoid locale resetting on homepage

### DIFF
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -18,7 +18,8 @@ app:
 # the template can be rendered without executing any logic in your own controller.
 # See http://symfony.com/doc/current/cookbook/templating/render_without_controller.html
 homepage:
-    path: /
+    path: /{_locale}
     defaults:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
:+1: it looks good to me.

app/config/routing.yml
```diff
@@ -18,7 +18,8 @@ app:
 # the template can be rendered without executing any logic in your own controller.
 # See http://symfony.com/doc/current/cookbook/templating/render_without_controller.html
 homepage:
-    path: /
+    path: /{_locale}
     defaults:
         _controller: FrameworkBundle:Template:template
         template:    'default/homepage.html.twig'
+        _locale:     "%locale%"
```
bocharsky-bw: The homepage without obvious locale is still work (used locale by default)

javiereguiluz:
Thanks @bocharsky-bw.

PR #109, opened 2015-07-21
Fixed tests

javiereguiluz:
This fixes functional tests in preparation of Travis activation.


app/Resources/views/admin/blog/index.html.twig
```diff
@@ -26,7 +26,7 @@
                             {{ 'action.show'|trans }}
                         </a>
 
-                        {% if post.isAuthor(app.user) %}
+                        {% if app.user and post.isAuthor(app.user) %}
```
stof: isn't it required to be authenticated to access the admin section ?

src/AppBundle/Entity/Post.php
```diff
@@ -133,13 +133,13 @@ public function setAuthorEmail($authorEmail)
     /**
      * Is the given User the author of this Post?
      *
-     * @param User $user
+     * @param User|string $user
```
stof: IMO, the entity should not expect strings here

app/Resources/views/admin/blog/index.html.twig
javiereguiluz: This is not needed. I've removed it. Thanks.

src/AppBundle/Entity/Post.php
javiereguiluz: The original code generated these failing tests:

failed_tests

Should we just remove those two tests?

stof: well, are you authenticated in these tests ?

stof: and if yes, are you authenticated with a User object or no ?

javiereguiluz: We are not authenticated because that is a smoke test that guarantees that all private URLs redirect to the login form if you are unauthenticated.

stof: in any case, only a User object should reach this method call. If you have places where you don't have a User, it should be handled in the code calling this method, not in the Entity. For the entity, it makes perfect sense to expect a User object

stof: Well, can you debug *why* it is not a redirect then ? The failure output does not provide much details

javiereguiluz: This is the original code:

```php
    public function isAuthor(User $user = null)
    {
        return $user->getEmail() == $this->getAuthorEmail();
    }
```

Which generates this error:

before

And this is the new code:

```php
    public function isAuthor($user = null)
    {
        return $user instanceof User && $user->getEmail() == $this->getAuthorEmail();
    }
```

Which correctly redirects to the login form:

after

stof: Well, I would need the stack trace to be able to see where it comes from

stof: and what I'm saying is that the fix would be to stop passing a string to this method, not to accept strings and reject them as author

javiereguiluz:
I've refactored some backend controllers (https://github.com/javiereguiluz/symfony-demo/commit/1cae0a24328f81980f03c7abcbdf9115f4f3f42b) to fix tests while maintaining the `isAuthor()` original code. Tests work again, although Travis has some problems on its own.

bocharsky-bw:
Great work! :+1:

PR #80, opened 2015-07-06
Finish application translation

javiereguiluz:
This fixes #8 thanks to the work made by @b-durand in #78.

PR #116, opened 2015-07-27
Romanian translation

cristi-contiu:
Romanian translations based on #80


javiereguiluz:
@cristi-contiu thanks for updating this translation! *Mulţumesc mult*

PR #117, opened 2015-07-28
$user parameter is not optional

olvlvl:


src/AppBundle/Entity/Post.php
```diff
@@ -139,7 +139,7 @@ public function setAuthorEmail($authorEmail)
      */
     public function isAuthor(User $user = null)
     {
-        return $user->getEmail() == $this->getAuthorEmail();
+        return $user && $user->getEmail() === $this->getAuthorEmail();
```
xabbuh: Should be `null !== $user` according to the Symfony CS.

xabbuh: Though I think it doesn't make sense to allow `null` here.

javiereguiluz: I agree. The `null` doesn't make sense.

olvlvl: Yeah, the ` = null` part should be removed, right?

javiereguiluz: @olvlvl indeed I think it should be removed. Thanks for updating your branch!

olvlvl: Here you go :)

javiereguiluz:
Thank you @olvlvl.

PR #119, opened 2015-07-28
Italian translation

dborsatto:
Italian XLIFF language files added, and config.yml updated.

On a side note, I added "it" after all previous languages, but for readibility purposes I suggest ordering the list alphabetically, with "en" possibly kept at the front.

### DIFF
--- a/app/Resources/translations/messages.it.xliff
+++ b/app/Resources/translations/messages.it.xliff
@@ -29,7 +29,7 @@
             </trans-unit>
             <trans-unit id="title.controller_code">
                 <source>title.controller_code</source>
-                <target>Codice del controller</target>
+                <target>Codice del controllore</target>
             </trans-unit>
             <trans-unit id="title.twig_template_code">
                 <source>title.twig_template_code</source>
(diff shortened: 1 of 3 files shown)


app/Resources/translations/messages.it.xliff
```diff
@@ -0,0 +1,241 @@
...
+                <source>title.homepage</source>
+                <target><![CDATA[Benvenuto nell'applicazione <strong>Symfony Demo</strong>]]></target>
+            </trans-unit>
+            <trans-unit id="title.source_code">
+                <source>title.source_code</source>
+                <target>Codice sorgente usato per effettuare il render di questa pagina</target>
+            </trans-unit>
+            <trans-unit id="title.controller_code">
+                <source>title.controller_code</source>
+                <target>Codice del controller</target>
```
javiereguiluz: In the italian Symfony docs, it's more common to use the word `controllore`. What do you think?

dborsatto: Personally I don't agree much with `controllore`, because to me in Italian it has the sense of strictly "performing controls" or "overseeing", whereas a controller in a Symfony app does more than that, and the foreign term helps conveying that different meaning.
However, in order to keep it consistent with the docs, I can just change that.

dborsatto:
I updated the translation for consistency reason with previous docs.

javiereguiluz:
@dborsatto thanks a lot for this translation and for making the needed changes!

PR #121, opened 2015-07-29
Mistranslation en to de

frag-seb:
This is a small mistranslation. (FÜr to Für)


javiereguiluz:
Thanks @frag-seb.

frag-seb:
you're welcome

xabbuh:
:+1: good catch

PR #122, opened 2015-07-29
pt_br translation to the "label.role" source.

dudemelo:


javiereguiluz:
@dudemelo @davialexandre we have similar doubts in Spanish. Sometimes we translate it to `Rol`, but most of the times we maintain the original `Role` word. I think it's better to maintain the current translation. And please, @dudemelo, don't hesitate to submit new pull request in the future to tweak and improve the translation. Thanks!

PR #123, opened 2015-07-30
Fix localizeddate filter error when php5-intl lib do not installed

bocharsky-bw:
If `locale` is null (by default for [localizeddate](https://github.com/twigphp/Twig-extensions/blob/master/lib/Twig/Extensions/Extension/Intl.php#L46) filter) and `php5-intl` package do not installed - we get [MethodArgumentValueNotImplementedException](https://github.com/symfony/Intl/blob/master/DateFormatter/IntlDateFormatter.php#L148-L150) exception.

So if we pass `app.request.locale` as third argument to `localizeddate` filter we get worked english version. However other langs will thrown exception while `php5-intl` will be installed.

fix-localized-date-filter

### DIFF
--- a/app/Resources/views/admin/blog/index.html.twig
+++ b/app/Resources/views/admin/blog/index.html.twig
@@ -19,7 +19,7 @@
             <tr>
                 <td>{{ post.title }}</td>
                 <td>{{ post.authorEmail }}</td>
-                <td>{% if post.publishedAt %}{{ post.publishedAt|localizeddate('short', 'short') }}{% endif %}</td>
+                <td>{% if post.publishedAt %}{{ post.publishedAt|localizeddate('short', 'short', app.request.locale) }}{% endif %}</td>
                 <td>
                     <div class="item-actions">
                         <a href="{{ path('admin_post_show', { id: post.id }) }}" class="btn btn-sm btn-default">
--- a/app/Resources/views/admin/blog/show.html.twig
+++ b/app/Resources/views/admin/blog/show.html.twig
@@ -21,7 +21,7 @@
             </tr>
             <tr>
                 <th>{{ 'label.published_at'|trans }}</th>
-                <td><p>{{ post.publishedAt|localizeddate('long', 'medium') }}</p></td>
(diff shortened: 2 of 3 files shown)


stof:
This looks like a bug in the stub intl implementation. The default behavior when passing ``null`` is to use the default locale, and ``Locale::getDefault()`` is stubbed. so it would make sense to support this properly.

and twig-extension relies on the fact that the default locale is used by default, avoiding to make this choice explicitly: https://github.com/twigphp/Twig-extensions/pull/94

bocharsky-bw:
@stof I don't understand where default locale injecting to `IntlDateFormatter`. The `twig_localized_date_filter` filter creates formatter:

``` php
    $formatter = IntlDateFormatter::create(
        $locale,
        $formatValues[$dateFormat],
        $formatValues[$timeFormat],
        $date->getTimezone()->getName(),
        IntlDateFormatter::GREGORIAN,
        $format
    );
```
and if we don't manually pass locale  then `$locale = null`. So creation is always fails because in constructor we had:

``` php
        if ('en' !== $locale) {
            throw new MethodArgumentValueNotImplementedException(__METHOD__, 'locale', $locale, 'Only the locale "en" is supported');
        }
```

because `$locale = null` in this case.

stof:
no we should not. We should fix the stub implementation to consider that ``null`` as locale means using the default intl locale (and so ``en`` in the case of the stub), to match the ext-intl behavior

stof:
and btw, this change is actually wrong. It will break things when not having intl and trying a different locale than ``en``. I'm currently writing a patch for Symfony to fix the root issue.

bocharsky-bw:
@stof thanks!

stof:
The fix is now available in Symfony. It will be part of the upcoming 2.7.4 release (and also of the upcoming 2.3.32 release)

bocharsky-bw:
Thanks, this fix work for all locales. I think we could close this PR now or after `2.7.4` will be released.

javiereguiluz:
Closing this issue as per @bocharsky-bw comments.

PR #124, opened 2015-07-30
Fix web check filename

bocharsky-bw:
We renamed `web/config.php` to `web/check.php` in https://github.com/symfony/symfony-demo/pull/74

### DIFF
--- a/app/check.php
+++ b/app/check.php
@@ -68,7 +68,7 @@ echo_style('title', '~~~~');
 echo '  than the one used with your web server. To be on the'.PHP_EOL;
 echo '      safe side, please check the requirements from your web'.PHP_EOL;
 echo '      server using the ';
-echo_style('yellow', 'web/config.php');
+echo_style('yellow', 'web/check.php');
 echo ' script.'.PHP_EOL;
 echo PHP_EOL;
 


javiereguiluz:
Good catch, thanks @bocharsky-bw.

bocharsky-bw:
@javiereguiluz Wow, sorry, but it's look like this file override every time when we run `Composer`. The source file located in [app/check.php](https://github.com/sensiolabs/SensioDistributionBundle/blob/78faf147315523aaa5c3fefbe6a0e57d2f6745d3/Resources/skeleton/app/check.php#L71)

Maybe better live `app/config.php`? What do you think?

stof:
Well, this file is updated from SensioDistributionBundle by a composer hook. So there is 2 solutions:

- keep the file unmodified
- remove the composer hook updating the requirement checker

bocharsky-bw:
I hesitate what would be better.

xabbuh:
Well, the hook does not only install the `app/check.php` file, but also updates the `app/SymfonyRequirements.php` file which contains the actual rules which can be more often subject to changes. Changing the file in the `web` directory is possibly the better solution.

bocharsky-bw:
Maybe we should rename `web/check.php` to `web/config.php` to be consistent with Symfony Standard?

xabbuh:
@bocharsky-bw Imo that's the best solution.

bocharsky-bw:
I agree with you.
@javiereguiluz What do you think?

PR #125, opened 2015-07-30
Fixing post content constraint

andesk:
Post content field is not nullable, so we have to validate against NotBlank constraint.

Sadly, this validation message misses now all other translations... Will need others to fix that afterwards, I guess.

### DIFF
--- a/src/AppBundle/Entity/Post.php
+++ b/src/AppBundle/Entity/Post.php
@@ -53,6 +53,7 @@ class Post
 
     /**
      * @ORM\Column(type="text")
+     * @Assert\NotBlank(message="post.blank_content")
      * @Assert\Length(min = "10", minMessage = "post.too_short_content")
      */
     private $content;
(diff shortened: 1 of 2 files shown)


andesk:
Thanks to @bicpi for encouraging me to do my first PR to a symfony repository :)

bocharsky-bw:
yes, we get error on server side if html5 validation disabled :+1:

javiereguiluz:
@andesk thanks for this improvement and congratulations on your first Symfony code contribution! Thanks to @bicpi for encouraging you :)

src/AppBundle/Entity/Post.php
```diff
@@ -53,6 +53,7 @@ class Post
 
     /**
      * @ORM\Column(type="text")
+     * @Assert\NotBlank(message="post.blank_content")
```
bocharsky-bw: I think we could use `post.too_short_content` in this assert too

javiereguiluz: You are right. That will prevent us from having to update all translations. I've done that in https://github.com/symfony/symfony-demo/commit/d68caa051ae7717b4ddb0e61650fa78bad226ab9 to merge this contribution "as is". Thanks!

andesk: Good point! :)

PR #127, opened 2015-07-31
Remove point to PHP platform from composer.json

bocharsky-bw:
The main problem is to have `v5.6.4` PHP version on my local machine I can't add other bundles depend of PHP `>v5.3.9` (it points in platform section of `composer.json`) and I have next error when doing `$ composer require knplabs/doctrine-behaviors:dev-master`:
```
  Problem 1
    - Installation request for knplabs/doctrine-behaviors dev-master -> satisfiable by knplabs/doctrine-behaviors[dev-master].
    - knplabs/doctrine-behaviors dev-master requires php >=5.4 -> your PHP version (5.6.4-4ubuntu6.2) does not satisfy that requirement.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "ext-pdo_sqlite"                       : "*",
         "doctrine/doctrine-bundle"             : "~1.5",
         "doctrine/doctrine-fixtures-bundle"    : "~2.2",
-        "doctrine/orm"                         : "~2.4",
+        "doctrine/orm"                         : "~2.4.7",
         "erusev/parsedown"                     : "~1.5",
         "incenteev/composer-parameter-handler" : "~2.1",
         "ircmaxell/password-compat"            : "~1.0",
@@ -47,10 +47,7 @@
         ]
     },
     "config": {
-        "bin-dir": "bin",
-        "platform": {
-            "php": "5.3.9"
-        }
+        "bin-dir": "bin"
     },
     "extra": {
         "symfony-app-dir": "app",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "252b684b5ecfedd66558dbf523c883aa",
+    "hash": "58dc2c2379791cfbcde5f43e582bcfe4",
     "packages": [
         {
             "name": "doctrine/annotations",
@@ -1980,8 +1980,5 @@
         "php": ">=5.3.9",
         "ext-pdo_sqlite": "*"
     },
-    "platform-dev": [],
-    "platform-overrides": {
-        "php": "5.3.9"
-    }
+    "platform-dev": []
 }


javiereguiluz:
If we remove the `platform` information, some tests fails due to Doctrine. The 2.5 version of the ORM requires PHP 5.4, which breaks this application for PHP 5.3.

bocharsky-bw:
@javiereguiluz Yes, I see tests fails. I try to fix it over dependencies versions.
But I don't like this `platform` hack and believe we can rid it of. Lets see if I can do it.

TomasVotruba:
@bocharsky-bw :+1:

stof:
this is not a platform hack. It is a way to tell composer that it should resolve dependencies compatible with PHP 5.3.9 rather than with your system PHP. The error message for mismatching php version is broken in composer though (it displays the wrong one)

stof:
And changing the Doctrine requirement is the wrong way to ensure PHP 5.3 compatibility. Any other package could bump the requirement in newer versions

javiereguiluz:
> It is a way to tell composer that it should resolve dependencies compatible with PHP 5.3.9 rather than with your system PHP.

My question is: why does Travis in PHP 5.3 environemnt installs Doctrine 2.5 which requires PHP 5.4?

stof:
@javiereguiluz it installs what is locked in the lock file

bocharsky-bw:
The [symfony/symfony-standard](https://github.com/symfony/symfony-standard) don't use `platform` to test for PHP `>=5.3.9.` compatibility, so why `symfony/symfony-demo` should?
Anyway, with `~2.4.7` of `doctrine/orm` dependency all tests passed.

javiereguiluz:
@stof so we shouldn't provide a `.lock` file because we want this application to work everywhere and we don't want to lock the specific versions.

javiereguiluz:
@bocharsky-bw but why should we limit to `~2.4.7`. Most people use 5.4, 5.5 or 5.6 ... they should enjoy Doctrine 2.5.*

stof:
@bocharsky-bw given that symfony-standard is the skeleton for user projects, configuring a PHP platform would be a bad idea (the meaningful value for your own project would be the PHP version used on your prod server, and we don't know it).

@javiereguiluz the fact that we provide an application is precisely the reason why we should provide a lock file and use the platform config to force the resolution to be compatible with our lowest PHP requirement. This is what composer does.
The weird things is the fact that we change the dependencies on Travis instead of testing the locked ones. It does not make sense for a project.

And @bocharsky-bw had an issue when trying to add a new package in the demo, because it is not compatible with PHP 5.4. this is totally expected. As we want to keep the demo compatible with PHP 5.3, we **want** you to have such error when you try to add this incompatible package.

javiereguiluz:
OK. So then, we can close this issue as "won't fix" ?

bocharsky-bw:
So why we need to keep compatible with PHP `5.3.9` if we can't add any libs required `>5.3.9`.
IMO, it more hinders progress then we get something useful from compatibility.
It turns out that the more correct way to develop custom things based on `symfony-demo` is to manually remove `platform` and require whatever libs you want, right?

javiereguiluz:
@bocharsky-bw Symfony 3 requires `"php": ">=5.5.9"`. When this version is released at the end of November, we'll make the switch in this Symfony Demo application.

bocharsky-bw:
@javiereguiluz Yes, thanks. 
I test some libraries using symfony demo application because it already has a sandbox and it's a very convenience for me but I stumbled on this problem and don't do this changes each time manually

stof:
@bocharsky-bw if you create your own project, you can change your own PHP requirement to drop the PHP 5.3 support. but the policy for the symfony demo is to support all PHP versions supported by Symfony.

@javiereguiluz given that the installer actually requires PHP 5.4+, we may choose to require 5.4+ in the demo too. What do you think ?

javiereguiluz:
@stof I'd prefer to not make that change. Reasons:

  * Installer requires 5.4 because we have no other choice if we want to display the cool download progressbar that you developed. I wish we could use 5.3 there too.
  * The Installer was advertised as 5.4 app since day one and the installer was advertised as a 5.3 app. I prefer to maintain those promises until the massive 5.5 update that will introduce Symfony 3.

PR #130, opened 2015-08-02
Update Symfony up to latest v2.7.3 and update dependencies

bocharsky-bw:


javiereguiluz:
Nice! Thank you @bocharsky-bw

PR #131, opened 2015-08-02
Removed not used namespace

mantulo:
Removed not used namespace


bocharsky-bw:
:+1:

javiereguiluz:
Good catch, thanks @mantulo.

PR #134, opened 2015-08-09
fix #133: XSS injection

### LINKED ISSUE #133: XSS injection in comments, post content and summary
It is possible to exploit XSS injection in comments, post content and summary. All you need is to post this content into forms:

```
<script>alert(123);</script>
```

e-moe:
Q |	A
---------|---------
Bug fix? |	yes
New feature? |	no
BC breaks?	| no
Deprecations? |	no
Tests pass? |	yes
Fixed tickets |	#133 
License |	MIT
Doc | PR

### DIFF
--- a/src/AppBundle/Twig/AppExtension.php
+++ b/src/AppBundle/Twig/AppExtension.php
@@ -36,7 +36,7 @@ class AppExtension extends \Twig_Extension
     public function getFilters()
     {
         return array(
-            new \Twig_SimpleFilter('md2html', array($this, 'markdownToHtml'), array('is_safe' => array('html'))),
+            new \Twig_SimpleFilter('md2html', array($this, 'markdownToHtml'), array('pre_escape' => 'html', 'is_safe' => array('html'))),
         );
     }
 


stof:
See https://michelf.ca/blog/2010/markdown-and-xss/ for the explanation about why this is not solving the issue

javiereguiluz:
For those who don't want to read the entire article:

> So the conclusion is that, if you want real security, you need to filter Markdown’s output, not the input. There’s no other choice.

e-moe:
@javiereguiluz oh, but Markdown’s output is already html. it will be not so easy to filter it

bocharsky-bw:
Does a post XSS injection really problem for us in this case? Posts could add administrator only, so if he need to use html tag - that's ok I think. User haven't enough premissions and couldn't create a posts.

What about comments, IMO we should use `strip_tags` or `htmlentities` and do it only on create/update comment, then we read already escaped value from DB. Require a vendor bundle for it in this demo isn't a good idea I think.

voronkovich:
@bocharsky-bw you're right, `strip_tags` could solve the problem

stof:
@bocharsky-bw @voronkovich no it won't. Please read the article I sent

@bocharsky-bw we should not advocate that Markdown is output a safe value. And allowing XSS in admin forms because you are trusting admins is a bad idea. you should not trust users.

e-moe:
Closing it for not being the right solution.

PR #135, opened 2015-08-09
Enable content escaping in the Markdown extension

javiereguiluz:
This prevents XSS and other issues reported by @e-moe in #133.

### DIFF
--- a/src/AppBundle/Utils/Markdown.php
+++ b/src/AppBundle/Utils/Markdown.php
@@ -25,7 +25,10 @@ class Markdown
 
     public function __construct()
     {
-        $this->parser = new \Parsedown();
+        $parsedown = new \Parsedown();
+        $parsedown->setMarkupEscaped(true);
+
+        $this->parser = $parsedown;
     }
 
     public function toHtml($text)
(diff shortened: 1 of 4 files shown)


e-moe:
thank for fast response! I think it's better to use `pre_escape` option in twig extension (see http://twig.sensiolabs.org/doc/advanced.html#automatic-escaping)

see my initial pull request #134 (already updated)

src/AppBundle/Utils/Markdown.php
```diff
@@ -25,7 +25,10 @@ class Markdown
 
     public function __construct()
     {
-        $this->parser = new \Parsedown();
+        $parsedown = new \Parsedown();
```
e-moe: it's not related to original bug but it's looking weird to have `new` in construct. maybe we can define `Parsedown` it as service and use DI to pass it here as a constructor parameter?

bocharsky-bw: For simplicity, the `AppBundle\Utils\Markdown` is already service wrapper for `Parsedown` and to have `new` in its constructor is acceptable IMO.

e-moe: ok, I see. thanks for explanation

javiereguiluz:
Closing it for not being the right solution.

PR #136, opened 2015-08-10
Fix slug bug when using non alphabet title

bravewood:
If you create a post using non alphabet title, the slug of this new post will be set empty.
http://127.0.0.1:8000/en/admin/post/new

For example:
In case I use 'русский язык' for a title of a new post, the post will be create with no error.
But, the slug will be set empty.
Therefore, indexAction in BlogController will return an error like this:
An exception has been thrown during the rendering of a template ("Parameter "slug" for route "blog_post" must match "[^/]++" ("" given) to generate a corresponding URL.") in blog/index.html.twig at line 9.

This pull request fix the problem.

### DIFF
--- a/src/AppBundle/Tests/Utils/SluggerTest.php
+++ b/src/AppBundle/Tests/Utils/SluggerTest.php
@@ -44,6 +44,12 @@ class SluggerTest extends \PHPUnit_Framework_TestCase
             array(' lOrEm  iPsUm  ' , 'lorem-ipsum'),
             array('!Lorem Ipsum!'   , 'lorem-ipsum'),
             array('lorem-ipsum'     , 'lorem-ipsum'),
+            array('русский язык'    , 'русский-язык'),
+            array('  русский язык  ', 'русский-язык'),
+            array('日本語'           , '日本語'),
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@bravewood thanks for reporting this error. The problem of the proposed solution is that it uses `mb_*()` functions and we don't want to make them a mandatory requirement for this project. Maybe it's time to use a real slugger to avoid any issue? (for example this one: https://github.com/javiereguiluz/easyslugger ^.^)

bravewood:
@javiereguiluz thanks for replying me. I agree with you. I understand the slugging contains a difficult problem. My PR is not a perfect solution for this problem, so it is necessary to use a third party slug library to avoid this problem. I think it's better that Symfony demo uses a slug library than my PR.

stof:
a solution could be to rely on https://github.com/Behat/Transliterator/ to make the transliteration

bravewood:
I have no idea which library to be selected for this project.
If you could choose one slug library, I'd change this PR.
Thanks.

javiereguiluz:
I'm closing this in favor of #389. I'm sorry it took us so long to fix this.

PR #137, opened 2015-08-11
Add requirements for _locale parameter on homepage

bocharsky-bw:
Allow only availiable locales for `_locale` parameter in `homepage` route


javiereguiluz:
:+1:

javiereguiluz:
Thank you @bocharsky-bw.

PR #139, opened 2015-08-12
Move independent of env parameter to config.yml

bocharsky-bw:
I think better to move this parameter to `app/config/config.yml` assuming that we added there parameters which independent of environment to locate them in one place.


javiereguiluz:
With the recent merges, this PR needs a rebase. I'm sorry!

bocharsky-bw:
No problem, I'll rebase it

javiereguiluz:
Thanks @bocharsky-bw.

PR #141, opened 2015-08-12
Enforce best pratices for form button

b-durand:
See http://symfony.com/doc/current/best_practices/forms.html#form-button-configuration.

One more thing: fix missing translation for a hidden label in the same view.

### DIFF
--- a/app/Resources/views/blog/_comment_form.html.twig
+++ b/app/Resources/views/blog/_comment_form.html.twig
@@ -17,7 +17,7 @@
         <legend>{{ 'title.add_comment'|trans }}</legend>
 
         <div class="form-group{% if not form.vars.valid %} has-error{% endif %}">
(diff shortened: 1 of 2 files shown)


bocharsky-bw:
Yes, it's true :+1: . However I like form buttons because it allows to do in controllers something like:

``` php
// create/update action
if ($form->get('save_and_close')->isClicked()) {
    return $this->redirectToRoute('post');
}

return $this->redirectToRoute('post_edit', ['id' => $entity->getId()]);
```

voronkovich:
I think you should remove the `createCommentForm` method, because it's unnecessary

bocharsky-bw:
Yes, to be consistent with other forms creation in this demo

javiereguiluz:
It took a while but this is finally merged. Thanks @b-durand

PR #142, opened 2015-08-12
Register DoctrineFixturesBundle in the env bundles

bocharsky-bw:
Move DoctrineFixturesBundle registration needs for developing only to the environment bundles section


javiereguiluz:
A minor but nice change! Thanks.

PR #143, opened 2015-08-13
Replace query builder with DQL

bocharsky-bw:
The `select` and `from` methods are optional

### DIFF
--- a/src/AppBundle/Repository/PostRepository.php
+++ b/src/AppBundle/Repository/PostRepository.php
@@ -26,13 +26,15 @@ class PostRepository extends EntityRepository
 {
     public function findLatest($limit = Post::NUM_ITEMS)
     {
-        return $this
-            ->createQueryBuilder('p')
-            ->select('p')
-            ->where('p.publishedAt <= :now')->setParameter('now', new \DateTime())
-            ->orderBy('p.publishedAt', 'DESC')
+        return $this->getEntityManager()
+            ->createQuery('
+                SELECT p
+                FROM AppBundle:Post p
+                WHERE p.publishedAt <= :now
+                ORDER BY p.publishedAt DESC
+            ')
+            ->setParameter('now', new \DateTime())
             ->setMaxResults($limit)
-            ->getQuery()
             ->getResult()
         ;
     }


javiereguiluz:
Even if this method call is not strictly necessary, don't you think it improves code readability? Removing it looks like "magic" and Symfony always prefer "explicit". What do you think?

bocharsky-bw:
Yes, it looks a bit magically. However with `select` there is still without `from`, I think it could a bit confused too. As for me I don't like use `select` when fetch all entity.

I based this PR on [Querying for Objects Using Doctrine's Query Builder](http://symfony.com/doc/current/book/doctrine.html#querying-for-objects-using-doctrine-s-query-builder) but it's a tutorial, not best practices.

Also I surfed knpuniversity, Ryan don't use `select` with query builder in his [screencasts](http://knpuniversity.com/screencast/doctrine-queries/query-builder#building-the-query).

Do you think we should keep `select`?

JHGitty:
:+1: I think we should remove ``select``.

xabbuh:
@javiereguiluz Why is the query builder used at all in this example? Imho we can simply use a plain DQL query here as the query is not built based on some argument.

javiereguiluz:
@xabbuh yes, I think DQL should be used in this example. By the way, anyone knows a good reference in the official Doctrine documentation that explains when to use each of their different ways to build queries? I'd like to add a help note linking to that article. Thanks.

bocharsky-bw:
@javiereguiluz I think [Symfony docs](http://symfony.com/doc/current/book/doctrine.html#querying-for-objects) has nice explanation about quering with DQL vs Query Builder.

javiereguiluz:
Honestly, I find the explanation very lacking. For example, when they explain the "query builder": 

> Instead of writing a DQL string, you can alternatively use a helpful object called the QueryBuilder to build that string for you:

But they never say why the query builder is better or worse and when to use it or avoid it.

bocharsky-bw:
It'll be nice to improve Symfony docs for this.

bocharsky-bw:
The minor problem is that in PhpStorm I have autocompletion with Query Builder what I haven't with DQL :(

src/AppBundle/Repository/PostRepository.php
```diff
@@ -26,13 +26,15 @@ class PostRepository extends EntityRepository
...
-            ->select('p')
-            ->where('p.publishedAt <= :now')->setParameter('now', new \DateTime())
-            ->orderBy('p.publishedAt', 'DESC')
+        return $this->getEntityManager()
+            ->createQuery("
+                SELECT p
+                FROM AppBundle:Post p
+                WHERE p.publishedAt <= :now
+                ORDER BY p.publishedAt DESC
+            ")
```
xabbuh: Please use single quotes instead of double quotes.

bocharsky-bw: Ok. I'm interesting why does it better?

xabbuh: It's the Symfony code style which we should imho follow here too.

bocharsky-bw: thanks for this explanation, I replaced quotes

javiereguiluz:
:+1:

bocharsky-bw:
@javiereguiluz What do you think about add next explanation?

> Instead of writing a DQL string, you can alternatively use a helpful object called the QueryBuilder to build that string for you if you need to build a query based on some result of conditions. The QueryBuilder allow you to avoid manual query string concatenation by using suitable methods calls and do this concatenation for you. Also, with QueryBuilder you could easy take away some parts of query to separate methods and reuse it in other queries. Check more information about querying for objects in http://symfony.com/doc/current/book/doctrine.html#querying-for-objects

javiereguiluz:
@bocharsky-bw I think it's a very good starting point to make a pull request to the symfony-docs repo. My only concern is that phrases are too long :)  But please, make the pull request and our cool doc managers will guide you to tweak the original text. Thanks!

bocharsky-bw:
I created PR https://github.com/symfony/symfony-docs/pull/5637

javiereguiluz:
Thank you @bocharsky-bw.

rasanga:
DQL has better performance over Query Builder?

stof:
@rasanga the query builder will ultimately produce the DQL string. So if the query is not dynamic, you only perform extra work for no additional value if you build the string dynamically with many method calls.
The difference should be small though in most cases. And as soon as your DQL query becomes dynamic (adding some condition in the WHERE clause only in some cases for instance, or changing the sorting), it becomes much more maintainable to use the query builder.

PR #145, opened 2015-08-13
Remove published date creation repeatedly

bocharsky-bw:
We already create it in [AppBundle\Entity\Comment::__construct](https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Entity/Comment.php#L70)


javiereguiluz:
Minor but nice change! Thanks @bocharsky-bw

PR #146, opened 2015-08-13
fix #133 XSS injection (with Htmlpurifier)

### LINKED ISSUE #133: XSS injection in comments, post content and summary

e-moe:
Q |	A
---------|---------
Bug fix? |	yes
New feature? |	no
BC breaks?	| no
Deprecations? |	no
Tests pass? |	yes
Fixed tickets |	#133 
License |	MIT
Doc | PR

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
         "doctrine/doctrine-fixtures-bundle"    : "~2.2",
         "doctrine/orm"                         : "~2.4",
         "erusev/parsedown"                     : "~1.5",
+        "ezyang/htmlpurifier"                  : "~4.7",
         "incenteev/composer-parameter-handler" : "~2.1",
         "ircmaxell/password-compat"            : "~1.0",
         "leafo/scssphp"                        : "~0.1.5",
--- a/src/AppBundle/Utils/Markdown.php
+++ b/src/AppBundle/Utils/Markdown.php
@@ -23,15 +23,19 @@ class Markdown
 {
     private $parser;
 
+    private $purifier;
+
     public function __construct()
     {
         $this->parser = new \Parsedown();
+        $this->purifier = new \HTMLPurifier();
     }
 
     public function toHtml($text)
     {
         $html = $this->parser->text($text);
+        $safeHtml = $this->purifier->purify($html);
 
-        return $html;
+        return $safeHtml;
     }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "252b684b5ecfedd66558dbf523c883aa",
+    "hash": "04755c35eaa37511db107f6bb4f50ca0",
     "packages": [
         {
             "name": "doctrine/annotations",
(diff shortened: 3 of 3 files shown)


src/AppBundle/Twig/AppExtension.php
```diff
@@ -28,9 +29,12 @@ class AppExtension extends \Twig_Extension
 {
     private $parser;
 
-    public function __construct(Markdown $parser)
+    private $purifier;
+
+    public function __construct(Markdown $parser, HTMLPurifierExtension $purifier)
```
stof: injecting a Twig extension into another one does not make sense. you should inject HtmlPurifier itself

stof: and I think this should be done in the Markdown class itself, by making ``AppBundle\Utils\Markdown`` a safe markdown parser

e-moe: ok, thanks for comment. I'll try to update this pull request

e-moe:
@stof , thanks for fast response. your comments applied ;)

composer.json
```diff
@@ -13,6 +13,7 @@
         "doctrine/doctrine-fixtures-bundle"    : "~2.2",
         "doctrine/orm"                         : "~2.4",
         "erusev/parsedown"                     : "~1.5",
+        "ezyang/htmlpurifier"                  : "~4.0",
```
stof: please make it ``~4.7`` instead. We don't want older versions (4.6 makes messy things with permissions in its cache folder)

e-moe: done

e-moe:
@stof , sorry for disturbing. pull request is updated due to all your notes :)

src/AppBundle/Utils/Markdown.php
```diff
@@ -23,15 +23,19 @@ class Markdown
 {
     private $parser;
 
+    private $purifier;
+
     public function __construct()
     {
         $this->parser = new \Parsedown();
+        $this->purifier = new \HTMLPurifier();
```
bocharsky-bw: Could you move `HTMLPurifier` creation out to a separate class and define it as a service? I think we could create `AppBundle\Utils\Purifier` class for it and then we could inject it into the `Markdown` service. Also, in this case, we could easily use `Purifier` as a standalone service in any part of our demo application.

javiereguiluz: Before doing that, are we sure we'll use the Purifier service anywhere else in the application?

bocharsky-bw: Anyway, I thought it'll be clearer to get separate service for purifier.

Do you think it'll be a bit overhead for available services?

javiereguiluz: I agree with you if we needed to purify other contents in the application. But for now, it doesn't look like we need it. So I prefer to not create that service.

bocharsky-bw: OK, then we could refactor and create it when it's needed.

javiereguiluz:
Thanks for working on this feature @e-moe

stof:
Given you don't use the bundle anymore, you should configure HtmlPurifier to write its cache inside the Symfony cache (where we are sure we can write)

e-moe:
ok, I'll try to create one more pull request to fix HtmlPurifier  cache

PR #147, opened 2015-08-14
Add underscore prefix to the template that not extend base one

bocharsky-bw:
It seems we add "_" prefix to the name of all templates that not extend the base one.


javiereguiluz:
The `_` is added to any template fragment which is included in other templates. So I'm :+1: to this change.

bocharsky-bw:
Yes, it make sense. Ryan use it in his [screencasts](http://knpuniversity.com/screencast/twig/including-other-templates#using-the-include-function) too

javiereguiluz:
Thanks for all your work on this project!

bocharsky-bw:
I love Symfony! :)

PR #148, opened 2015-08-14
Comment framework.ide setting

nicolas-grekas:
This setting takes priority over any locally configured one, and thus prevents using source links with the demo app.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -24,7 +24,7 @@ framework:
     # Supported values are 'textmate', 'macvim', 'emacs' and 'sublime' shortcuts
     # and any custom configuration string, such as: "phpstorm://open?file=%%f&line=%%l"
     # See http://symfony.com/doc/current/reference/configuration/framework.html#ide
-    ide: sublime
+    #ide: sublime
 
     # esi:             ~
     translator:      { fallback: "%locale%" }


javiereguiluz:
If what you say is true ... then Symfony docs are wrong [in this explanation](http://symfony.com/doc/current/reference/configuration/framework.html#ide):

> This can be done by setting the xdebug.file_link_format in the php.ini configuration to the url string. If this configuration value is set, then the ide option will be ignored.

By the way, this demo application is mostly targeted at newcomers. Most of them don't have Xdebug installed and/or enabled, oso I think it makes sense to enable this option by default. Let's see what others think about this.

nicolas-grekas:
You're right that there is something suspicious. Looking quickly (too quickly?) at the code, the doc looks wrong (or the code).
The demo app is not only for newcomers. Trainers use it a lot also. I do use it for demoing.
And "sublime" is wrong for many (even newcomers) that use e.g. phpstorm.

javiereguiluz:
Taking a look at [this code](https://github.com/symfony/symfony/blob/2.8/src/Symfony/Bridge/Twig/Extension/CodeExtension.php):

```php
    public function __construct($fileLinkFormat, $rootDir, $charset)
    {
        $this->fileLinkFormat = $fileLinkFormat ?: ini_get('xdebug.file_link_format') ?: get_cfg_var('xdebug.file_link_format');
        // ...
    }
```

It seems that the explanation is wrong ... although the described behavior seems better than the current one.

bocharsky-bw:
Yes, it's true. `framework.ide` override `xdebug.file_link_format` value

bocharsky-bw:
I think we shoud change this code and keep explanation in docs as is, right?

javiereguiluz:
@bocharsky-bw I agree that it makes sense that your local settings override anything else. The code should be changed. I don't know if that could be considered a BC break.

javiereguiluz:
I'm closing this issue as "won't fix" because Nicolas did something even better: he fixed an issue in Symfony that prevented you local settings from overriding the application settings. See https://github.com/symfony/symfony/pull/15549.

PR #149, opened 2015-08-17
Remove parameters.yml from repository and add to gitignore

bocharsky-bw:
I also add `--no-interaction` argument to the installation to avoid any interactive question in console if this project will be install with Composer not Symfony installer in order to get work project out-of-the-box.

Intend to close issue https://github.com/symfony/symfony-demo/issues/42


javiereguiluz:
Are we 100% sure that if we merge this change the application will keep working perfectly without having to execute any command or create (copy/paste) any file? Thanks.

bocharsky-bw:
The only problem I see is if a developer already have cloned project and pull this new changes he loses his `parameter.yml` file. So he should do `$ composer install`.

javiereguiluz:
I've tested all the building process and there should be no problem by adding this improvement. Thanks @bocharsky-bw!

PR #150, opened 2015-08-17
Remove unnecessary docblock comments

voronkovich:


bocharsky-bw:
Why do you want to delete docblock comments?

voronkovich:
http://phpdoc.org/docs/latest/guides/inheritance.html#methods

bocharsky-bw:
So I think we should to use `@inheritdoc` instead

voronkovich:
No, we shouldn't, because https://github.com/symfony/symfony/issues/13696

bocharsky-bw:
But Symfony doesn't implement this yet

theofidry:
As in symfony/symfony#13696, until the PSR is out, it may be better to keep using it like this.

voronkovich:
But, why we need this docblock comments?

src/AppBundle/Form/CommentType.php
```diff
@@ -54,9 +47,6 @@ public function configureOptions(OptionsResolver $resolver)
         ));
     }
 
-    /**
-     * @return string
-     */
     public function getName()
```
TomasVotruba: Here is removed valuable information. I need to find `return` statement otherwise.

src/AppBundle/Form/PostType.php
```diff
@@ -64,9 +57,6 @@ public function configureOptions(OptionsResolver $resolver)
         ));
     }
 
-    /**
-     * @return string
-     */
     public function getName()
```
TomasVotruba: Same here.

src/AppBundle/Form/CommentType.php
bocharsky-bw: It's inherited from `AbstractType`

voronkovich: To be more precise, it's inherited from [FormTypeInterface](https://github.com/symfony/Form/blob/2.7/FormTypeInterface.php), but anyway this docblock comment is useless.

TomasVotruba: @bocharsky-bw I'm talking about sight, not IDE.

bocharsky-bw: However, if this method has only `{@inheritdoc}` - it also seems that method hasn't a return value. But `{@inheritdoc}` considers as a better way than duplicate annotation from a superclass.

TomasVotruba: This PR is about removing docblocks, not adding `{@inheritdoc}`. I just command about changed made here.

javiereguiluz:
I don't know what to do. Some peole say "remove them" and others say "keep them" :confused:

theofidry:
Then for now maybe it's better to keep with Symfony practices and keep them :)

TomasVotruba:
@theofidry :+1: Convention over opinions

voronkovich:
But if we want to follow Symfony conventions we should use a `{@inheritdoc}` tag

theofidry:
Then so be it!

bocharsky-bw:
@theofidry :+1:

bocharsky-bw:
I think we wanted to merge https://github.com/symfony/symfony-demo/pull/153 . So we need to close https://github.com/symfony/symfony-demo/pull/153 ?

javiereguiluz:
After merging both PR I think we got what we wanted. See: https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Form/CommentType.php

PR #152, opened 2015-08-19
Remove unnecessary createCommentForm method

voronkovich:


bocharsky-bw:
Yes, it was missed in https://github.com/symfony/symfony-demo/pull/141#issuecomment-130432599

I think we shoud get rid of it :+1:

xabbuh:
:+1:

javiereguiluz:
@voronkovich thanks a lot for this improvement. I love it when some change removes more things that it adds!

PR #154, opened 2015-08-21
Added a paginator using KnpPaginatorBundle

ccab:
Please review and let me know


src/AppBundle/Controller/BlogController.php
```diff
@@ -36,12 +36,18 @@ class BlogController extends Controller
     /**
      * @Route("/", name="blog_index")
      */
-    public function indexAction()
+    public function indexAction(Request $request)
     {
-        $em = $this->getDoctrine()->getManager();
-        $posts = $em->getRepository('AppBundle:Post')->findLatest();
```
javiereguiluz: I'd like the controller to be:

```php
$page = $request->query->getInt('page', 1);
$posts = $em->getRepository('AppBundle:Post')->findLatest($page);
```

src/AppBundle/Controller/BlogController.php
```diff
@@ -36,12 +36,18 @@ class BlogController extends Controller
...
 
-        return $this->render('blog/index.html.twig', array('posts' => $posts));
+        $paginator  = $this->get('knp_paginator');
+        $pagination = $paginator->paginate(
+            $query,
+            $request->query->getInt('page', 1)/*page number*/,
+            10/*limit per page*/
+        );
+
+        return $this->render('blog/index.html.twig', array('pagination' => $pagination));
```
javiereguiluz: I'd like to maintain the variable name as `posts` instead of `pagination`. Being paginated or not, this varialbe stores a collection a posts and therefore, the template is more readable.

app/AppKernel.php
```diff
@@ -21,6 +21,7 @@ public function registerBundles()
             new Doctrine\Bundle\DoctrineBundle\DoctrineBundle(),
             new Sensio\Bundle\FrameworkExtraBundle\SensioFrameworkExtraBundle(),
             new AppBundle\AppBundle(),
+            new \Knp\Bundle\PaginatorBundle\KnpPaginatorBundle()
```
javiereguiluz: No need for the leading slash in `\Knp`

app/Resources/views/blog/index.html.twig
```diff
@@ -14,6 +14,11 @@
             {{ post.summary|md2html }}
         </article>
     {% endfor %}
+
+    {# display navigation #}
```
javiereguiluz: This comment probably should be removed because the code below it's very easy to understand.

app/check.php
```diff
@@ -68,7 +68,7 @@
 echo '  than the one used with your web server. To be on the'.PHP_EOL;
 echo '      safe side, please check the requirements from your web'.PHP_EOL;
 echo '      server using the ';
-echo_style('yellow', 'web/check.php');
+echo_style('yellow', 'web/config.php');
```
bocharsky-bw: I think this changes should be reverted. We fix it in https://github.com/symfony/symfony-demo/pull/140/files

composer.json
```diff
@@ -24,7 +24,8 @@
         "symfony/monolog-bundle"               : "~2.7",
         "symfony/swiftmailer-bundle"           : "~2.3",
         "symfony/symfony"                      : "~2.7",
-        "twig/extensions"                      : "~1.2"
+        "twig/extensions"                      : "~1.2",
+        "knplabs/knp-paginator-bundle": "^2.4"
```
bocharsky-bw: Could you align version like in previous dependencies?

app/AppKernel.php
ccab: Sorry for this one, my IDE

composer.json
ccab: nice catch @bocharsky-bw

src/AppBundle/Controller/BlogController.php
```diff
@@ -36,12 +36,18 @@ class BlogController extends Controller
...
-        $em = $this->getDoctrine()->getManager();
-        $posts = $em->getRepository('AppBundle:Post')->findLatest();
+        $query = $this->getDoctrine()->getRepository('AppBundle:Post')->queryLatest();
 
-        return $this->render('blog/index.html.twig', array('posts' => $posts));
+        $paginator  = $this->get('knp_paginator');
+        $pagination = $paginator->paginate(
+            $query,
+            $request->query->getInt('page', 1)/*page number*/,
+            10/*limit per page*/
```
bocharsky-bw: We have nice constant `Post::NUM_ITEMS` for limit.

app/check.php
ccab: Thanks I notice the changes but now I'll modify it manually as you say there

src/AppBundle/Controller/BlogController.php
bocharsky-bw: The page already handled by paginator. Why we should use it in query manually?

composer.json
bocharsky-bw: And you could change `^` to `~` too

src/AppBundle/Controller/BlogController.php
ccab: @javiereguiluz I agree the $page but I'm kind of lost in the $post

javiereguiluz: I'm sorry but I don't understand your comment. Maybe KnpPaginator behaves differently, but I usually put all this paginator logic inside the specific method of the custom repository. In the controller I only get the `page` number and pass it to the query.

ccab: in this case the query doesn't need to know the current page, these parameter is handled by the paginator

voronkovich:
Please, add more fixtures for posts. Now it's only 10.

src/AppBundle/Controller/BlogController.php
bocharsky-bw: We should pass paginator and request objects into the repository to move out all this pagination logic from the controller. But I think better to keep it in the controller.

ccab:
Actually the $limit argument in the queries isn't necessary, what should I do remove it or not?

src/AppBundle/Controller/BlogController.php
```diff
@@ -36,10 +36,17 @@ class BlogController extends Controller
...
      */
-    public function indexAction()
+    public function indexAction(Request $request)
     {
-        $em = $this->getDoctrine()->getManager();
-        $posts = $em->getRepository('AppBundle:Post')->findLatest();
+        $query = $this->getDoctrine()->getRepository('AppBundle:Post')->queryLatest();
+        $page = $request->query->getInt('page', 1);
+
+        $paginator  = $this->get('knp_paginator');
```
stof: you have an extra space here

src/AppBundle/Controller/BlogController.php
stof: @javiereguiluz the issue with this is that KnpPaginator is a service doing the pagination. so using it in the Doctrine repository makes things complex (this could be an argument in favor of Pagerfanta btw)

src/AppBundle/Controller/BlogController.php
ccab: thanks @stof

src/AppBundle/Controller/BlogController.php
```diff
@@ -34,12 +34,18 @@
 class BlogController extends Controller
 {
     /**
-     * @Route("/", name="blog_index")
+     * @Route("/{page}", name="blog_index", defaults={"page" = 1}, requirements={"page" : "\d+"})
```
javiereguiluz: This would generate URLs like the following:

```
/blog
/blog/2
/blog/3
...
```

Don't you think those URLs are a bit ugly? I'd prefer to have something like:

```
/blog
/blog/page-2
/blog/page-3
...
```

Or:

```
/blog
/blog/page/2
/blog/page/3
...
```

What do you think?

bocharsky-bw: I think you're right. We are losing page context. In case `/blog/2` it looks like we request post with id = 2.

I like last example.Could we do that without adding additional `@Route` annotation?

ccab: Ok you have a point there I'll working on that, but I think it needs a new @Route

ccab:
Ok I got something like this:

@Route("/", name="blog_index", defaults={"page" = 1})
@Route("/page/{page}", name="page_blog_index", defaults={"page" = 1}, requirements={"page" : "\d+"})

The default for page in both routes is because I don't want to catch the page as a query string but as a  route parameter. It's a little ugly doesn't it ?

bocharsky-bw:
What about `blog_index_paged` name to be consistent with first route name?
Nonetheless, I don't like to use 2 routes for this action :(

javiereguiluz:
@bocharsky-bw if we don't want to define two different routes, then we need to pass the page number as a query string. @ccab I agree with @bocharsky-bw about the new route name: I personally use and I've seen a lot of times the practice of naming it as `*_paginated`, so it could be `blog_index_paginated`. (The other option would be to not name it at all, because we are not going to generate this URL in any template).

bocharsky-bw:
What about next route definition:
``` php
/**
 *@Route("/{page_prefix}{page}", name="blog_index", defaults={"page" = 1, "page_prefix" = "page-"}, requirements={"page" : "\d+", "page_prefix" = "page-"})
 */
```
Does it too confused?

javiereguiluz:
@bocharsky-bw yes, it looks very confusing to me.

ccab:
yea @bocharsky-bw I turn around something like that but we need to think in code readability too, for newcomers it could be a little confusing. @javiereguiluz I'll remove the name for the second @Route

ccab:
A clue about Travis-CI ??

bocharsky-bw:
Try to check this [suggestion](https://github.com/symfony/symfony/issues/14790).

ccab:
Tanks @bocharsky-bw, as you may see in [this commit](https://github.com/KnpLabs/KnpPaginatorBundle/commit/3f00050cfbed246bb6f195ebee6badb50190fcf6) this should be solved but I am using the 2.4.2 release and I still got it, so I'm following directions from [UPGRADE-3.0](https://github.com/symfony/symfony/commit/1dbf67ed667c99f0d08ad61bf2bf4d2ec9fbada1) and I'll enable php as templating engine

ccab:
There is something weird with `deps="low"` env variable

app/config/config.yml
```diff
@@ -36,7 +36,7 @@ framework:
     csrf_protection: ~
     validation:      { enable_annotations: true }
     templating:
-        engines: ['twig']
+        engines: ['twig','php']
```
javiereguiluz: This change makes me cry :sob: Is there any chance to avoid it?

I'm starting to think that this paginator is seriously flawed or deprecated (because of this change, the other one in Composer, the impossibility of using the paginator in the entity to avoid polluting the controller, etc.)

composer.json
```diff
@@ -23,8 +23,10 @@
         "symfony/assetic-bundle"               : "~2.6",
         "symfony/monolog-bundle"               : "~2.7",
         "symfony/swiftmailer-bundle"           : "~2.3",
-        "symfony/symfony"                      : "~2.7",
-        "twig/extensions"                      : "~1.2"
+        "symfony/symfony"                      : "~2.7.3",
```
bocharsky-bw: Why we change `~2.7` to `~2.7.3` for paginator?

app/config/config.yml
stof: why is it added ?

composer.json
```diff
@@ -23,8 +23,10 @@
...
+        "twig/extensions"                      : "~1.2",
+        "knplabs/knp-paginator-bundle"         : "~2.4.2",
+        "knplabs/knp-components"               : "~1.2.4"
```
stof: is this needed ? KnpPaginatorBundle already depends on it

composer.json
```diff
@@ -23,8 +23,10 @@
         "symfony/assetic-bundle"               : "~2.6",
         "symfony/monolog-bundle"               : "~2.7",
         "symfony/swiftmailer-bundle"           : "~2.3",
-        "symfony/symfony"                      : "~2.7",
-        "twig/extensions"                      : "~1.2"
+        "symfony/symfony"                      : "~2.7.3",
+        "twig/extensions"                      : "~1.2",
+        "knplabs/knp-paginator-bundle"         : "~2.4.2",
```
stof: until now, the order of requirements was the following:

- php
- extensions
- packages in alphabetical order

It would make sense to put the package in the right place

app/AppKernel.php
```diff
@@ -21,6 +21,7 @@ public function registerBundles()
             new Doctrine\Bundle\DoctrineBundle\DoctrineBundle(),
             new Sensio\Bundle\FrameworkExtraBundle\SensioFrameworkExtraBundle(),
             new AppBundle\AppBundle(),
+            new Knp\Bundle\PaginatorBundle\KnpPaginatorBundle()
```
stof: this bundle should be registered before AppBundle. It generally makes sense to keep AppBundle at the end of the list

app/config/config.yml
ccab: Ok ok lets talk about it, I just following the recommendations made in this [symfony commit](https://github.com/symfony/symfony/commit/1dbf67ed667c99f0d08ad61bf2bf4d2ec9fbada1) because the `templating.helper.router` is needed by `knplabs/knp-paginator-bundle` prior to 2.4.2

ccab:
I make all the restrictions in the packages versions in order to pass all test with the `composer update --prefer-lowest`, but I must confess I'am not happy about the end result, so what do you think if we change to `Pagerfanta`

app/config/config.yml
stof: well, this has been resolved in the bundle, so there is no need to add it anymore. Enabling it was a workaround for the KnpPaginatorBundle bug

stof:
@javiereguiluz this is a project with locked dependencies. Why are we running builds with different deps than the locked ones ?

javiereguiluz:
@stof I'm not sure I understand you. Please tell me what should I change or add. Thanks!

composer.json
ccab: I just want to ensure `knplabs/knp-components` install with this version at least because of [this bug](https://github.com/KnpLabs/knp-components/issues/94)

composer.json
ccab: Same https://github.com/symfony/symfony-demo/pull/154#discussion_r38647331

stof:
@javiereguiluz why are we running ``composer update`` in some Travis jobs ? Dependencies are locked in the project, so the CI server should only run ``composer install``

composer.json
bocharsky-bw: Do you mean it won't work on `2.8`?

bocharsky-bw:
I think `composer update` needs for legacy PHP compatibility, doesn't it?

ccab:
Actually this is the main reason for my last 4 commits, because I need to pass `composer update` with the latest versions :-) & `composer update --prefer-lowest` with the lowest versions :-(

bocharsky-bw:
Well, you could try to constraint lowest version `knplabs/knp-paginator-bundle": "~2.4,>=2.4.2"`

composer.json
ccab: @stof said it's fixed so it should work

composer.json
ccab: Is there another way to do it?

ccab:
Yea @bocharsky-bw it could solve the problem now, but in the future we will need to add some other bundles and we could have the same problem again, always watching lowest versions for bugs, I'm with @stof on this, locked dependencies is the answer

src/AppBundle/Controller/BlogController.php
```diff
@@ -34,12 +34,20 @@
 class BlogController extends Controller
 {
     /**
-     * @Route("/", name="blog_index")
+     * @Route("/", name="blog_index", defaults={"page" = 1})
+     * @Route("/page/{page}", name="blog_index_paginated", defaults={"page" = 1}, requirements={"page" : "\d+"})
```
sebastianblum: Why do you add an default for page?

/page/ is not a valid path for me, I would remove the default for page on the "blog_index_paginated"

bocharsky-bw: After some refactoring with separate route definitions you have a point now.
We should avoid defaults for page in second route annotation.

ccab: Done ! :+1:

app/AppKernel.php
```diff
@@ -20,7 +20,8 @@ public function registerBundles()
             new Symfony\Bundle\AsseticBundle\AsseticBundle(),
             new Doctrine\Bundle\DoctrineBundle\DoctrineBundle(),
             new Sensio\Bundle\FrameworkExtraBundle\SensioFrameworkExtraBundle(),
-            new AppBundle\AppBundle(),
+            new Knp\Bundle\PaginatorBundle\KnpPaginatorBundle(),
+            new AppBundle\AppBundle()
```
xabbuh: please keep the comma

ccab: Done !

xabbuh: seems like you need to push the commit :)

ccab: I have to ask, any particular reason for this ?

javiereguiluz: @ccab http://symfony.com/doc/current/contributing/code/standards.html#structure

> Add a comma after each array item in a multi-line array, even after the last one;

javiereguiluz: The official reasoning for this is that when you add a new element to the array, you just need to add that element, not the element and a comma before it. This makes the diff operation cleaner for the pull request.

ccab: Sorry I miss that :+1:  for the contributing book

ccab:
What if we change the basic template for another one bootstrap oriented, this can be easily done changing a config key

Default:
snapshot5

Now with Bootstrap:

snapshot6

bocharsky-bw:
Bootstrap oriented looks better :+1:

voronkovich:
:+1: Yeah, bootstrap oriented is totaly better. Now it seems like the Mathias Noback blog http://php-and-symfony.matthiasnoback.nl :)

ccab:
@javiereguiluz what do you think, it's finished ?

javiereguiluz:
@ccab sorry for the delay. Your feature is now finally merged (I made some cosmetic changes in 97e8348). Thanks a lot for working on this :)

PR #158, opened 2015-08-29
Add flash messages

voronkovich:
Post created
Post updated
Post deleted

### DIFF
--- a/app/Resources/translations/messages.ru.xliff
+++ b/app/Resources/translations/messages.ru.xliff
@@ -195,6 +195,18 @@
                 <source>post.no_comments</source>
                 <target>Будьте первыми кто прокомментирует эту запись.</target>
             </trans-unit>
+            <trans-unit id="post.created_successfully">
+                <source>post.created_successfully</source>
+                <target>Запись успешно создана!</target>
+            </trans-unit>
+            <trans-unit id="post.updated_successfully">
+                <source>post.updated_successfully</source>
+                <target>Запись успешно обновлена!</target>
+            </trans-unit>
+            <trans-unit id="post.deleted_successfully">
+                <source>post.deleted_successfully</source>
+                <target>Запись успешно удалена!</target>
+            </trans-unit>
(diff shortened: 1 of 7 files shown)


app/Resources/translations/messages.ru.xliff
```diff
@@ -195,6 +195,18 @@
                 <source>post.no_comments</source>
                 <target>Будьте первыми кто прокомментирует эту запись.</target>
             </trans-unit>
+            <trans-unit id="post.created_successfully">
+                <source>post.created_successfully</source>
+                <target>Запись успешно сохранена!</target>
```
bocharsky-bw: I think better is `создана` instead of `сохранена` in this case to be consistent with English translation and translation key.

voronkovich: You're right, @bocharsky-bw. I've fixed the translation.

bocharsky-bw:
:+1:

javiereguiluz:
@voronkovich thanks for working on this. Even if I don't like to display flash messages for when things go right ... I like this PR because it shows a complete example of how to display flash messages, and that's definitely something we want to do.

My only concern was that the original code was a bit overcomplicated for the demo application. That's why I took your work and made some simplifications in #163. I'm closing this PR then, but I reused your commits, so you'll get full credit for your work. Thanks!

app/Resources/views/admin/layout.html.twig
```diff
@@ -7,6 +7,9 @@
 #}
 {% extends 'base.html.twig' %}
 
+{# Import macros needed for displaying flash messages. See http://twig.sensiolabs.org/doc/tags/import.html #}
+{% import 'macro/messages.html.twig' as messages %}
```
stof: macros should be imported in the template where they are used. Importing them in a parent template and expecting them to work in child templates is just a side effect of the Twig 1.x implementation, and does not work anymore in Twig 2

PR #159, opened 2015-08-30
Update uk translations for flash messages

bocharsky-bw:
In addition to https://github.com/symfony/symfony-demo/pull/158


javiereguiluz:
Thanks @bocharsky-bw.

PR #161, opened 2015-08-30
Add indonesian translation

egig:


javiereguiluz:
@egig thanks for providing this new translation! *Terima kasih*

PR #162, opened 2015-08-30
Fix locale dropdown url

voronkovich:


bocharsky-bw:
:+1:

javiereguiluz:
@voronkovich thanks for this improvement. I like it a lot! 

I only have a question. What would happen when a 404 error without an associated route is triggered? In those cases, there is no route defined:

routing-404-error

voronkovich:
@javiereguiluz , I think we could fix this by adding a checking `app.request.get('_route') ? app.request.get ('_route') : 'blog_index'`

javiereguiluz:
The `get()` method can define a default value as second param, so maybe we can do this: `app.request.get('_route', 'blog_index')`

app/Resources/views/base.html.twig
```diff
@@ -83,7 +83,7 @@
                                     <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false"><i class="fa fa-globe"></i> <span class="caret"></span></a>
                                     <ul class="dropdown-menu locales" role="menu">
                                         {% for locale in locales() %}
-                                            <li {% if app.request.locale == locale.code %}class="active"{% endif %}><a href="{{ path('blog_index', { _locale: locale.code }) }}">{{ locale.name|capitalize }}</a></li>
+                                            <li {% if app.request.locale == locale.code %}class="active"{% endif %}><a href="{{ path(app.request.get('_route'), app.request.get('_route_params')|merge({ _locale: locale.code })) }}">{{ locale.name|capitalize }}</a></li>
```
stof: you should pass an empty array as default value for the route_params, to get an empty array rather than null when the error happens without route

voronkovich:
@javiereguiluz, @stof thanks! I've updated the PR.

xabbuh:
:+1:

javiereguiluz:
@voronkovich thanks for this improvement!

PR #163, opened 2015-08-31
Add flash messages

javiereguiluz:
This finishes #158 by simplifying its contents.


app/Resources/views/admin/layout.html.twig
```diff
@@ -7,6 +7,12 @@
 #}
 {% extends 'base.html.twig' %}
 
+{#
+    Import macros used for displaying flash messages.
+    See http://twig.sensiolabs.org/doc/tags/import.html
+#}
+{% import 'macro/messages.html.twig' as helper %}
```
stof: same comment that in the original PR. Importing macros in the parent to use them in the child is not a supported way to use macros and will be removed in Twig 2.x

app/Resources/views/admin/blog/index.html.twig
```diff
@@ -5,6 +5,8 @@
 {% block main %}
     <h1>{{ 'title.post_list'|trans }}</h1>
 
+    {{ helper.render_flash_messages() }}
```
stof: Rather than using a macro, I suggest using an include of a template rendering flash messages

app/Resources/views/macro/messages.html.twig
```diff
@@ -0,0 +1,25 @@
...
+   Macros are comparable with functions in regular programming languages.
+   They are useful to put often used HTML idioms into reusable elements
+   to not repeat yourself.
+   See http://twig.sensiolabs.org/doc/tags/macro.html
+#}
+{% macro render_flash_messages() %}
+    {% from _self import alert %}
+
+    <div class="messages">
+        {% for flash_message in app.session.flashBag.get('success') %}
```
stof: this will start a session even if the user has no session yet. You should wrap this in a ``if app.session.started`` check

app/Resources/views/admin/blog/index.html.twig
javiereguiluz: I had some doubts about this. But after reading your comment, I've created the template and removed te macros. We'll add some macros in other feature. Thanks!

app/Resources/views/macro/messages.html.twig
javiereguiluz: Good catch. I still belive this is a Symfony bug ... but until it's fixed, let's use the `if ...` trick.

stof: well, you are asking to read from the session. currently, Symfony requires to start the session to read from it. There is proposal to change it, but it cannot be considered as a bugfix (it is a big refactoring), and I'm not sure it is ready.

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -87,6 +87,11 @@ public function newAction(Request $request)
             $em->persist($post);
             $em->flush();
 
+            // Flash messages are used to notify the user about the result of the
+            // actions. They only last as long as the next request.
```
xabbuh: We should improve this description like we did in the docs (flash messages last until they are retrieved from the bag).

javiereguiluz: Where can I find the updated explanation? In http://symfony.com/doc/current/book/controller.html#flash-messages I read:

> You can also store small messages that will be stored on the user's session for exactly one additional request. [...]
>
> [...] By design, flash messages are meant to live for exactly one request (they're "gone in a flash").

xabbuh: Sorry, it's not merged yet (see symfony/symfony-docs#5640).

javiereguiluz: Thanks for the reference. I've just reworded this help note in 3eefa76.

PR #160, opened 2015-08-30
Show message when no posts found

voronkovich:
Frontend
Backend

PR #164, opened 2015-08-31
Update uk translations according to #160

bocharsky-bw:


javiereguiluz:
Thanks @bocharsky-bw.

PR #165, opened 2015-08-31
Updated Czech translation

hason:


javiereguiluz:
@hason merged! Thanks *Děkuji*

PR #166, opened 2015-08-31

davialexandre:
Sync with changes introduced by #160


javiereguiluz:
Thank you @davialexandre.

PR #170, opened 2015-09-01
Add to the readme a note about how to run the application in production environment

voronkovich:


bocharsky-bw:
:+1:

xabbuh:
I am :-1: here. Adding this note adds the misleading impression that it might be a good idea to run PHP's built-in web server in production environments.

bocharsky-bw:
@xabbuh You have a point there. What about to add more commets, for example, never use it in prodaction and use only for developing things like checking custom error pages, etc. What do you think?

javiereguiluz:
Another argument against adding this change would be that we want this application to always being executed on `dev` environment. Why? Because this app is for newcomers and for learning purposes. It makes sense in those scenarios to use the toolbar + profiler (moreover when we've just redesigning them from scratch).

voronkovich:
@javiereguiluz, I think newcomers should know about differences between `prod` and `dev` environments.

voronkovich:
Imgur

javiereguiluz:
@voronkovich your last screenshot is definitive to me :)  I don't think it's right to mention something that we explicitly discourage or not recommend. That's why I'm closing this pull request without merging it. Thanks for understanding it.

PR #171, opened 2015-09-03
Updated Travis CI configuration

javiereguiluz:
Changes made as recommended by @stof in https://github.com/symfony/symfony-demo/pull/154#issuecomment-137465466

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ matrix:
   allow_failures:
     - php: hhvm
     - php: nightly
-    - env: SYMFONY_VERSION="2.7.x-dev"
     - env: SYMFONY_VERSION="2.8.x-dev"
 
 before_install:
@@ -36,7 +35,7 @@ before_install:
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require --dev --no-update symfony/symfony:$SYMFONY_VERSION; fi
 
 install:
-  - if [ "$deps" = "low" ]; then composer update --prefer-lowest; fi
+  - if [ "$deps" = "low" ]; then composer install --prefer-lowest; fi
   - if [ "$deps" != "low" ]; then composer install; fi
 
 script:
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
Seems `--prefer-lowest` is an argument for `composer up` command only.

ccab:
It makes sense

javiereguiluz:
Closing it as "won't fix".

stof:
@javiereguiluz my point in the comment you linked is that it does not really make sense to test against different deps versions as we are locking deps

javiereguiluz:
@stof you are right. Sorry for not understanding you earlier. I've just reopened and made a lot of simplifications to the Travis config. Thanks!

.travis.yml
```diff
@@ -1,43 +1,23 @@
...
     - php: hhvm
     - php: nightly
   allow_failures:
     - php: hhvm
-    - php: nightly
-    - env: SYMFONY_VERSION="2.7.x-dev"
-    - env: SYMFONY_VERSION="2.8.x-dev"
 
 before_install:
   - composer self-update
```
stof: you can even get rid of this (it was necessary to have a Composer version supporting ``--prefer-lowest`` because Travis has not updated its environment since February)

bocharsky-bw:
Cool! It has become much easier now :+1:

.travis.yml
javiereguiluz: Done. Thanks!

javiereguiluz:
@bocharsky-bw I agree! @stof's comments are always spot on. I love the new Travis config.

stof:
@javiereguiluz I also suggest enabling the Travis caching for the composer cache, so that it does not to re-download the same archives for each build (project with locked dependencies benefit from it even more than others because the deps change only when you commit a change, not when a new release of a deps is done:

```yaml
cache:
  directories:
    - $HOME/.composer/cache/files
```

javiereguiluz:
Nice! Enabled the cache too. Thanks.

stof:
:+1:

PR #173, opened 2015-09-04
Add an instant search (via AJAX)

voronkovich:
It's almost like the Google... but better
img1
img2
img3
img4


bocharsky-bw:
Wow, it's cool! But I think it's a bit overhead for demo project.

javiereguiluz:
@voronkovich :+1: I like this idea a lot. Thanks for working on it.

However, after a first quick review of the code, I find the `TermSplitter` and the `TermEmphasizer` features a bit overkill for the demo project. Please don't make any changes yet, but I think we should get rid of both of them and focus just on the Ajax feature.

voronkovich:
@javiereguiluz, if we remove the `TermSplitter` we will have to validate the ajax request anyway. The `TermSplitter` just only remove invalid characters from the query string and splits it into words.
I agree with removing `TermEmphasizer`. It is overhead, you're right.

voronkovich:
I've refactored the code. All overheads were beheaded.
overhead

src/AppBundle/Controller/BlogController.php
```diff
@@ -114,4 +116,43 @@ public function commentFormAction(Post $post)
...
+
+    /**
+     * @Route("/search", name="blog_search")
+     * @Method("GET")
+     *
+     * @return JsonResponse
+     */
+    public function searchAction(Request $request)
+    {
+        $q = $request->query->get('q');
```
theofidry: let's avoid a 1 char variable name :)

src/AppBundle/Controller/BlogController.php
```diff
@@ -114,4 +116,43 @@ public function commentFormAction(Post $post)
...
+        // length is less than 2
+        $terms = array_unique(explode(' ', strtolower($query)));
+        $terms = array_filter($terms, function($term) {
+            return 2 <= strlen($term);
+        });
+
+        $posts = new ArrayCollection();
+
+        if (!empty($terms)) {
+            $em = $this->getDoctrine()->getManager();
```
theofidry: again, better to use `$manager` rather than `$em`

bocharsky-bw: @theofidry Why do you think so? We get exactly the **entity** manager here, so I prefer to use `$em`, it's more specific I think.

voronkovich: By the way, the `em` is using in the Symfony documentation. See http://symfony.com/doc/current/book/doctrine.html#persisting-objects-to-the-database

src/AppBundle/Repository/PostRepository.php
```diff
@@ -38,4 +38,24 @@ public function findLatest($limit = Post::NUM_ITEMS)
             ->getResult()
         ;
     }
+
+    public function findByTerms(array $terms, $limit = Post::NUM_ITEMS)
+    {
+        $queryBuilder = $this->createQueryBuilder('p');
```
bocharsky-bw: What about to shorten it with `$qb`?

src/AppBundle/Controller/BlogController.php
voronkovich: It's vastly using name for a search request. The Symfony search uses this name http://api.symfony.com/2.7/index.html?q=request

src/AppBundle/Repository/PostRepository.php
```diff
@@ -38,4 +38,24 @@ public function findLatest($limit = Post::NUM_ITEMS)
...
+
+        $i = 0;
+        foreach ($terms as $term) {
```
bocharsky-bw: We could use unique index on traversing `$terms` like:
`foreach ($terms as $i => $term) {`

src/AppBundle/Repository/PostRepository.php
theofidry: its really a bad practice to get variable names under 4 characters so IMO that's definitely not something that should be encouraged in this project

src/AppBundle/Controller/BlogController.php
bocharsky-bw: Well, variable's name is matching with input name, so I prefer to use `$q` to be consistent.

src/AppBundle/Controller/BlogController.php
theofidry: to met its more some legacy left in the doc and its a better practice to have consistent names. The `$id` property in an entity being the only exception. Also it's something [PHPMD](http://phpmd.org/rules/naming.html#shortvariable) does not accept too.

src/AppBundle/Controller/BlogController.php
theofidry: yeah, as an URL param, not a PHP variable

src/AppBundle/Repository/PostRepository.php
bocharsky-bw: Symfony Best Practices don't mention about short variable's name, so I prefer to based on Symfony docs. If it'll be changed - so we could change it in Symfony demo

theofidry: It's not a Symfony best practice, it's a PHP one. Having `$em` as a variable is more a legacy from Doctrine than a Symfony convention.

voronkovich: @theofidry, @bocharsky-bw I think it's better to create a particular issue to discuss code style for this project. What do you think?

theofidry: @voronkovich true

bocharsky-bw: You're right. I think now we should following styles already used in this project.

src/AppBundle/Repository/PostRepository.php
voronkovich: Thanks,  @bocharsky-bw.  I've fixed this.

voronkovich:
I've renamed the `$q` variable to `$query` and removed the `$em`, because it's not needed here

src/AppBundle/Controller/BlogController.php
```diff
@@ -114,4 +116,47 @@ public function commentFormAction(Post $post)
...
+        $query = preg_replace('/[^[:alnum:] ]/', '', trim(preg_replace('/[[:space:]]+/', ' ', $query)));
+
+        // Splits the query into terms and removes all terms which
+        // length is less than 2
+        $terms = array_unique(explode(' ', strtolower($query)));
+        $terms = array_filter($terms, function($term) {
+            return 2 <= strlen($term);
+        });
+
+        $posts = new ArrayCollection();
```
stof: this should be an array. Doctrine does not return collection objects when doing queries

src/AppBundle/Controller/BlogController.php
```diff
@@ -114,4 +116,47 @@ public function commentFormAction(Post $post)
...
+                ->getRepository('AppBundle:Post')
+                ->findByTerms($terms)
+            ;
+        }
+
+        $results = array();
+
+        foreach ($posts as $post) {
+            array_push($results, array(
+                'result' => $post->getTitle(),
```
stof: I would rather use ``title`` as key if possible

app/Resources/assets/js/jquery.instantSearch.js
```diff
@@ -0,0 +1,79 @@
...
+        $input.focusin(function(e) {
+            $preview.fadeIn();
+            updatePreview();
+        });
+
+        $input.keyup(function(e) {
+            throttle(updatePreview, config.previewDelay);
+        });
+    }
+})(window.jQuery)
```
stof: please add a final semicolon to avoid issues when concatenating files with Assetic (it works currently because this file is the last one)

src/AppBundle/Controller/BlogController.php
voronkovich: @stof, It's not possible, the `result` can be any html string.

voronkovich:
Fixed. Thanks @stof!

app/Resources/assets/js/jquery.instantSearch.js
```diff
@@ -0,0 +1,79 @@
...
+
+        var setPreviewItems = function(items) {
+            $preview.empty();
+
+            $.each(items, function(index, item) {
+                if (index > config.maxPreviewItems) {
+                    return;
+                }
+
+                addItemToPreview(item)
```
javiereguiluz: do we need a semicolon here to avoid issues in some browsers?

app/Resources/views/base.html.twig
```diff
@@ -155,6 +163,12 @@
                 $(document).ready(function() {
                     hljs.initHighlightingOnLoad();
                 });
+
+                $(function() {
```
javiereguiluz: Could we move this JS code to the above `$(document).ready(function() { ... }` block?

src/AppBundle/Controller/BlogController.php
stof: HTML ? If you expect HTML strings, the title needs to be escaped here, otherwise it is an XSS issue (I would rather change the JS code to allow giving text only, not HTML though, as it would be much safer for XSS and we don't need HTML here)

app/Resources/assets/js/jquery.instantSearch.js
stof: there is no issue here. Browsers are consistent in the way they insert semi-colons

app/Resources/assets/js/jquery.instantSearch.js
```diff
@@ -0,0 +1,79 @@
...
+
+    var defaultConfig = {
+        minQueryLength: 2,
+        maxPreviewItems: 10,
+        previewDelay: 500,
+        noItemsFoundMessage: 'No items found'
+    };
+
+    var throttle = (function(){
+        var timer = 0;
```
stof: your throttling function is flawed: it can be used only once as all usages of the ``throttle`` function are sharing the same timer

app/Resources/assets/js/jquery.instantSearch.js
javiereguiluz: OK. Then I ask for this change because of our own code consistency. Either add this semicolon ... or remove all the other semicolons. Thanks!

app/Resources/assets/js/jquery.instantSearch.js
```diff
@@ -0,0 +1,79 @@
+// jQuery plugin for an instant searching
```
stof: is this a plugin you wrote specifically for the Symfony demo, or some third-part plugin ?

app/Resources/assets/js/jquery.instantSearch.js
```diff
@@ -0,0 +1,79 @@
...
+        var timer = 0;
+        return function(callback, ms){
+            clearTimeout (timer);
+            timer = setTimeout(callback, ms);
+        };
+    })();
+
+    var initInstantSearch = function(el, config) {
+        var $input = $(el);
+        var $form = $input.parents('form').first();
```
stof: use ``$form = $input.closest('form')`` instead. It will be more efficient as it will stop traversing the parents when it finds the first form instead of searching them all to keep only the first one.

app/Resources/assets/js/jquery.instantSearch.js
```diff
@@ -0,0 +1,79 @@
...
+                if (index > config.maxPreviewItems) {
+                    return;
+                }
+
+                addItemToPreview(item)
+            });
+        }
+
+        var addItemToPreview = function(item) {
+            $preview.append('<li class="list-group-item"><a href="' + item.url + '">' + item.result + '</a></li>');
```
stof: you should not assume that the data returned from the server has already escaped the result and the URL for usage in HTML:

- it opens an XSS hole
- making the HTML escaping server-side in the JSON response does not make sense: the server should not assume that the value will be used in an HTML context without escaping.

app/Resources/assets/js/jquery.instantSearch.js
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        var noItemsFound = function() {
+            $preview.empty();
+            $preview.append('<li class="list-group-item">' + config.noItemsFoundMessage + '</li>');
```
stof: this is also missing HTML escaping

voronkovich:
Fixed again. Thanks, @stof!

xabbuh:
@voronkovich Seems like this needs a rebase.

voronkovich:
@xabbuh can you shortly explain what exactly I have to do with it?

stof:
See http://symfony.com/doc/current/contributing/code/patches.html#rework-your-patch

voronkovich:
Thanks for the link, @stof!

javiereguiluz:
This will be merged very soon. Last chance for comments or questions. Thanks!

app/Resources/assets/js/jquery.instantSearch.js
```diff
@@ -0,0 +1,134 @@
...
+        try {
+            parser.href = url;
+            return !!parser.hostname;
+        } catch (e) {
+            return false;
+        }
+    }
+
+    // See http://phpjs.org/functions/strip_tags/
+    function stripTags(input, allowed) {
```
stof: Don't strip tags. This is not the right way to escape HTML.
And it produces a broken output (if a post title contains a ``<``, it should be displayed, not removed).

app/Resources/assets/js/jquery.instantSearch.js
stof: this still has the XSS hole for item.result and item.url here. Here would be the proper way to do it:

```js
var $link = $('<a>').attr('href', item.url).text(item.result);
var $li = $('<li class="list-group-item">').append($link);

$preview.append($li);
```

This would let the browser handle the escaping of the text and prevent the ability to set extra thinks through the URL by having ``#" onclick="alert('xss')`` in the URL

app/Resources/assets/js/jquery.instantSearch.js
stof: btw, I hope you don't rely on ``strip_tags`` in your own PHP code to avoid XSS. It is **not** a way to prevent XSS.

app/Resources/assets/js/jquery.instantSearch.js
```diff
@@ -0,0 +1,134 @@
...
+        $input.focusout(function(e) {
+            $preview.fadeOut();
+        });
+
+        $input.focusin(function(e) {
+            $preview.fadeIn();
+            updatePreview();
+        });
+
+        $input.keyup(function(e) {
```
stof: no need to create a function proxying just a function.

```js
$input.keyup(debouncedUpdatePreview);
```

app/Resources/assets/js/jquery.instantSearch.js
```diff
@@ -0,0 +1,134 @@
...
+ */
+(function($) {
+    $.fn.instantSearch = function(config) {
+        return this.each(function() {
+            initInstantSearch(this, $.extend(true, defaultConfig, config || {}));
+        });
+    };
+
+    var defaultConfig = {
+        allowedTags: '',
```
stof: there is no reason to allow tags IMO. Your controller is **not** returning HTML currently.

src/AppBundle/Controller/BlogController.php
```diff
@@ -118,4 +119,48 @@ public function commentFormAction(Post $post)
...
+                ->findByTerms($terms)
+            ;
+        }
+
+        $results = array();
+
+        foreach ($posts as $post) {
+            array_push($results, array(
+                // We should never trust to admin user
+                'result' => strip_tags($post->getTitle()),
```
stof: if your contract is that ``results`` must be an HTML string, this is the wrong way to produce this HTML. you are not returning an HTML string here.

stof:
-1 on this implementation for now. We should not advocate ``strip_tags`` as a way to produce safe HTML strings, as it is not its behavior (and even less its goal)

javiereguiluz:
OK then. Let's wait a bit to solve the XSS issues :)

voronkovich:
Fixed! Thanks, @stof!

voronkovich:
@javiereguiluz, what about to merge this PR?

voronkovich:
@javiereguiluz, I've refactored this PR to use some of it's code also in the backend post filtering

src/AppBundle/Controller/BlogController.php
```diff
@@ -150,4 +151,42 @@ public function commentFormAction(Post $post)
...
+        $posts = [];
+
+        if (!empty($terms)) {
+            $posts = $this->getDoctrine()->getRepository(Post::class)->findByTerms($terms);
+        }
+
+        $results = [];
+
+        foreach ($posts as $post) {
+            array_push($results, [
```
Koc: why not just `$results[] = xxx` ?

voronkovich: @Koc, I don't know :) Anyway, this code was changed in #613.

javiereguiluz: We finally used `$results[] = xxx` ... mostly because PHPStorm displayed a pretty annoying message suggesting that 😄

PR #177, opened 2015-09-07
Rename Twig extensions according to best practices

### LINKED ISSUE #176: Refactor Twig extension names according to best practices guild
According to the Symfony2 best practices guild (http://symfony.com/doc/current/best_practices/templates.html#twig-extensions) twig extension naming convension would be the bundle name followed by underscore (_) and the extension name

ex: app_locale_extension (not app.locale_extension)

bocharsky-bw:
This fixes #176.

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -9,9 +9,9 @@ services:
 
     # These are the Twig extensions that create new filters and functions for
     # using them in the templates
-    app.twig.app_extension:
+    app.twig.markdown_extension:
         public:    false
-        class:     AppBundle\Twig\AppExtension
+        class:     AppBundle\Twig\MarkdownExtension
         arguments: ["@markdown"]
         tags:
             - { name: twig.extension }
--- a/src/AppBundle/Twig/AppExtension.php
+++ b/src/AppBundle/Twig/AppExtension.php
@@ -49,9 +49,9 @@ class AppExtension extends \Twig_Extension
     }
 
     // the name of the Twig extension must be unique in the application. Consider
-    // using 'app.extension' if you only have one Twig extension in your application.
+    // using 'app' if you only have one Twig extension in your application.
     public function getName()
     {
-        return 'app.extension';
+        return 'app';
     }
 }
--- a/src/AppBundle/Twig/MarkdownExtension.php
+++ b/src/AppBundle/Twig/MarkdownExtension.php
@@ -24,7 +24,7 @@ use AppBundle\Utils\Markdown;
  * @author Ryan Weaver <weaverryan@gmail.com>
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  */
-class AppExtension extends \Twig_Extension
+class MarkdownExtension extends \Twig_Extension
 {
     private $parser;
 
@@ -52,6 +52,6 @@ class AppExtension extends \Twig_Extension
     // using 'app' if you only have one Twig extension in your application.
     public function getName()
     {
-        return 'app';
+        return 'app_markdown';
     }
 }
--- a/src/AppBundle/Twig/LocaleExtension.php
+++ b/src/AppBundle/Twig/LocaleExtension.php
@@ -53,6 +53,6 @@ class LocaleExtension extends \Twig_Extension
     // the name of the Twig extension must be unique in the application
     public function getName()
     {
-        return 'app.locale_extension';
+        return 'app_locale';
     }
 }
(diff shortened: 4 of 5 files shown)


src/AppBundle/Twig/AppExtension.php
```diff
@@ -52,6 +52,6 @@ public function markdownToHtml($content)
     // using 'app.extension' if you only have one Twig extension in your application.
```
sebastianblum: you should change it also here in the comment

bocharsky-bw: Oh, thank you! I definitely missed it.

voronkovich:
I think the `_extension` suffix (we don't use a suffix `_type` in form types) not needed. Also I think we should rename `AppExtension` to `MarkdownExtension`.

bocharsky-bw:
Well, `app_extension` came from [best practices](http://symfony.com/doc/current/best_practices/templates.html#twig-extensions) like a `_extension` suffix. In this case we should change it in best practices first.

src/AppBundle/Twig/AppExtension.php
```diff
@@ -49,9 +49,9 @@ public function markdownToHtml($content)
     }
 
     // the name of the Twig extension must be unique in the application. Consider
-    // using 'app.extension' if you only have one Twig extension in your application.
+    // using 'app_extension' if you only have one Twig extension in your application.
```
voronkovich: @bocharsky-bw,  see this comment about one extension in the appplication.

bocharsky-bw: What if we had `UserType` form type and `UserExtension` twig extension in app bundle? Then we had a collision with `app_user` if we remove `_extension` suffix. So I don't like idea to remove `_extension` suffix. However, form types use much often then Twig extensions, so they had exclusion from `_type` suffix. I'll be better to add `_type` suffix to form type than remove `_extension` one.

stof: form type names and twig extension names cannot conflict. They are identifiers in different systems. We will never register a form type in Twig or a twig extension in the form registry (it is impossible)

bocharsky-bw: @stof So we could get rid of `_extension` suffix?

stof: yes

app/config/services.yml
```diff
@@ -11,7 +11,7 @@ services:
     # using them in the templates
     app.twig.app_extension:
```
xabbuh: Imo the service name should be changed too.

bocharsky-bw: Oh, thanks! I missed it

xabbuh:
:+1:

src/AppBundle/Twig/MarkdownExtension.php
```diff
@@ -24,7 +24,7 @@
  * @author Ryan Weaver <weaverryan@gmail.com>
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  */
-class AppExtension extends \Twig_Extension
```
stof: there is a comment referring to this file in SourceCodeExtension. Please update it too

bocharsky-bw: Thank you. I updated

javiereguiluz:
My only concern with this change is the following:

1) Previously we had an `AppExtension`, where you can obviously define all the custom filters and functions your application needs (we onle had one, but we could add more in the future).
2) Now we have a `MarkdownExtension` which obviously is just for Markdown and nothing else.

Some developers may think that you must define a Twig extension for each filter/function.

bocharsky-bw:
You're right, @javiereguiluz . So maybe better to refactor all Twig extensions with `AppExtension` except `SourceCodeExtension`?

stof:
the solution could indeed be to put the locale stuff in AppExtension, showing a case where the different functions needed by the app are provided by a single extension (SourceCodeExtension should indeed stay separate as it is not really here to be part of the application, but to expose a debugging stuff for the demo)

voronkovich:
I disagree with @javiereguiluz. I think that `AppExtension` is only good for importing native php functions (and simple custom functions) into Twig templates. We shouldn't put code that have no cohesion in a one class.

bocharsky-bw:
I think Javier idea is nice if you have only a few Twig filters like in our case.

bocharsky-bw:
So what to do? Refactor all Twig extensions except `SourceCodeExtension` with `AppExtension`?

javiereguiluz:
@bocharsky-bw I'd like to do exactly that, but let's wait a bit to see if anyone proposes something different. Thanks.

javiereguiluz:
Closing it in favor of #220, which implements the idea of merging all Twig extensions.

PR #179, opened 2015-09-07
Add source code templating helper to separate twig logic from templating logic

rasanga:
It is good to have a separate templating logic if the templating engine need to be changed later on.

### DIFF
--- a/src/AppBundle/Templating/Helper/SourceCodeHelper.php
+++ b/src/AppBundle/Templating/Helper/SourceCodeHelper.php
@@ -1,0 +1,129 @@
+<?php
+/**
+ * This file is part of the Symfony package.
+ *
+ * @copyright (c) 2015, symfony-demo
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace AppBundle\Templating\Helper;
+
+use ReflectionClass;
+use Symfony\Component\Templating\EngineInterface;
+use Symfony\Component\Templating\Helper\Helper;
+
+/**
+ * Class SourceCodeHelper
+ *
+ * @author Rasanga Perera <rasangaperera@gmail.com>
+ */
+class SourceCodeHelper extends Helper
+{
+    /**
+     * @var EngineInterface
+     */
+    private $templating;
+
+    /**
+     * @var string
+     */
+    private $kernelRootDir;
+
+    /**
+     * @param EngineInterface $templating
+     * @param string $kernelRootDir
+     */
+    public function __construct(
+        EngineInterface $templating,
+        $kernelRootDir
+    ) {
+        $this->templating = $templating;
+        $this->kernelRootDir = $kernelRootDir;
+    }
+
+    /**
+     * @param string $templateName
+     * @param string $templateSource
+     * @param array|null $controller
+     *
+     * @return string
+     */
+    public function showSourceCode(
+        $templateName,
+        $templateSource,
+        $controller = null
+    ) {
+        return $this->templating->render('default/_source_code.html.twig', array(
+            'controller' => $this->getController($controller),
+            'template'   => array(
+                'file_path'     => $this->kernelRootDir.'/Resources/views/'.$templateName,
+                'starting_line' => 1,
+                'source_code'   => $templateSource,
+            )
+        ));
+    }
+
+    /**
+     * @param array|null $controller
+     *
+     * @return array|void
+     */
+    private function getController($controller = null)
+    {
+        // this happens for example for exceptions (404 errors, etc.)
+        if (null === $controller) {
+            return;
+        }
+
+        $className = get_class($controller[0]);
+        $class = new ReflectionClass($className);
+        $method = $class->getMethod($controller[1]);
+
+        $classCode = file($class->getFilename());
+        $methodCode = array_slice($classCode, $method->getStartline() - 1, $method->getEndLine() - $method->getStartline() + 1);
+        $controllerCode = '    '.$method->getDocComment()."\n".implode('', $methodCode);
+
+        return array(
+            'file_path'     => $class->getFilename(),
+            'starting_line' => $method->getStartline(),
+            'source_code'   => $this->unindentCode($controllerCode)
+        );
+    }
+
+    /**
+     * Utility method that "unindents" the given $code when all its lines start
+     * with a tabulation of four white spaces.
+     *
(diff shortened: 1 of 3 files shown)


src/AppBundle/Twig/SourceCodeExtension.php
```diff
@@ -20,20 +26,47 @@
...
+     */
+    private $controller;
+
+    /**
+     * @param ContainerInterface $container
+     * @param Twig_LoaderInterface $loader
+     * @param $kernelRootDir
+     */
+    public function __construct(
+        ContainerInterface $container,
```
stof: injecting the container is the opposite of best practices

src/AppBundle/Templating/Helper/SourceCodeHelper.php
```diff
@@ -0,0 +1,129 @@
...
+use ReflectionClass;
+use Symfony\Component\Templating\EngineInterface;
+use Symfony\Component\Templating\Helper\Helper;
+
+/**
+ * Class SourceCodeHelper
+ *
+ * @author Rasanga Perera <rasangaperera@gmail.com>
+ */
+class SourceCodeHelper extends Helper
```
stof: adding a dependency on the Templating component for the Twig extension is precisely the opposite of what we are doing in Symfony 2.7+, where we are making all Twig extension usable without having the Templating component.

src/AppBundle/Twig/SourceCodeExtension.php
rasanga: Yes, but it's exceptional when injecting templating engine service if I inject SourceCodeHelper service instead it throws ServiceCircularReferenceException.

Same problem: https://github.com/symfony/symfony/issues/2347

stof:
-1 for this PR:

- getting rid of Twig is not planned at all: the Symfony best practices are to use Twig and this demo is here to showcase the best practices
- your PR is making the PHP templating engine first-class for this feature (while the PHP engine will move out of the core for 3.0). It does not decouple the logic from templating layers. It just changes the templating layer
- we don't want to encourage people to implement extra templating helpers in their own projects for each of the twig extensions. this does not make sense for projects (and it clearly does not make sense to do it this way for shared bundles)

src/AppBundle/Twig/SourceCodeExtension.php
```diff
@@ -11,6 +11,12 @@
 
 namespace AppBundle\Twig;
 
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Twig_Extension;
+use Twig_LoaderInterface;
+use Twig_SimpleFunction;
+use Twig_Template;
```
stof: We don't add use statements for classes of the global namespace in Symfony

src/AppBundle/Twig/SourceCodeExtension.php
rasanga: @stof I can add a comment referring to the issue and explaining the best practice and problem.

stof: @rasanga see my other comments on this PR. The whole PR is flawed

src/AppBundle/Templating/Helper/SourceCodeHelper.php
```diff
@@ -0,0 +1,129 @@
...
+     * @param array|null $controller
+     *
+     * @return string
+     */
+    public function showSourceCode(
+        $templateName,
+        $templateSource,
+        $controller = null
+    ) {
+        return $this->templating->render('default/_source_code.html.twig', array(
```
stof: if you were really writing a templating helper for the PHP engine, making it use a Twig templating for its implementation would be a very bad idea, as it would force to enable Twig to be able to use the PHP template helper

javiereguiluz:
@rasanga first of all, thanks for submitting this PR. This is a community project and everyone is encouraged to help us developing this application.

Having said that, I agree with @stof here. The problem of this pull request is that it proposes changes that are the exact opposite of the current and future Symfony trend. We are working hard to make Twig a first-class citizen inside Symfony and to get rid of most of the templating things ([see related post about this](http://symfony.com/blog/new-in-symfony-2-7-twig-as-a-first-class-citizen)).

Therefore, I'm closing this pull request as "won't fix". But please, don't hesitate to keep contributing with comments, code reviews, pull requests, etc. Thanks!

src/AppBundle/Twig/SourceCodeExtension.php
rasanga: Thank you for the tip :+1:

src/AppBundle/Templating/Helper/SourceCodeHelper.php
rasanga: We are not specifying any templating engine on the class and any templating engine that implements EngineInterface can be injected on the service configuration file.

src/AppBundle/Templating/Helper/SourceCodeHelper.php
rasanga: Yes, that make sense. :+1:

rasanga:
I also agree with @stof about using Twig as the primary templating engine for Symfony, but still I think it's good to have a seperating between the Twig extension and the template renderer.

My argument is that using a twig extension resctrict to use template renderer only in a twig view, what if we want to use the same template renderer in a model or controller?

In that case we need to inject @templating service and rewrite the view logic same as we did before in the extension class.

stof:
@rasanga this creates again a circular dependency, and forces to enable the templating system to use Twig, which is something we want to avoid.
Symfony 2.8 can be used without enabling the templating system at all, assuming you don't use AsseticBundle (which still depends on the templating system). Symfony 2.7 already can do it for the most part but it is missing support for that in the ``Controller::render()`` method.

What *could* be extracted (but not worth it IMO) is the logic used to extract the source code of the controller. And this is not even what you extracted (the controller is still stored in the Twig extension).
But the rendering logic is specific to Twig anyway as it relies on Twig templates. Adding extra layers creating a circular reference between twig and the extension is a bad idea if it is just for the sake of unused reusability. The template is also specific to the design of the demo.
And finally, the logic extracting the source of the template is specific to the demo too. It takes several shortcomings thanks to the knowledge of how we use it. The logic would not work in a general case where templates can come from different places. So there is no point extracting it IMO. The logic for the controller also takes some shortcomings.

rasanga:
It all make sense now, so there is no having a templating helper to render source code. Thank you for the explanation.

PR #180, opened 2015-09-07
Remove an unneeded state in the SourceCodeExtension

stof:


xabbuh:
:+1:

src/AppBundle/Twig/SourceCodeExtension.php
```diff
@@ -48,13 +47,11 @@ public function getFunctions()
...
 
-    public function showSourceCode(\Twig_Environment $twig, $template)
+    public function showSourceCode(\Twig_Environment $twig, \Twig_Template $template)
     {
-        $this->template = $template;
-
         return $twig->render('default/_source_code.html.twig', array(
             'controller' => $this->getController(),
-            'template'   => $this->getTemplate(),
+            'template'   => $this->getTemplate($template),
```
javiereguiluz: I agree with this PR, but I believe this code is now hard to understand: `getTemplate($template)` We can change the variable name or the method name or both.

stof: I renamed the method.

Renaming the variable would be a BC break btw, because Twig relies on it for its named arguments.

javiereguiluz:
Nice! Thanks @stof.

PR #181, opened 2015-09-07
Refactor AddUserCommand to use the Symfony validator component

voronkovich:


javiereguiluz:
The problem of this refactorization is that we no longer show how to use validators inside commands. I need more opinions before merging this :)

By the way, keep in mind that this application is developed to show the official best practices and as a learning resource for newcomers. That's why sometimes we may need to do something "good" instead of "perfect". In case of doubt, it's preferable to open an issue in this repository to discuss about a change instead of creating a pull request with that change. Thanks!

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -139,7 +140,7 @@ protected function interact(InputInterface $input, OutputInterface $output)
         $password = $input->getArgument('password');
         if (null === $password) {
             $question = new Question(' > <info>Password</info> (your type will be hidden): ');
-            $question->setValidator(array($this, 'passwordValidator'));
+            $question->setValidator($this->getValidatorForProperty('plainPassword'));
```
voronkovich: @javiereguiluz, this string show how to use validators

javiereguiluz: yes :)  but in my previous comment I was referring to validators defined inside the command. However, maybe doing that it's a wrong practice. That's why I want to read more opinions.

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -193,6 +187,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $encoder = $this->getContainer()->get('security.password_encoder');
         $encodedPassword = $encoder->encodePassword($user, $plainPassword);
         $user->setPassword($encodedPassword);
+        $user->setPlainPassword($plainPassword);
```
stof: setting the plain password here for the validator is too late: you already encoded it anyway. The validation should run before

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -193,6 +187,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+        // See http://symfony.com/doc/current/book/validation.html
+        $errors = $this->validator->validate($user, null, array('Default', 'registration'));
+
+        if (count($errors) > 0) {
```
bocharsky-bw: According to [Symfony Coding Standards](http://symfony.com/doc/current/contributing/code/standards.html) expected value should be first: `if (0 < count($errors)) {`

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -208,42 +215,25 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
-        if (strlen(trim($plainPassword)) < 6) {
-            throw new \Exception('The password must be at least 6 characters long');
-        }
+        return function($value) use ($validator, $property) {
+            $errors = $validator->validatePropertyValue('\AppBundle\Entity\User', $property, $value, array('Default', 'registration'));
 
-        return $plainPassword;
-    }
+            if (count($errors) > 0) {
```
bocharsky-bw: `if (0 < count($errors)) {`

voronkovich:
Fixed. Thanks @bocharsky-bw and @stof!

voronkovich:
@javiereguiluz,  I think we could create a command for deleting users and show how to use an input validator there.

javiereguiluz:
Closing it as "fixed" because the command to delete users has already been merged. Thanks!

PR #182, opened 2015-09-07
Add pt_BR translations for flash messages

davialexandre:


javiereguiluz:
Thanks @davialexandre.

PR #184, opened 2015-09-09
Updated the app to Symfony 2.7.4

### LINKED ISSUE #157: Running without native intl generates 500
At least `/en/admin/post/` fails with the following error:

An exception has been thrown during the rendering of a template ("The Symfony\Component\Intl\DateFormatter\IntlDateFormatter::__construct() method's argument $locale value NULL behavior is not implemented. Only the locale "en" is supported. Please install the "intl" extension for full localization capabilities.")

javiereguiluz:
This fixes #157.


xabbuh:
:+1:

PR #185, opened 2015-09-09
Add support for non-method callables in the source code extension

stof:
The additional complexity is low (just about getting the callable reflector, and this logic is not new as we already have it in Symfony). And the benefit of supporting it is that it shows that a controller can be any callable in Symfony, not only a class method (even though our own controllers are all class methods). This covers other cases:

- functions
- invokable objects (supported in Symfony both for class names and for services, even though there is a bug for the case of class names where it would not inject the container properly in it. PR is coming)
- closures (even though this cannot be used in the fullstack frameworks when specifying the controller in the routing because we cannot dump it, but a special controller resolver or a kernel.controller listener changing the controller could lead to such case)

### DIFF
--- a/src/AppBundle/Twig/SourceCodeExtension.php
+++ b/src/AppBundle/Twig/SourceCodeExtension.php
@@ -62,19 +62,41 @@ class SourceCodeExtension extends \Twig_Extension
             return;
         }
 
-        $className = get_class($this->controller[0]);
-        $class = new \ReflectionClass($className);
-        $method = $class->getMethod($this->controller[1]);
+        $method = $this->getCallableReflector($this->controller);
 
-        $classCode = file($class->getFilename());
+        $classCode = file($method->getFilename());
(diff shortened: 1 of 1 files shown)


voronkovich:
What about to create simple separate bundle (ShowSourceBundle) and move there the ControllerListener, SourceCodeExtension and related views? All this stuff is not needed for newcomers.

javiereguiluz:
@voronkovich at first sight I like your idea. This could even let us show how easy can you create additional bundles in your application when needed. Please let's us think over this idea a bit more before implementing it. Thanks!

voronkovich:
@javiereguiluz,  ok. I've created an issue about it.

stof:
@javiereguiluz this should still be merged before starting the move of classes to avoid conflicts

javiereguiluz:
@stof sure! I'm going to merge it right now. Thanks a lot for providing this improvement.

PR #188, opened 2015-09-09
add missing various document comments

rasanga:
Document comments are missing on various classes


src/AppBundle/EventListener/ControllerListener.php
```diff
@@ -28,13 +28,22 @@
  */
 class ControllerListener
 {
+    /**
+     * @var SourceCodeExtension
+     */
     protected $twigExtension;
 
+    /**
+     * @param SourceCodeExtension $twigExtension
```
stof: this adds exactly no value compared to what is provided by the method signature. In such case, we don't add meaningless phpdoc in Symfony

src/AppBundle/Twig/SourceCodeExtension.php
```diff
@@ -58,6 +68,9 @@ public function showSourceCode(\Twig_Environment $twig, $template)
         ));
     }
 
+    /**
+     * @return array|void
```
stof: null, not void

stof:
Your PR cannot be merged, because you worked on an outdated codebase. Please rebase it to fix conflicts

src/AppBundle/Twig/SourceCodeExtension.php
rasanga: Good spot, I will correct that

src/AppBundle/EventListener/ControllerListener.php
rasanga: Yes, I agree

rasanga:
Fixed all issues mentioned in code review.

src/AppBundle/EventListener/ControllerListener.php
```diff
@@ -35,6 +38,9 @@ public function __construct(SourceCodeExtension $twigExtension)
         $this->twigExtension = $twigExtension;
     }
 
+    /**
+     * @param FilterControllerEvent $event
+     */
```
xabbuh: This docblock doesn't add any value (the parameter type can be derived from the type hint).

src/AppBundle/Twig/AppExtension.php
```diff
@@ -26,8 +26,14 @@
...
 class AppExtension extends \Twig_Extension
 {
+    /**
+     * @var Markdown
+     */
     private $parser;
 
+    /**
+     * @param Markdown $parser
+     */
```
xabbuh: same here

javiereguiluz:
@rasanga thanks for adding all the missing PHPdocs and for your patience during the review process!

PR #189, opened 2015-09-10
"textarea" resize vertical only.

yceruto:
It would be better if we avoid the textarea is resized horizontally.

### DIFF
--- a/app/Resources/assets/scss/main.scss
+++ b/app/Resources/assets/scss/main.scss
@@ -159,7 +159,7 @@ footer {
 }
 
 textarea {
-    resize: vertical;
+    max-width: 100%;
 }
 
 //
--- a/app/Resources/assets/scss/main.scss
+++ b/app/Resources/assets/scss/main.scss
@@ -158,6 +158,10 @@ footer {
     }
 }
 
+textarea {
+    resize: vertical;
+}
+
 //
 // Page: 'Technical Requirements Checker'
 // --------------------------------------------------


javiereguiluz:
I don't usually like to restrict the behaviour of the native UI elements unless there is a good reason for it. But I'll wait to read more opinions. Thanks.

yceruto:
@javiereguiluz I understand your point. In this case we could avoid this behavior that affects the design when creating or editing an post. 

textarea-horizontal-resize

stof:
another solution is to add ``max-width: 100%``. This would keep the possibility to resize it as long as it fits in its parent

voronkovich:
@stof, you're rigth. The `resize` property is not supported by IE and Opera at all, so it's better to use `max-width`. https://developer.mozilla.org/en-US/docs/Web/CSS/resize

yceruto:
@stof, @voronkovich Fixed!

javiereguiluz:
Thanks @yceruto.

javiereguiluz:
@yceruto a minor comment for the future: when we make a change in the source stylesheet, we also need to recompile the styles that we ship with the application (`web/css/app.css`). To do that, just uncomment Assetic tags and execute `php app/console a:d --env=prod` I did that in https://github.com/symfony/symfony-demo/commit/43a80ae3e11b1f7a740aa139aa65ee11e0547610

yceruto:
@javiereguiluz I'm sorry, I will not forget for the next, thank you.

yceruto:
I'm sorry, I will not forget for the next, thank you.

PR #190, opened 2015-09-10
Add a console command for deleting users

voronkovich:
With example of using the console validator. See https://github.com/symfony/symfony-demo/pull/181

### DIFF
--- a/src/AppBundle/Command/DeleteUserCommand.php
+++ b/src/AppBundle/Command/DeleteUserCommand.php
@@ -1,0 +1,210 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace AppBundle\Command;
+
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
+use Doctrine\Common\Persistence\ObjectManager;
+
+/**
+ * A command console that deletes users and from the database.
+ * To use this command, open a terminal window, enter into your project
+ * directory and execute the following:
+ *
+ *     $ php app/console app:delete-user
+ *
+ * To output detailed information, increase the command verbosity:
+ *
+ *     $ php app/console app:delete-user -vv
+ *
+ * See http://symfony.com/doc/current/cookbook/console/console_command.html
+ *
+ * @author Oleg Voronkovich <oleg-voronkovich@yandex.ru>
+ */
+class DeleteUserCommand extends ContainerAwareCommand
+{
+    const MAX_ATTEMPTS = 5;
+
+    /**
+     * @var ObjectManager
+     */
+    private $em;
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            // a good practice is to use the 'app:' prefix to group all your custom application commands
+            ->setName('app:delete-user')
+            ->setDescription('Deletes users from the database')
+            ->setHelp($this->getCommandHelp())
+            // commands can optionally define arguments and/or options (mandatory and optional)
+            // see http://symfony.com/doc/current/components/console/console_arguments.html
+            ->addArgument('userIdentity', InputArgument::OPTIONAL, 'The username or email of an existing user')
+        ;
+    }
+
+    /**
+     * This method is executed before the interact() and the execute() methods.
+     * It's main purpose is to initialize the variables used in the rest of the
+     * command methods.
+     *
+     * Beware that the input options and arguments are validated after executing
+     * the interact() method, so you can't blindly trust their values in this method.
+     */
+    protected function initialize(InputInterface $input, OutputInterface $output)
+    {
+        $this->em = $this->getContainer()->get('doctrine')->getManager();
+    }
+
+    /**
+     * This method is executed after initialize() and before execute(). Its purpose
+     * is to check if some of the options/arguments are missing and interactively
+     * ask the user for those values.
+     *
+     * This method is completely optional. If you are developing an internal console
+     * command, you probably should not implement this method because it requires
+     * quite a lot of work. However, if the command is meant to be used by external
+     * users, this method is a nice way to fall back and prevent errors.
+     */
+    protected function interact(InputInterface $input, OutputInterface $output)
+    {
+        if (null !== $input->getArgument('userIdentity')) {
+            return;
+        }
+
+        // multi-line messages can be displayed this way...
+        $output->writeln('');
+        $output->writeln('Delete User Command Interactive Wizard');
+        $output->writeln('-----------------------------------');
+
+        // ...but you can also pass an array of strings to the writeln() method
+        $output->writeln(array(
+            '',
+            'If you prefer to not use this interactive wizard, provide the',
+            'arguments required by this command as follows:',
+            '',
+            ' $ php app/console app:delete-user username or email',
+            '',
+        ));
+
+        $output->writeln(array(
+            '',
+            'Now we\'ll ask you for the value of all the missing command arguments.',
+            '',
+        ));
+
+        // See http://symfony.com/doc/current/components/console/helpers/questionhelper.html
+        $console = $this->getHelper('question');
+
+        $userIdentity = $input->getArgument('userIdentity');
+        if (null === $userIdentity) {
+            $question = new Question(' > <info>Username or email</info>: ');
+            $question->setValidator(array($this, 'userIdentityValidator'));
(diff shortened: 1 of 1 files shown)


src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,210 @@
...
+    protected function configure()
+    {
+        $this
+            // a good practice is to use the 'app:' prefix to group all your custom application commands
+            ->setName('app:delete-user')
+            ->setDescription('Deletes users from the database')
+            ->setHelp($this->getCommandHelp())
+            // commands can optionally define arguments and/or options (mandatory and optional)
+            // see http://symfony.com/doc/current/components/console/console_arguments.html
+            ->addArgument('userIdentity', InputArgument::OPTIONAL, 'The username or email of an existing user')
```
stof: the argument is not optional. It is required (you are marking it as optional and then throwing an exception in case it is not provided)

src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,210 @@
...
+ *
+ * To output detailed information, increase the command verbosity:
+ *
+ *     $ php app/console app:delete-user -vv
+ *
+ * See http://symfony.com/doc/current/cookbook/console/console_command.html
+ *
+ * @author Oleg Voronkovich <oleg-voronkovich@yandex.ru>
+ */
+class DeleteUserCommand extends ContainerAwareCommand
```
theofidry: Personally I'm not a big fan of container aware as it make it difficult to test. But that's my personal opinion I don't know what's your policy on that

javiereguiluz: The official best practices don't mention the ContainerAware classes, so that means we don't discourage them.

src/AppBundle/Command/DeleteUserCommand.php
javiereguiluz: I have some concerns about using a single argument for two different things (usernames and emails). What if we simplify the command to only use usernames or only emails?

voronkovich: @javiereguiluz, what concerns you? Do you think it's a bad practice? We use a single class for few different things (AppBundle\Twig\AppExtension), and it's ok :)

javiereguiluz: @voronkovich the AppExtension class in fact only has one purpose: define all custom filters/functions of the application. If the app defined 200 filters, we would create different extensions/classes to group them logically.

In this case, we are using one single argument to store two different things: sometimes is a username and sometimes is an email.

voronkovich: @javiereguiluz, that makes sense. I will remove an email handling.

voronkovich: @stof, in the `add-user` command we use optional arguments too. Is it correct?

voronkovich:
Fixed.

src/AppBundle/Command/DeleteUserCommand.php
stof: @voronkovich this is also a mistake in this add-user command (I have not reviewed it when merging the PR)

javiereguiluz:
I'll merge this feature very soon. This is the last chance to make comments or ask questions about it. Thanks!

src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,204 @@
...
+            ->setHelp($this->getCommandHelp())
+            // commands can optionally define arguments and/or options (mandatory and optional)
+            // see http://symfony.com/doc/current/components/console/console_arguments.html
+            ->addArgument('username', InputArgument::REQUIRED, 'The username of an existing user')
+        ;
+    }
+
+    /**
+     * This method is executed before the interact() and the execute() methods.
+     * It's main purpose is to initialize the variables used in the rest of the
```
xabbuh: Its

src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,204 @@
...
+            '',
+            'Now we\'ll ask you for the value of all the missing command arguments.',
+            '',
+        ));
+
+        // See http://symfony.com/doc/current/components/console/helpers/questionhelper.html
+        $console = $this->getHelper('question');
+
+        $username = $input->getArgument('username');
+        if (null === $username) {
```
xabbuh: This check is needless. We would already have returned earlier before if the `username` argument were passed.

src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,204 @@
...
+            $username = $console->ask($input, $output, $question);
+            $input->setArgument('username', $username);
+        } else {
+            $output->writeln(' > <info>Username</info>: ' . $username);
+        }
+    }
+
+    /**
+     * This method is executed after interact() and initialize(). It usually
+     * contains the logic to execute to complete this command task.
```
xabbuh: I would remove "to execute" here.

src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,204 @@
...
+        ));
+
+        $output->writeln(array(
+            '',
+            'Now we\'ll ask you for the value of all the missing command arguments.',
+            '',
+        ));
+
+        // See http://symfony.com/doc/current/components/console/helpers/questionhelper.html
+        $console = $this->getHelper('question');
```
xabbuh: `$console` doesn't seem the best fitting variable name to me.

src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,204 @@
...
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $startTime = microtime(true);
+
+        $username = $input->getArgument('username');
+
+        $this->usernameValidator($username);
+
+        $user = null;
```
xabbuh: not needed

src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,204 @@
...
+        $repository = $this->entityManager->getRepository('AppBundle:User');
+
+        $user = $repository->findOneBy(array('username' => $username));
+
+        if (null === $user) {
+            throw new \RuntimeException(sprintf('User with username "%s" not found.', $username));
+        }
+
+        $userId = $user->getId();
```
xabbuh: There is no advantage in having the `$userId` variable.

src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,204 @@
...
+        if ($output->isVerbose()) {
+            $finishTime = microtime(true);
+            $elapsedTime = $finishTime - $startTime;
+
+            $output->writeln(sprintf('[INFO] Elapsed time: %.2f ms', $elapsedTime*1000));
+        }
+    }
+
+    /**
+     * This internal method should be private, but it's declared as public to
```
xabbuh: [...] declared public [...]

src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,204 @@
...
+     * maintain PHP 5.3 compatibility when using it in a callback.
+     *
+     * @internal
+     */
+    public function usernameValidator($username)
+    {
+        if (empty($username)) {
+            throw new \Exception('The username can not be empty');
```
xabbuh: missing period at the end of the exception message

src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        if (1 !== preg_match('/^[a-z_]+$/', $username)) {
+            throw new \Exception('The username must contain only lowercase latin characters and underscores');
```
xabbuh: same here

src/AppBundle/Command/DeleteUserCommand.php
voronkovich: @xabbuh, I think you're wrong. See the doctrine documentation: http://doctrine-orm.readthedocs.org/en/latest/reference/working-with-objects.html#removing-entities

```After an entity has been removed its in-memory state is the same as before the removal, except for generated identifiers.```

**Comment by voronkovich** [comment]
Fixed. Thanks, @xabbuh!

File: src/AppBundle/Command/DeleteUserCommand.php [hunk 12]
```diff
@@ -0,0 +1,201 @@
...
+    private $entityManager;
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            // a good practice is to use the 'app:' prefix to group all your custom application commands
+            ->setName('app:delete-user')
```
xabbuh: I would have chosen something like `app:user:delete`, but this is in line with the `AddUserCommand`. So we could discuss changing both names later if anyone else agrees with me.

xabbuh:
:+1:

javiereguiluz:
@voronkovich sorry it took me son long to merge this feature. I'm going to merge it but I'm also going to propose some changes in a separate pull request. Thanks!

PR #192, opened 2015-09-11
Add tags to posts with many-to-many relation 

### LINKED ISSUE #44: Add tags to posts with many-to-many relation
Hi there!
What about to add `Tag` entity relates **many-to-many** with `Post`?
I think it will be very useful for beginners.
I vote :+1: for it. What do you think?

rasanga:
This fixes #44 still need adding tags to admin back-end and I will work on that later.

## Posts list view
Post list view

## Post show view
Post show view


src/AppBundle/Entity/Tag.php
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Add post
+     *
+     * @param Post $post
+     *
+     * @return Tag
+     */
+    public function addPost(Post $post)
+    {
+        $this->posts[] = $post;
```
voronkovich: You should check whether the post already added, because Doctrine's collection doesn't do this

rasanga: so it would be as following?

if (false === $this->posts->indexOf($post)) {
    $this->posts[] = $post;
}

voronkovich: It should look like this:

```php
public function addTag(Tag $tag)
{
    if (!$this->tags->contains($tag)) {
        $this->tags->add($tag);
        $tag->addPost($this);
    }

    return $this;
}

public function removeTag(Tag $tag)
{
    if ($this->tags->contains($tag)) {
        $this->tags->removeElement($tag);
        $tag->removePost($this);
     }

    return $this;
}    
```

rasanga: @voronkovich thanks, didn't think of contains(), I will amend the change

voronkovich: For example https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Product/Model/Product.php#L269

rasanga: nice example, thanks :+1:

src/AppBundle/DataFixtures/ORM/LoadFixtures.php
```diff
@@ -91,11 +96,114 @@ private function loadPosts(ObjectManager $manager)
             }
 
             $manager->persist($post);
+
+            $this->addReference('post-' . $i, $post);
```
stof: using the reference system while being in the same fixture class is overkill: you don't clear the entity manager, so you could simply keep the same objects. Return an array of posts from this method, and pass them to the other method

src/AppBundle/Entity/Tag.php
```diff
@@ -0,0 +1,124 @@
...
+     */
+    private $name;
+
+    /**
+     * @var Collection|Post[]
+     *
+     * @ORM\ManyToMany(
+     *      targetEntity="Post",
+     *      mappedBy="tags"
+     * )
```
stof: Do we really need this relation here ? bidirectional relations are more complex to handle, and the Doctrine team recommendation is to prefer unidirectional relations anywhere you can

src/AppBundle/DataFixtures/ORM/LoadFixtures.php
rasanga: Good thinking, for while I am wondering why aren't we using multiple fixture classes (fixture class for each entity class) in this example?

src/AppBundle/Entity/Tag.php
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Remove post
+     *
+     * @param Post $post
+     */
+    public function removePost(Post $post)
+    {
+        $this->posts->removeElement($post);
```
stof: your adder and remover are broken: this class is the inversed side of the relation, and so changes in the collection are not tracked by Doctrine. You need to update the owning side otherwise things won't be saved in the DB (and this is one of the reason to avoid bidirectional relations btw)

src/AppBundle/Entity/Post.php
```diff
@@ -179,4 +192,38 @@ public function setSummary($summary)
...
+    /**
+     * Add tag
+     *
+     * @param Tag $tag
+     *
+     * @return Post
+     */
+    public function addTag(Tag $tag)
+    {
+        $this->tags[] = $tag;
```
stof: you need to check for duplicates here

src/AppBundle/Entity/Post.php
```diff
@@ -80,6 +82,17 @@ class Post
...
+    /**
+     * @var Collection|Tag[]
+     *
+     * @ORM\ManyToMany(
+     *      targetEntity="Tag",
+     *      inversedBy="posts"
+     * )
+     * @ORM\JoinTable(name="posts_tags")
+     */
+    private $tags;
```
stof: you need to initialize the collection in the constructor

src/AppBundle/Entity/Tag.php
```diff
@@ -0,0 +1,124 @@
...
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * Tag
+ *
+ * @ORM\Entity(repositoryClass="AppBundle\Repository\TagRepository")
+ *
+ * Defines the properties of the User entity to represent the application users.
```
stof: which user entity ?

src/AppBundle/Entity/Tag.php
```diff
@@ -0,0 +1,124 @@
...
+ * See http://symfony.com/doc/current/book/doctrine.html#creating-an-entity-class
+ *
+ * Tip: if you have an existing database, you can generate these entity class automatically.
+ * See http://symfony.com/doc/current/cookbook/doctrine/reverse_engineering.html
```
stof: I would remove this tip. Reverse engineering in doctrine is a poor-man feature and is not the way the ORM is meant to be used

src/AppBundle/Entity/Tag.php
```diff
@@ -0,0 +1,124 @@
...
+     * @ORM\Column(name="id", type="integer")
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="AUTO")
+     */
+    private $id;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column(name="tag_name", type="string", length=255)
```
stof: why not keeping ``name`` as column name ?

src/AppBundle/Entity/Post.php
theofidry: Certainly checking for duplicates or setting the relation for the other side avoid lots of errors but it can be a very heavy task for large collection.

src/AppBundle/Entity/Tag.php
stof: I can answer: we don't need it. You never use it (and it does not really make sense to have it anyway as tags don't need to be aware of posts).

src/AppBundle/Entity/Tag.php
rasanga: Because `name' is a reserved keyword in most DB servers

theofidry: why putting `name`? Let's keep it DRY and let the default value operate

src/AppBundle/Entity/Post.php
stof: if you don't check for duplicates here, you will get an exception when flushing because the constraint will fail in the DB

src/AppBundle/Entity/Tag.php
stof: it is not a reserved keyword in any platform supported by Doctrine (see https://github.com/doctrine/dbal/tree/master/lib/Doctrine/DBAL/Platforms/Keywords for the reserved keywords on each platform).
And the demo runs on MySQL anyway, so only mysql reserved keywords are important.

rasanga:
Fixed all issues mentioned in code review.

app/Resources/views/blog/post_show.html.twig
```diff
@@ -5,6 +5,13 @@
 {% block main %}
     <h1>{{ post.title }}</h1>
 
+    <p>
+        {% if 0 < post.tags | length %}Tags: {% endif %}
```
xabbuh: I think should put the `if` block around the paragraph to avoid rendering an empty one (the same can be done above).

src/AppBundle/DataFixtures/ORM/LoadFixtures.php
```diff
@@ -29,8 +31,9 @@
  *
  * @author Ryan Weaver <weaverryan@gmail.com>
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ * @author Rasanga Perera <rasangaperera@gmail.com>
  */
-class LoadFixtures implements FixtureInterface, ContainerAwareInterface
+class LoadFixtures extends AbstractFixture implements FixtureInterface, ContainerAwareInterface
```
xabbuh: This is not needed, is it?

src/AppBundle/Entity/Post.php
```diff
@@ -80,10 +82,21 @@ class Post
      */
     private $comments;
 
+    /**
+     * @var Collection|Tag[]
```
xabbuh: I wouldn't use the `Collection` type here. Client code shouldn't need to care about the actual type (it's sufficient to know that there are traversable tags).

voronkovich: @xabbuh, but collections support many useful methods: `each`, `map`, `filter` and `partition`

xabbuh: Sure, but you then tie code using the entities to your internal implementation and I don't think we should encourage that.

theofidry: As far as I can tell the default is `ArrayCollection`. You should either consider it as a bad practice and do it with just `array` or what you consider better; or use the default and show how to do it right

voronkovich: Some examples:
https://github.com/elcodi/elcodi/blob/master/src/Elcodi/Component/Product/Entity/Product.php#L506
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Archetype/Model/ArchetypeInterface.php#L39

src/AppBundle/DataFixtures/ORM/LoadFixtures.php
rasanga: Good shout, no since I removed getReference() it's redundant

app/Resources/views/blog/post_show.html.twig
rasanga: I agree

src/AppBundle/Entity/Post.php
rasanga: So, what's the verdict?

I suggest we should keep the `Collection` type hint

voronkovich: @rasanga yes, let's keep them as is

app/Resources/views/blog/index.html.twig
```diff
@@ -10,6 +10,13 @@
                     {{ post.title }}
                 </a>
             </h2>
+            {% if 0 < post.tags | length %}
```
noniagriconomie: What about updating the repository to load with tags? To avoid lazy loading
And btw, inform what is lazy loading :)

src/AppBundle/Entity/Post.php
```diff
@@ -180,4 +193,56 @@ public function setSummary($summary)
...
+     * Remove tag
+     *
+     * @param Tag $tag
+     *
+     * @return Post
+     */
+    public function removeTag(Tag $tag)
+    {
+        if (!$this->hasTag($tag)) {
+            $this->tags->removeElement($tag);
```
versh23: is it correct condition?

voronkovich: @versh23, no. It should be just `$this->tags->removeElement($tag);`, without any checkings because the `removeElement` function doesn't throw an exception when the element doesn't exist.

PR #193, opened 2015-09-11
Rename $em to $entityManager

voronkovich:


rasanga:
:+1: I like the change

javiereguiluz:
@voronkovich thanks for making this change. Although I don't agree with it, I acknowledge that most developers prefer this variable name. Thanks!

PR #194, opened 2015-09-11
feature #191 Add confirmation message before to delete a post

### LINKED ISSUE #191: Delete confirmation
Why not create a confirmation message before deleting a post?. Something like this:

**Option One:** Using javascript into html of the delete button:
```html
<input type="submit" value="Delete post" class="btn btn-lg btn-block btn-danger" onclick="confirm('Are you sure to delete this post?') || event.preventDefault()">
```
**Option Two**: Create a modal window using Twitter Bootstrap modal.
**Option Three**: Create a new page with the appropriate message. So, the current delete button was converted in an anchor to this page.

Or is there some reason why this behavior is not used?

yceruto:
The confirmation dialog is meant to ensure that a user intends to perform an action. This is often done in cases where actions cannot be undone (such as deleting something) or when something substantial will happen (like sending email updates, perhaps).

The user will be responsible for choosing "accept" or "cancel" and the action will not be taken if they choose the latter. You can also use the confirmation dialog to give users any important message before they make their decision.

Fixed #191

### DIFF
--- a/app/Resources/views/admin/blog/_form.html.twig
+++ b/app/Resources/views/admin/blog/_form.html.twig
@@ -6,7 +6,7 @@
     {{ form_start(form, { attr: { novalidate: 'novalidate' } }) }}
 #}
 
-{% if show_confirmation_message is defined %}
+{% if show_confirmation_message is defined and show_confirmation_message == true %}
     {% set attr = {'data-confirmation-message': 'action.delete_post_confirmation'|trans} %}
 {% endif %}
 
--- a/app/Resources/views/admin/blog/_form.html.twig
+++ b/app/Resources/views/admin/blog/_form.html.twig
@@ -10,6 +10,9 @@
     {{ form_widget(form) }}
 
     <input type="submit" value="{{ button_label|default('label.create_post'|trans) }}"
+           {% if confirmation_message is defined %}
+               data-toggle="confirmation" data-message="{{ confirmation_message }}"
+           {% endif %}
            class="{{ button_css|default("btn btn-primary") }}" />
 
     {% if include_back_to_home_link is not defined or include_back_to_home_link == true %}
--- a/app/Resources/views/admin/blog/_form.html.twig
+++ b/app/Resources/views/admin/blog/_form.html.twig
@@ -14,9 +14,6 @@
     {{ form_widget(form) }}
 
     <input type="submit" value="{{ button_label|default('label.create_post'|trans) }}"
-           {% if confirmation_message is defined %}
-               data-toggle="confirmation" data-message="{{ confirmation_message }}"
-           {% endif %}
            class="{{ button_css|default("btn btn-primary") }}" />
 
     {% if include_back_to_home_link is not defined or include_back_to_home_link == true %}
--- a/app/Resources/views/admin/blog/_form.html.twig
+++ b/app/Resources/views/admin/blog/_form.html.twig
@@ -6,11 +6,11 @@
     {{ form_start(form, { attr: { novalidate: 'novalidate' } }) }}
 #}
 
-{% if confirmation_message is defined %}
-    {% set attr = {'data-confirmation-message': confirmation_message} %}
+{% if show_confirmation_message is defined %}
+    {% set attr = {'data-confirmation-message': 'action.delete_post_confirmation'|trans} %}
 {% endif %}
 
-{{ form_start(form, {attr: attr|default({})}) }}
+{{ form_start(form, { attr: attr|default({}) }) }}
     {{ form_widget(form) }}
 
     <input type="submit" value="{{ button_label|default('label.create_post'|trans) }}"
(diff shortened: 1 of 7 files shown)


app/Resources/views/base.html.twig
```diff
@@ -154,6 +154,12 @@
             <script>
                 $(document).ready(function() {
                     hljs.initHighlightingOnLoad();
+
+                    $('[data-toggle="confirmation"][data-message]').on('click', function (event) {
```
stof: this should not be done on click. It should be done on form submission. Otherwise other ways to submit the form will not trigger the confirmation

app/Resources/views/base.html.twig
```diff
@@ -154,6 +154,12 @@
...
+                        var message = $(this).data('message');
+
+                        return confirm(message) || event.preventDefault();
```
stof: returning does not make sense here. ``event.preventDefault()`` does not return.

app/Resources/views/base.html.twig
stof: and btw, using event delegation would be better than doing a DOM traversal on dom ready to find all such elements

app/Resources/views/base.html.twig
yceruto: thanks you @stof, right! Fixed.

app/Resources/views/base.html.twig
yceruto: thanks @stof fixed.

app/Resources/views/admin/blog/edit.html.twig
```diff
@@ -19,7 +19,8 @@
             form: delete_form,
             button_label: 'action.delete_post'|trans,
             button_css: 'btn btn-lg btn-block btn-danger',
-            include_back_to_home_link: false
+            include_back_to_home_link: false,
+            confirmation_message: 'action.delete_post_confirmation'|trans,
```
javiereguiluz: My concern with this feature is that we are using one variable (`confirmation_message `) for two different purposes: 1) enabling/disabling this feature, 2) defining the message to display if enabled.

Personally, multi-purpose variables makes me very nervous :cold_sweat:

Could we instead use a boolean `show_confirmation_message` option and then use the `'action.delete_post_confirmation'|trans` where needed to display the actual message contents?

yceruto: I understand your concern and I agree with you @javiereguiluz. :+1:

app/Resources/views/base.html.twig
```diff
@@ -154,6 +154,15 @@
             <script>
                 $(document).ready(function() {
                     hljs.initHighlightingOnLoad();
+
+                    $('form[data-confirmation-message]').on('submit', function (event) {
```
stof: my second comment about event delegation has not been taken into account though.

yceruto: stof sorry, I did it on purpose, you could tell me how to apply event delegation here? 

something like this?:
```javascript
$('body').delegate('form[data-confirmation-message]', 'submit', function (e) {...});
```

voronkovich: @yceruto As of jQuery 1.7, .delegate() has been superseded by the .on() method.

voronkovich: See good examples here: http://learn.jquery.com/events/event-delegation/

yceruto: thanks @voronkovich :+1:

app/Resources/views/base.html.twig
```diff
@@ -154,6 +154,15 @@
             <script>
                 $(document).ready(function() {
                     hljs.initHighlightingOnLoad();
+
+                    $('body').on('submit', 'form[data-confirmation-message]', function (event) {
```
stof: I generally use ``$(document)`` for delegation, which avoids a DOM query for ``'body'`` (even though jquery can probably optimize the DOM query for the body).

It also avoids the need to wait for DOM ready for that.

yceruto: :+1: fixed. thanks @stof :)

app/Resources/views/admin/blog/_form.html.twig
```diff
@@ -6,7 +6,11 @@
     {{ form_start(form, { attr: { novalidate: 'novalidate' } }) }}
 #}
 
-{{ form_start(form) }}
+{% if show_confirmation_message is defined and show_confirmation_message == true %}
```
javiereguiluz: @yceruto sorry for adding a new comment in this pull request which you already worked on a lot. I'm wondering if we could replace this long condition by just:

```twig
{% if show_confirmation_message|default(false) %}
```

I'm just asking. I don't really know which is the common practice in these cases.

voronkovich: Or we could use this:
```twig
{% set attr = show_confirmation_message|default(false) ? {'data-confirmation-message': 'action.delete_post_confirmation'|trans} %}
```

yceruto: javier do not worry, in this PR I learned a lot of comments. I am in favor of less code, more simple. fixed!

yceruto:
I released a new version of this PR using the bootstrap modal. https://github.com/symfony/symfony-demo/pull/202

javiereguiluz:
Closing it because we'll implement this using a modal window.

PR #195, opened 2015-09-11
set the focus on title when creating or editing a post

yceruto:
Perhaps it would be good to suggest to the user where to begin writing the post.

### DIFF
--- a/src/AppBundle/Form/PostType.php
+++ b/src/AppBundle/Form/PostType.php
@@ -39,7 +39,10 @@ class PostType extends AbstractType
         //     $builder->add('title', null, array('required' => false, ...));
 
         $builder
-            ->add('title', null, array('label' => 'label.title'))
+            ->add('title', null, array(
+                'attr' => array('autofocus' => true),
+                'label' => 'label.title',
+            ))
             ->add('summary', 'textarea', array('label' => 'label.summary'))
             ->add('content', 'textarea', array(
                 'attr' => array('rows' => 20),


src/AppBundle/Form/PostType.php
```diff
@@ -39,7 +39,10 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         //     $builder->add('title', null, array('required' => false, ...));
 
         $builder
-            ->add('title', null, array('label' => 'label.title'))
+            ->add('title', null, array(
+                'attr' => array('autofocus' => true),
```
rasanga: I prefer to keep attributes on the form view, but I would like to knows others intentions

voronkovich: @rasanga, I agree with you. Keeping attributes on the template is more flexible approach than this. But this PR just follows the existing code style. I think you could refactor the form templates and create a PR.

rasanga: @voronkovich good thinking let me create a PR with form attributes on the template

yceruto: @rasanga, I agree too, but as @voronkovich says, just follows the existing code style. https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Form/PostType.php#L45

Another way is using javascripts `$('#app_post_title').focus();` within edit and new templates, but I still prefer this PR. Is simple and without dependencies.

rasanga: @yceruto I didn't know we were using the 'attr' tag on the form type.

javiereguiluz:
@yceruto thanks for helping us improve the demo application!

yceruto:
:+1:

PR #197, opened 2015-09-14
CodeExplorerBundle

voronkovich:
Very simple implementation, without any overheads and overkills :)
See https://github.com/symfony/symfony-demo/issues/196


src/ShowSourceBundle/Twig/SourceCodeExtension.php
```diff
@@ -49,7 +49,7 @@ public function getFunctions()
 
     public function showSourceCode(\Twig_Environment $twig, \Twig_Template $template)
     {
-        return $twig->render('default/_source_code.html.twig', array(
+        return $twig->render('ShowSourceBundle::_source_code.html.twig', array(
```
javiereguiluz: I'm not sure, but maybe we should rename `_source_code.html.twig` to `source_code.html.twig`. The reason is that from the point of view of the new bundle, this is not a fragment but a "real template" used to display its contents.

javiereguiluz:
I like these changes. My only concern is about the name of the bundle: `ShowSourceBundle` looks too specific to me. I know that this bundle won't grow in the future with new features (at least not with lots of features), but nevertheless I'd prefer a name which is a bit more generic.

src/ShowSourceBundle/Twig/SourceCodeExtension.php
stof: I agree with you

voronkovich:
@javiereguiluz, I would suggest the `ShowSourceCodeBundle`. What do you think?

voronkovich:
`CodeExplorerBundle` sounds great for me

src/CodeExplorerBundle/Twig/SourceCodeExtension.php
```diff
@@ -49,7 +49,7 @@ public function getFunctions()
 
     public function showSourceCode(\Twig_Environment $twig, \Twig_Template $template)
     {
-        return $twig->render('default/_source_code.html.twig', array(
+        return $twig->render('CodeExplorerBundle::source_code.html.twig', array(
```
javiereguiluz: My final comment before merging this change: could/should we use `@CodeExplorer/source_code.html.twig` notation instead?  Symfony still has't made an official statement, but the namespaced syntax seems to be favored for modern apps.

voronkovich: @javiereguiluz,  I think we should use it here. This notation more clearly for understanding. I like it.

src/CodeExplorerBundle/DependencyInjection/CodeExplorerExtension.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CodeExplorerBundle\DependencyInjection;
+
+use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Loader;
```
xabbuh: imo you should import the class name and not the parent namespace here

voronkovich: @xabbuh, thanks! I've fixed this issue.

src/CodeExplorerBundle/Resources/config/services.yml
```diff
@@ -0,0 +1,13 @@
...
+        class:     CodeExplorerBundle\Twig\SourceCodeExtension
+        arguments: ["@twig.loader", "%kernel.root_dir%"]
+        tags:
+            - { name: twig.extension }
+
+    code_explorer.controller_listener:
+        class:     CodeExplorerBundle\EventListener\ControllerListener
+        arguments: ["@code_explorer.twig.source_code_extension"]
+        tags:
+            - { name: 'kernel.event_listener', event: kernel.controller, method: registerCurrentController }
```
xabbuh: single quotes around the argument are not needed

src/CodeExplorerBundle/Resources/config/services.yml
```diff
@@ -0,0 +1,13 @@
...
+    code_explorer.twig.source_code_extension:
+        public:    false
+        class:     CodeExplorerBundle\Twig\SourceCodeExtension
+        arguments: ["@twig.loader", "%kernel.root_dir%"]
+        tags:
+            - { name: twig.extension }
+
+    code_explorer.controller_listener:
+        class:     CodeExplorerBundle\EventListener\ControllerListener
+        arguments: ["@code_explorer.twig.source_code_extension"]
```
xabbuh: double quotes around the argument are not needed here

voronkovich: @xabbuh, you're right. I've removed all quotes.

javiereguiluz:
@voronkovich you did a really nice work with this pull request. Sorry it took so long to merge it.

PR #199, opened 2015-09-17
Redirect a client to an appropriate locale

voronkovich:
A simple example of how to use the event listeners

### DIFF
--- a/src/AppBundle/EventListener/RedirectToPreferedLocaleListener.php
+++ b/src/AppBundle/EventListener/RedirectToPreferedLocaleListener.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace AppBundle\EventListener;
+
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\Routing\RouterInterface;
+
+/**
+ * This listener subscribes to the 'kernel.request' event to redirect a client
+ * to an appropriate locale, according to browser settings.
+ *
+ * See http://symfony.com/doc/current/components/http_kernel/introduction.html#the-kernel-request-event
+ *
+ * @author Oleg Voronkovich <oleg-voronkovich@yandex.ru>
+ */
+class RedirectToPreferedLocaleListener
+{
+    /**
+     * @var RouterInterface
+     */
+    private $router;
+
+    /**
+     * List of supported locales.
+     *
+     * @var array
+     */
+    private $locales = array();
+
+    public function __construct($locales = '', RouterInterface $router)
+    {
+        $this->locales = explode('|', $locales);
+        $this->router = $router;
+    }
+
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        $request = $event->getRequest();
+
+        if ('/' !== $request->getPathInfo()) {
+            return;
+        }
+
+        $preferedLanguage = $request->getPreferredLanguage($this->locales);
+
+        if ('' !== $preferedLanguage && $request->getDefaultLocale() !== $preferedLanguage) {
+            $response = new RedirectResponse($this->router->generate('homepage', array('_locale' => $preferedLanguage)));
+            $event->setResponse($response);
+        }
+    }
+}
--- a/src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
+++ b/src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
@@ -49,7 +49,7 @@ class RedirectToPreferredLocaleListener
     {
         $request = $event->getRequest();
 
-        if ('/' !== $request->getPathInfo()) {
+        if (!$event->isMasterRequest() || '/' !== $request->getPathInfo()) {
             return;
         }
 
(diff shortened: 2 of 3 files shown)


src/AppBundle/EventListener/RedirectToPreferedLocaleListener.php
```diff
@@ -0,0 +1,52 @@
...
+    private $router;
+
+    /**
+     * List of supported locales.
+     *
+     * @var array
+     */
+    private $locales = array();
+
+    public function __construct($locales = '', RouterInterface $router)
```
stof: typehint the UrlGeneratorInterface here as this is what you really need, not the full RouterInterface

src/AppBundle/EventListener/RedirectToPreferedLocaleListener.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /**
+     * @var RouterInterface
+     */
+    private $router;
+
+    /**
+     * List of supported locales.
+     *
+     * @var array
```
stof: ``string[]``

src/AppBundle/EventListener/RedirectToPreferedLocaleListener.php
```diff
@@ -0,0 +1,52 @@
...
+
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        $request = $event->getRequest();
+
+        if ('/' !== $request->getPathInfo()) {
+            return;
+        }
+
+        $preferedLanguage = $request->getPreferredLanguage($this->locales);
```
stof: typo in the variable name: ``preferredLanguage``

src/AppBundle/EventListener/RedirectToPreferedLocaleListener.php
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * This listener subscribes to the 'kernel.request' event to redirect a client
+ * to an appropriate locale, according to browser settings.
+ *
+ * See http://symfony.com/doc/current/components/http_kernel/introduction.html#the-kernel-request-event
+ *
+ * @author Oleg Voronkovich <oleg-voronkovich@yandex.ru>
+ */
+class RedirectToPreferedLocaleListener
```
stof: same typo in the class name btw

voronkovich:
Fixed. Thanks, @stof!

src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /**
+     * @var UrlGeneratorInterface
+     */
+    private $urlGenerator;
+
+    /**
+     * List of supported locales.
+     *
+     * @var string[]
```
yceruto: array|string[]

src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    private $locales = array();
+
+    public function __construct($locales = '', UrlGeneratorInterface $urlGenerator)
```
yceruto: Add PHPDoc blocks for all methods, and functions

src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
stof: not needed. the phpdoc standard already says that ``string[]`` is an array of strings

voronkovich: @yceruto, an empty array is an array of strings but without any one.

From phpdoc documentation:

```array - A collection of variables of unknown type.```

http://www.phpdoc.org/docs/latest/guides/types.html

yceruto: :+1:

src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
voronkovich: I think in this case it's not needed because the parameters types can be derived from the type hint

yceruto: right @voronkovich, but http://symfony.com/doc/current/contributing/code/standards.html does not says exceptions.

voronkovich: @yceruto yes, you're rigth. No exceptions.

voronkovich:
I'verefactored the listener, because Request::getPreferredLanguage and Request::getDefaultLocale don't behave as I expected.

javiereguiluz:
My concern with this feature is that I don't think this is a good practice we want to promote. First, executing another listener for each and every request of the application introduces a non-negligible performance hit.

Second, if I understand this feature well, I'm not sure this is what the user expects. When accessing the homepage, I think it's OK to redirect to the homepage of the user's language. But when accessing directly to an inner page, I think it's better to not change its language inadvertedly.

I propose to simplify this feature and apply it only for the homepage and without using a listener. But before doing any change, let's wait for other people's opinions.

voronkovich:
@javiereguiluz, this feature woks only when requesting the homepage. I've added a comment about it.

javiereguiluz:
@voronkovich you are right! Sorry for not having looked at your code with more detail.

javiereguiluz:
Merged! Thanks @voronkovich

PR #201, opened 2015-09-17
Adding translation config to pagination controls

### LINKED ISSUE #198: Translation for sliding pagination controls
The `KnpPaginatorBundle:Pagination:twitter_bootstrap_v3_pagination.html.twig` template configured for pagination has `Previous` and `Next` strings ready for translation, what we should do: translate this strings directly or override this template and include new keys like `pagination.previous` & `pagination.next` ?

ccab:
Fixed #198


javiereguiluz:
I don't like the idea of creating such a long template just to translate the pagination controls. I know that Pagerfanta includes dozens of translations. Does anyone that uses KnpPaginatorBundle know if translations are supported anyhow or will they be supported anytime soon?

davialexandre:
Looking at the templates in KnpPaginatorBundle I see they do support translations. They are using 'Next' and 'Previous' as IDs, so all we have to do is add translations for them

javiereguiluz:
@davialexandre I was referring to have the actual contents translated, not just ready to be translated. Pagerfanta provides tens of translation files, which makes sense because all of us are probably going to translate these strings the same way.

davialexandre:
That makes sense. In that case, they don't support it and it doesn't seem that they plan to support it anytime soon, given this closed issue:
https://github.com/KnpLabs/KnpPaginatorBundle/issues/192

ccab:
So we should change the paginator to Pagerfanta ?

javiereguiluz:
@ccab I'd prefer not to do that ... but I'd like to implement a better way to translate the paginator without having to recreate its entire template.

ccab:
@javiereguiluz I agree, but as you can see they don't care about this, I could create a issue for this topic in the KnpPaginatorBundle repo but it could be a slow way, how we could ask help to the community seasoned developers ?

javiereguiluz:
I've opened a "feature request" to ask for translations in KnpPaginatorBundle: https://github.com/KnpLabs/KnpPaginatorBundle/issues/334

Meanwhile, @ccab could you try just to provide the translations in a regular XLF file instead of recreating the entire template. The original template uses `'Next'|trans` and `'Previous'|trans`, so it's ready to be translated.

ccab:
Ok @javiereguiluz I'll do that, I really appreciated your patience

javiereguiluz:
@ccab thanks for woring on this and for going through all the suggested changes. Now let's see if our friends at KnpPaginatorBundle accept to translate these strings themselves.

PR #202, opened 2015-09-18
confirmation message using bootstrap modal

yceruto:
Fixed #191. Option 2: Improving design with bootstrap modal.

confirmation-modal

PR #203, opened 2015-09-19
Use a modal window to confirm the blog post deletion

javiereguiluz:
This pull request finished #202 by making some tweaks to the design of the delete dialog.

**Before**

We used the classic Yes/No dialog mostly used in Windows systems:

dialog_before

**After**

I propose to use the Delete/Cancel dialog used in Apple systems, because it's more user friendly:

dialog_after

I've just tweaked this dialog contents. The rest of the pull request is the result of the great work made by @yceruto.


yceruto:
:+1: @javiereguiluz  Looks good for me.

app/Resources/views/admin/blog/_form.html.twig
```diff
@@ -6,13 +6,18 @@
...
+{% endif %}
+
+{{ form_start(form, { attr: attr|default({}) }) }}
     {{ form_widget(form) }}
 
     <input type="submit" value="{{ button_label|default('label.create_post'|trans) }}"
            class="{{ button_css|default("btn btn-primary") }}" />
 
-    {% if include_back_to_home_link is not defined or include_back_to_home_link == true %}
+    {% if include_back_to_home_link|default(false) %}
```
xabbuh: Is it intended to change the behaviour here?

yceruto: The behavior does not change, what was done was to shorten the condition by suggestion of @javiereguiluz.
Discussion: https://github.com/symfony/symfony-demo/pull/194#discussion_r39792102

xabbuh: The condition was `true` before when the `include_back_to_home` variable was not defined. Now it will be false. If you want to keep the same behaviour, the argument passed to the `default` filter must be `true`.

stof: @xabbuh this would break the feature entirely, because ``false`` is considered to be an empty value in Twig, and so would trigger the default filter.

xabbuh: @stof You're right, I would then suggest to not modify this check at all.

stof: I agree with you

yceruto: @xabbuh the behavior does not change because of the following changes in these files:
https://github.com/javiereguiluz/symfony-demo/blob/pr/202/app/Resources/views/admin/blog/edit.html.twig#L13
https://github.com/javiereguiluz/symfony-demo/blob/pr/202/app/Resources/views/admin/blog/new.html.twig#L9

javiereguiluz: I've tested this code in my machine and things seem to work as expected. I'm not sure I understand what's wrong with the code.

xabbuh: Well, if you are show that the variable will now always be passed when this part should be shown, the change will work just fine. Though I would then propose to simplify it to:

```twig
{% if include_back_to_home_link %}
```

javiereguiluz: I prefer not to change it by `{% if include_back_to_home_link %}` because any variable could be not defined at some point. The `if ... |default` would never fail (the original code wouldn't either).

PR #205, opened 2015-09-20
Render global errors of comment form

xelaris:
Without this addition the message about a spam comment isn't rendered, when entering an `@` character. (see `@Assert\IsTrue()` validation on `Comment::isLegitComment()`)

### DIFF
--- a/app/Resources/views/blog/_comment_form.html.twig
+++ b/app/Resources/views/blog/_comment_form.html.twig
@@ -16,12 +16,14 @@
     <fieldset>
         <legend>{{ 'title.add_comment'|trans }}</legend>
 
+        {# Render any "global" errors (e.g. when a constraint on a public getter method failed). #}
+        {{ form_errors(form) }}
+
         <div class="form-group{% if not form.vars.valid %} has-error{% endif %}">
             {{ form_label(form.content, 'label.content', { label_attr: { class: 'hidden' }}) }}
 
-            {% if not form.vars.valid %}
-                {{ form_errors(form.content) }}
-            {% endif %}
+            {# Render any errors for the "content" field (e.g. when a class property constraint failed). #}
+            {{ form_errors(form.content) }}
 
             {{ form_widget(form.content, { attr: { rows: 10 } }) }}
         </div>


app/Resources/views/blog/_comment_form.html.twig
```diff
@@ -20,6 +20,7 @@
             {{ form_label(form.content, 'label.content', { label_attr: { class: 'hidden' }}) }}
 
             {% if not form.vars.valid %}
+                {{ form_errors(form) }}
                 {{ form_errors(form.content) }}
```
voronkovich: @xelaris, I would remove this string:
```twig
{{ form_errors(form.content) }}
```
Now it's not needed.
And I think you should remove the `if`  statement,  because the  `form_errors` doesn't produce any output when no errors occurred.

xelaris: @voronkovich  I don't think `{{ form_errors(form.content) }}` is redundant, since `form_errors(form)` renders any "global" errors only, whereas `{{ form_errors(form.content) }}` renders the field error (e.g. when entering a comment with too few characters).
I think you are right about the `if` statement. I'm going to update the PR.

xelaris:
The PR has been updated. The `if` statement is removed as suggested by @voronkovich and I added comments to explain the difference between the two `form_errors()` calls.

app/Resources/views/blog/_comment_form.html.twig
voronkovich: @xelaris, thanks for the explanation!

app/Resources/views/blog/_comment_form.html.twig
```diff
@@ -16,12 +16,14 @@
     <fieldset>
         <legend>{{ 'title.add_comment'|trans }}</legend>
 
+        {# Render any "global" errors (e.g. when a constraint on a public getter method failed). #}
+        {{ form_errors(form) }}
+
         <div class="form-group{% if not form.vars.valid %} has-error{% endif %}">
```
voronkovich: @xelaris, please, change the `form.vars.valid` to  a `form.content.vars.valid`

javiereguiluz: @voronkovich you are right. I've made the change in #207. Thanks.

javiereguiluz:
@xelaris thanks for fixing this bug. I've finished this PR with very minor changes in #207. I've reused your original commit, so you'll get full credit for your work. Thanks.

PR #208, opened 2015-09-25
Remove not needed quoutes from services definitions

voronkovich:


javiereguiluz:
It's true that they are not strictly needed. However, in the Symfony docs we always put these quotes because otherwise the YAML syntax highlighting doesn't work (e.g. http://symfony.com/doc/current/book/service_container.html#service-parameters)

javiereguiluz:
Thanks @voronkovich.

PR #213, opened 2015-10-06
Add missing Ukrainian translations

bocharsky-bw:
Add UK translations for modal popup


javiereguiluz:
Thank you @bocharsky-bw.

PR #215, opened 2015-10-06

davialexandre:
Translations for the delete post modal window


javiereguiluz:
Thank you @davialexandre.

PR #218, opened 2015-10-08
Tweaked the new app:delete-user command

javiereguiluz:
Following the merge of #190, I propose these changes because I don't think all the commands should be exactly the same:

  * Removed the explanations that are the same as in `app:add-user` command.
  * Removed the `isVerbose()` support for this command because we already show that in `app:add-user`
  * Removed the utility method that displays the command help.


voronkovich:
:+1:

PR #219, opened 2015-10-08
Update the way to retrieve the template source

stof:
Latest Twig versions allow to get the source from the Twig_Template object itself, making it work with any loader (and without needing the loader).

I also switched to private visibility in the CodeExplorerBundle according to the coding standards.

### DIFF
--- a/src/CodeExplorerBundle/EventListener/ControllerListener.php
+++ b/src/CodeExplorerBundle/EventListener/ControllerListener.php
@@ -28,10 +28,7 @@ use CodeExplorerBundle\Twig\SourceCodeExtension;
  */
 class ControllerListener
 {
-    /**
-     * @var SourceCodeExtension
-     */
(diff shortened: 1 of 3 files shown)


javiereguiluz:
I like this change. Thanks @stof.

I have one question: `->getSource()` is available only on Twig `>= 1.22.0`. Should we add a constraint for `twig/twig` in composer.json?

stof:
@javiereguiluz well, I considered this, and I would have updated the requirement if there was already one.
But this a project with locked dependencies, so it may be fine to write code against the locked dependencies (if you consider it is not fine, I can argue that you should then add explicit dependencies for all packages used directly in your own codebase)

javiereguiluz:
Thanks for this improvement. Your explanation about the dependencies convinced me, so I'm merging it as it is.

PR #221, opened 2015-10-09
Add a comment describing the code explorer shortcoming

stof:
The goal is to avoid letting people think that this is the generic way to access the file storing a template, because it is not.


javiereguiluz:
I like this help note!  Merged. Thanks.

PR #228, opened 2015-10-15
Fix #227: remove a wrong comment in the source_code.html.twig

### LINKED ISSUE #227: Comment in src/CodeExplorerBundle/Resources/views/source_code.html.twig contradicts file name
The comment states that fragments should be prefixed with an underscore. Is this just a typo in the file name?

voronkovich:


stof:
Can you move this comment to the ``_flash_messages.html.twig`` template, so that we still have this explanation in a template where it makes sense ?

voronkovich:
@stof, fixed!

javiereguiluz:
@voronkovich thanks for this fix.

PR #232, opened 2015-10-20
German pagination control translations

xabbuh:


xabbuh:
By the way: should we add them to the English file too in case someone uses it as a reference to search for missing translations in their language?

xabbuh:
ping @OskarStark

stof:
IMO, we should have all translations in the English files, yes. this way, you can easily find missing ones

xabbuh:
see #233

OskarStark:
:+1: @xabbuh, but i would like to use translations keys, instead of translating the language directly.

``previous_label`` or ``label_previous``

Oskar

xabbuh:
@OskarStark We can't control that (the string to be translated comes from a third-party library).

OskarStark:
oh, didn't know that :+1:

javiereguiluz:
Thanks @xabbuh.

PR #233, opened 2015-10-20
Add all translatable strings in the English translation file

xabbuh:
This way, the English translation file can be used as a reference to
find missing translations in all other languages.

### DIFF
--- a/app/Resources/translations/messages.en.xliff
+++ b/app/Resources/translations/messages.en.xliff
@@ -18,6 +18,14 @@
                 <source>mit_license</source>
                 <target>MIT License</target>
             </trans-unit>
+            <trans-unit id="previous">
+                <source>Previous</source>
+                <target>Previous</target>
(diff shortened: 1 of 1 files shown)


OskarStark:
i would prefer to use translation keys.

see: https://github.com/symfony/symfony-demo/pull/232#issuecomment-149547884

xabbuh:
@OskarStark We can't control that (the string to be translated comes from a third-party library).

OskarStark:
oh, didn't know that :+1:

OskarStark:
i think its better to provide this to the vendor bundle as a PR

KnpLabs/KnpPaginatorBundle#337

what do you thin?

ccab:
hello people I think @OskarStark initiative is the right one, as you can see in #201 @javiereguiluz asked for this a few days ago

javiereguiluz:
Thank you @xabbuh.

PR #236, opened 2015-10-21
Add an example about how to use the form's buttons

voronkovich:


src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -72,7 +72,14 @@ public function newAction(Request $request)
...
         $post->setAuthorEmail($this->getUser()->getEmail());
-        $form = $this->createForm(new PostType(), $post);
+
+        // We need a form builder to dynamically add a submit button.
+        // See http://symfony.com/doc/current/book/forms.html#submitting-forms-with-multiple-buttons
+        $form = $this
+            ->get('form.factory')
+            ->createBuilder(new PostType(), $post)
+            ->add('createAndShow', 'submit')
+            ->getForm();
```
yceruto: The same result with:
```
$form = $this->createForm(new PostType(), $post)
    ->add('createAndShow', 'submit');
```

voronkovich: @yceruto, `createForm` returns an instance of `Form` class, but we need an instance of `FormBuilder` to add a button dynamically.

yceruto: @voronkovich I tested it and it works. `Symfony\Component\Form::add`

app/Resources/views/admin/blog/new.html.twig
```diff
@@ -5,9 +5,21 @@
...
+        {{ form_row(form.publishedAt) }}
+
+        <input type="submit" value="{{ 'label.create_post'|trans }}" class="btn btn-primary" />
+
+        {{ form_widget(form.createAndShow, { label: 'label.create_post_and_show'|trans }) }}
+
+        <a href="{{ path('admin_post_index') }}" class="btn btn-link">
+            {{ 'action.back_to_list'|trans }}
+        </a>
+{{ form_end(form) }}
```
yceruto: minor details: indent this line, please

src/AppBundle/Controller/Admin/BlogController.php
voronkovich: @yceruto, you're right! Thanks!

app/Resources/views/admin/blog/new.html.twig
voronkovich: @yceruto, fixed.

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -93,6 +96,10 @@ public function newAction(Request $request)
             // See http://symfony.com/doc/current/book/controller.html#flash-messages
             $this->addFlash('success', 'post.created_successfully');
 
+            if ($form['createAndShow']->isClicked()) {
```
yceruto: this `$form['createAndShow']` is not wrong, but perhaps could bring confusion, rather use `$form->get('createAndShow')`: I have seen it used in the documentation always.

voronkovich: @yceruto, I think the `$form['createAndShow']` more short and clearly, let's wait for other people's opinions.

app/Resources/views/admin/blog/show.html.twig
```diff
@@ -5,6 +5,8 @@
 {% block main %}
     <h1>{{ post.title }}</h1>
 
+    {{ include('default/_flash_messages.html.twig') }}
```
stof: shoudln't flash messages be displayed in the layout instead of forcing each child page to displat them ?

voronkovich: @stof, I think you're right. I've moved the displaying of flash messages to the main layout.

src/AppBundle/Controller/Admin/BlogController.php
javiereguiluz: using `$form->get('createAndShow')` instead of `$form['createAndShow']` looks similar to other Symfony's practices ... but I know nothing about the Form component, so ignore my comment :)

voronkovich: @javiereguiluz,  the `Form` class implements the `ArrayAccess` interface, internally it uses the `get` method, so the `$form['createAndShow']` is equal to the `$form->get('createAndShow')`, but more short. http://github.com/symfony/form/blob/master/Form.php#L949

javiereguiluz: Yes, I guess it works. My comment was more about: "Symfony uses OOP/classes/methods and "others" use arrays". That's why I think `$form->get('createAndShow')` is more appropriate for Symfony. But again, I know nothing about this.

voronkovich: @yceruto, @javiereguiluz, ok, I've changed the `$form['createAndShow']` to the `$form->get('createAndShow')`

javiereguiluz:
@voronkovich I'd like to include this example to showcase the use of form buttons. However, the underlying behavior ("create and show") looks uncommon to me. What do you think about replacing the example to implement a more common "save and create new" feature? The app would save the created entity and it would redirect you to the "new" page again, so you can easily create lots of new elements of the same entity.

voronkovich:
@javiereguiluz, changed to "save and create new"

javiereguiluz:
@voronkovich thanks for implementing this feature. I've tested the last changes and it works great. Thanks!

PR #237, opened 2015-10-23
Finished Dutch Translation by Frank Beentjes

Fbeen:
Dont forget to add nl in the configuration file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@
 /vendor/
 /bin/
 /composer.phar
-/nbproject/private/
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /vendor/
 /bin/
 /composer.phar
+/nbproject/private/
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -8,3 +8,4 @@
 /vendor/
 /bin/
 /composer.phar
+/nbproject/private/
```
bocharsky-bw: I think this is a very personal and we should avoid it. BTW, many of devs use PhpStorm instead

bocharsky-bw:
You should revert changes in `web/config.php`.

Also I think we don't need to update `composer.lock` now. Better to update it when new version of Symfony will be avaliable. Anyway, `composer.lock` updates should be in separate PR.

.gitignore
Fbeen: Thank you for your comment.

javiereguiluz:
@Fbeen thanks for providing these translation fixes and congrats on your first contribution to this project!

javiereguiluz:
Please note that we use a tool that "squashes" commits before merging pull requests. That's why GitHub displays your pull request as "Closed" instead of "Merged". But trust me that your changes were correctly merged :)

PR #238, opened 2015-10-23
Fixing details

yceruto:
* Fixing imports (unused or missing).
* Adding type hinting.
* Fixed variable definition.

### DIFF
--- a/src/AppBundle/Command/DeleteUserCommand.php
+++ b/src/AppBundle/Command/DeleteUserCommand.php
@@ -11,10 +11,10 @@
 
 namespace AppBundle\Command;
 
+use AppBundle\Entity\User;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
-use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\Question;
 use Doctrine\Common\Persistence\ObjectManager;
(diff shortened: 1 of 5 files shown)


src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -109,6 +109,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $this->usernameValidator($username);
 
         $repository = $this->entityManager->getRepository('AppBundle:User');
+        /** @var User $user */
```
voronkovich: @yceruto, I think it's better to use a FQCN here.

yceruto: @voronkovich why it's better to use a FQCN here? please.

voronkovich: @yceruto, using FQCN doesn't require to import User class.

yceruto: Ok, fixed! thanks @voronkovich

src/AppBundle/Command/DeleteUserCommand.php
```diff
@@ -109,6 +108,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $this->usernameValidator($username);
 
         $repository = $this->entityManager->getRepository('AppBundle:User');
+        /** @var \AppBundle\Entity\User $user */
```
xabbuh: I would use a `use` statement for this class (that's what we do in the core).

voronkovich: @xabbuh, what the advantages of importing a class that we don't really use?

src/AppBundle/Command/DeleteUserCommand.php
yceruto: Changed by @xabbuh comment.

javiereguiluz:
Thanks @yceruto.

PR #239, opened 2015-10-26
add a custom form field type example

yceruto:
**The first purpose** of this PR is to provide a practical example of how to create a custom form field type.

**The second purpose** is to improve the user interface to the field "Published At" on edit a Post, 
using a JavaScript date library like bootstrap-datetimepicker for parsing, validating, manipulating, and formatting dates. http://eonasdan.github.io/bootstrap-datetimepicker.

>**Important:** Is not the purpose of this PR create a custom form field type to support all options and features of the datetimepicker plugin.

References:
* http://symfony.com/doc/current/cookbook/create_custom_field_type.html
* http://symfony.com/doc/current/best_practices/forms.html#custom-form-field-types

Main Topics:
* Defining the custom form Field Type.
* Creating your form Field Type as a Service.
* Creating a Template for the custom form field Type
* Configuring the custom form template.
* Initializing datetime picker JavaScript plugin.
* Using the form Field Type into PostType form.

Others Topics (refactor):
* Moving the initialization of the highligth JavaScript plugin for `main.js` file.

**Preview:**
datetimepicker


yceruto:
I need help to rebase this PR.

javiereguiluz:
This is really nice! Tell me, how can we help you? Thanks.

yceruto:
@javiereguiluz I mistakenly worked on this PR from a previous PR. Would like to leave just two last commit in this branch. Can you help me?

javiereguiluz:
In that case I always do a `git rebase -i`, delete the wrong commits and then I do a `git push -f`

voronkovich:
@yceruto, you can use `git cherry-pick` to copy the last commits to an appropriate branch and then `git rebase -i` to delete those commits from the current branch. https://nathanhoad.net/how-to-cherry-pick-changes-with-git

voronkovich:
@yceruto, I think it's not necessary to define `MomentFormatConverter` as a service,  you can create it's instance inside the form's type constructor method.

yceruto:
@javiereguiluz @voronkovich I've rebased successfully. Thanks.

app/Resources/views/Form/fields.html.twig
```diff
@@ -0,0 +1,17 @@
+{#
+   Each field type is rendered by a template fragment, which is determined
+   by the value of your getName() method and the suffix _widget.
+
+   See http://symfony.com/doc/current/cookbook/form/create_custom_field_type.html#creating-a-template-for-the-field
+#}
+
+{% block app_datetimepicker_widget %}
+    {% spaceless %}
+        <div class='input-group date' id="datetimepicker">
```
voronkovich: @yceruto, I think you should use a css class here (datetimepicker) instead of the id.

src/AppBundle/Form/Type/DateTimePickerType.php
```diff
@@ -0,0 +1,77 @@
...
+        $attr = isset($view->vars['attr']) ? $view->vars['attr'] : array();
+
+        $view->vars['attr'] = array_merge($attr, array(
+            'data-date-format' => $this->formatConverter->convert($options['format']),
+            'data-date-locale' => \Locale::getDefault()
+        ));
+    }
+
+    /**
+     * @inheritdoc
```
voronkovich: A wrong doctype tag

app/Resources/views/Form/fields.html.twig
yceruto: @voronkovich It's always good to explain why your suggestion to better understand the reason for the idea. Please, explain a bit more.

bocharsky-bw: @yceruto With css class you could reuse it for more than one field. With ID it's impossible if we want to keepvalid HTML structure.

voronkovich: @yceruto, you can have multiple datetime fields in a page.

yceruto: @voronkovich @bocharsky-bw are right, the better. Fixed using `data-toggle` attribute to avoid CSS bug.

src/AppBundle/Form/Type/DateTimePickerType.php
yceruto: @voronkovich Fixed, thanks.

src/AppBundle/Form/Type/DateTimePickerType.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        $this->formatConverter = new MomentFormatConverter();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function buildView(FormView $view, FormInterface $form, array $options)
+    {
+        $attr = isset($view->vars['attr']) ? $view->vars['attr'] : array();
```
voronkovich: This checking isn't needed, because the value of `$view->vars ['attr']` is already assigned to an array. See
https://github.com/symfony/form/blob/master/Extension/Core/Type/BaseType.php#L93
https://github.com/symfony/form/blob/master/Extension/Core/Type/BaseType.php#L117

yceruto: @voronkovich, that's right, fixed! thanks.

src/AppBundle/Form/Type/DateTimePickerType.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        $this->formatConverter = new MomentFormatConverter();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function buildView(FormView $view, FormInterface $form, array $options)
+    {
+        $view->vars['attr'] = array_merge($view->vars['attr'], array(
```
voronkovich: @yceruto, why you use the `array_merge` function?

yceruto: @voronkovich to avoid restricting the possibility of passing some attribute to this type of field. If I do not 'array_merge` overwrite the custom attributes that can be passed of the form:
```
$builder->add('publishedAt', 'app_datetimepicker', array(
    'label' => 'label.published_at',
    'attr' => //this attributes will be merge with array('data-date-format' => '?', 'data-date-locale' => '?')
));
```

voronkovich: @yceruto, Passing those attributes can break the form, isn't it? I think you should use options 'format' and 'locale' for this.

yceruto: @voronkovich I do not understand your comment. How can break the form? I already use the `format` option and the locale option is no necessary, use `\Locale::getDefault()` to find out the current locale.

voronkovich: @yceruto, what if user pass an attribute `data-date-format` to the field type? In what date format data will be submitted (`format` or `data-date-format`)?
I think that your code should look like this:
```php
$this-view['attr']['data-date-format'] = $this->formatConverter->convert($options['format']);
```

yceruto: @voronkovich I think you're missing something, if the attributes `data-date-format` or `data-date-locale` is passed will be overwritten due to the following `array_merge(['a' => 1], ['a' => 2]) == ['a' => 2]`. The way it is implemented now this is accomplished and always `format` option will be submit. I think that is not necessary do what you propose.

voronkovich: @yceruto ok, my bad. But you can do it without using `array_merge`:

```php
public function buildView(FormView $view, FormInterface $form, array $options)
{
    $view->vars['attr']['data-date-format'] = $this->formatConverter->convert($options['format']);
    $view->vars['attr']['data-date-locale'] = \Locale::getDefault():
}
```

yceruto: @voronkovich  you're right, it looks better that way. Thanks.

app/Resources/views/form/fields.html.twig
```diff
@@ -0,0 +1,17 @@
+{#
+   Each field type is rendered by a template fragment, which is determined
+   by the value of your getName() method and the suffix _widget.
+
+   See http://symfony.com/doc/current/cookbook/form/create_custom_field_type.html#creating-a-template-for-the-field
+#}
+
+{% block app_datetimepicker_widget %}
+    {% spaceless %}
+        <div class='input-group date' data-toggle="datetimepicker">
```
voronkovich: @yceruto, chage single quotes to double quotes here, please.

yceruto: Thanks! :)

voronkovich:
:+1:

javiereguiluz:
I'd like to merge this super-cool date picker as soon as possible. However, I have a problem with the feature that allows you to navigate from one month to another. The feature works right ... but the arrows aren't displayed:

datepicker

yceruto:
@javiereguiluz fixed, thanks!

javiereguiluz:
@yceruto and it's working like a charm! Thanks.

datepicker

javiereguiluz:
@yceruto thanks a lot for implementing this feature. I absolutely love it! I also enjoyed your discussion with @voronkovich during the development of the feature. You both should consider doing some "pair programmin" sometime because you are a great team :)

yceruto:
:+1:

PR #240, opened 2015-10-27
Update dependencies

bocharsky-bw:
According to the new Symfony's [v2.7.6](https://github.com/symfony/symfony/pull/16357) release.


javiereguiluz:
@bocharsky-bw thanks for updating to the new version so fast!

PR #241, opened 2015-10-30
feature #240 Update Dependencies

tonylds1:
[web] [config.php] this file is updated by composer.json script

### DIFF
--- a/web/config.php
+++ b/web/config.php
@@ -1,10 +1,13 @@
 <?php
 
 if (!isset($_SERVER['HTTP_HOST'])) {
-    exit('This script cannot be run from the console. Run it from a browser.');
+    exit('This script cannot be run from the CLI. Run it from a browser.');
 }
 
-if (!in_array(@$_SERVER['REMOTE_ADDR'], array('127.0.0.1', '::1'))) {
+if (!in_array(@$_SERVER['REMOTE_ADDR'], array(
+    '127.0.0.1',
+    '::1',
+))) {
     header('HTTP/1.0 403 Forbidden');
     exit('This script is only accessible from localhost.');
 }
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
:-1: This script was customized for Symfony Demo app

javiereguiluz:
@tonylds1 you are right, but this project uses a customized script because the Symfony script is a bit ugly and ours is better integrated with the overall design of the application.

By the way, anyone knows how could we prevent further issues related to this script? How can ignore the default Symfony script forever?

voronkovich:
@javiereguiluz,  we could add it to .gitignore

javiereguiluz:
@voronkovich yes, but the main problem is that when you execute `composer update`, our script is overridden by the default script.

voronkovich:
@javiereguiluz, I think the main problem is that every month somebody creates a PR with the updated `config.php`. :)

javiereguiluz:
A solution could be to redesign the original config.php so it sucks less. After all, we've already redesigned most of the Symfony elements.

voronkovich:
@javiereguiluz, I think we could add a command `git checkout web/config.php` to the `post-update-cmd` section of the composer.json

voronkovich:
@javiereguiluz what about to rename `config.php`?
See https://github.com/sensiolabs/SensioDistributionBundle/blob/master/Composer/ScriptHandler.php#L216

javiereguiluz:
@voronkovich we did that in the past ... and @bocharsky-bw still has nightmares about that ;)

bocharsky-bw:
Yeah, `app/check.php` script refers exactly to the `config.php`. It was in https://github.com/symfony/symfony-demo/pull/140 :)

voronkovich:
I think we could use a dirty hack, like this:
```json
"post-update-cmd": [
            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
            "mv web/config.php web/_config.php",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
            "mv web/_config.php web/config.php",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::removeSymfonyStandardFiles"
        ]
```

javiereguiluz:
Let's close this issue as "won't fix" for now and we'll revisit it when Symfony 2.8 and 3.0 are released. Thanks!

PR #247, opened 2015-11-17
Fix invalid YAML format

### LINKED ISSUE #246: Prepare the YAML configuration files for Symfony 3
I'm trying to make this demo app work with Symfony 3. Right now it's impossible because KnpPaginatorBundle is not compatible with 3. However, I managed to make it work somehow and I'm seeing these deprecation errors:

```
The file "app/config/services.yml" does not contain valid YAML.

The reserved indicator "@" cannot start a plain scalar; you need to
quote the scalar at line 16 (near "arguments: [@markdown, %app_locales%]").
```

In `services.yml` we need to wrap with quotes any string that starts with `@` or `%`. If you want to submit a pull request to solve this issue, please use single quotes instead of double quotes. Thanks!

bocharsky-bw:
Resolve #246

### DIFF
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -5,7 +5,7 @@
 # the routes in YAML, XML or PHP files.
 # See http://symfony.com/doc/current/book/routing.html
 app:
-    resource: @AppBundle/Controller/
+    resource: '@AppBundle/Controller/'
     type:     annotation
     prefix:   /{_locale}
     requirements:
@@ -23,5 +23,5 @@ homepage:
         _locale: %app_locales%
     defaults:
         _controller: FrameworkBundle:Template:template
-        template:    'default/homepage.html.twig'
-        _locale:     "%locale%"
+        template:    default/homepage.html.twig
+        _locale:     %locale%
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -12,7 +12,7 @@ services:
     app.twig.app_extension:
         public:    false
         class:     AppBundle\Twig\AppExtension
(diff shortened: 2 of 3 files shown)


app/config/routing.yml
```diff
@@ -9,9 +9,9 @@ app:
     type:     annotation
     prefix:   /{_locale}
     requirements:
-        _locale: %app_locales%
+        _locale: '%app_locales%'
```
javiereguiluz: I thought '%' also needed escaping :confused:

bocharsky-bw: What about to keep quotes for parameters (wrapped with `%`) anyway?

bocharsky-bw: Hm, seems Symfony plugin's autocomplete work incorrectly when `%` in quotes (it's doubled `%`), so I reverted changes for parameters, thanks @xabbuh !

xabbuh: Well, it should work the same whether or not you use quotes. So I suggest to report this to the Symfony plugin if it isn't already a known issue.

bocharsky-bw: OK, I created an [issue](https://github.com/Haehnchen/idea-php-symfony2-plugin/issues/624)

stof: ``%`` is not special in Yaml 2, so quoting is not necessary. The symfony-docs quotes it all the time because the Pygments highlighter follows the rule of Yaml 1.1 where quoting was needed

javiereguiluz:
Thanks @bocharsky-bw.

PR #249, opened 2015-11-18
Use a custom variable for getting the current Twig template instance

voronkovich:
This PR fixes https://github.com/symfony/symfony-demo/issues/242

### DIFF
--- a/app/Resources/TwigBundle/views/Exception/error.html.twig
+++ b/app/Resources/TwigBundle/views/Exception/error.html.twig
@@ -28,5 +28,5 @@
 {% block sidebar %}
     {{ parent() }}
 
-    {{ show_source_code(_self) }}
+    {{ show_source_code(_tmpl) }}
 {% endblock %}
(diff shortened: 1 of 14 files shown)


javiereguiluz:
The proposed solution is great ... although maybe too complex? I was thinking about solving this issue as:

```twig
{% import _self as current_template %}
{{ show_source_code(current_template) }}
```

But I guess @voronkovich's solution is more professional :)

voronkovich:
@javiereguiluz, yeah, it's an enterprise level solution! :)

javiereguiluz:
I'm reopening the pull request because I think it might be a good solution. It's complex ... but that doesn't mean it's bad. Maybe we could use it to show advanced Twig features. Let's wait to hear more opinions about this. Thanks.

stof:
@javiereguiluz the issue of your proposal is that it abuses the import tag (Twig 2 might change the way macros are imported, forbidding this again).
This PR matches my initial idea, except that I wanted to avoid giving such power to anyone (the current code duplicates the deprecated ``_self`` behavior for anyone, but with a different variable name)
I then thought about another way, which would solve other needs for which the removal of ``_self`` is painful, and I opened the discussion on twig (which is why I haven't opened a PR on the demo yet): https://github.com/twigphp/Twig/issues/1932

javiereguiluz:
Closing it in favor of #254 ( although I don't know if that would be the definitive solution either).

PR #252, opened 2015-11-18
Include asset Twig function for Symfony 3

bocharsky-bw:
Fix next error after update to the Symfony 3:

> Unknown "asset" function in twig template

P.S. following [v3.0.0-BETA1](https://github.com/symfony/symfony-standard/blob/51ca575c8f1339e62edb077f4f5e0ff1f5566423/app/config/config.yml#L34) release


javiereguiluz:
Thanks @bocharsky-bw.

PR #253, opened 2015-11-18
Fix #250: Replace form type object with classname string

### LINKED ISSUE #250: Fix the Form component deprecations for Symfony 3
If you try to create a new blog post (`/en/admin/post/new`):

admin_post_create

If you try to edit a blog post (`/en/admin/post/1/edit`):

admin_post_edit

bocharsky-bw:


javiereguiluz:
@bocharsky-bw thanks for this fix ... but I'm afraid I have bad news for you :'(  This project is compatible with `php >=5.3.9` so we can't use the `::class` trick and we must write down the entire fully qualified class name of the form types.

bocharsky-bw:
For now, but when we update to the Symfony 3 we should use `5.5.9` at least, right?

javiereguiluz:
Yes.

bocharsky-bw:
So keep it as is or update with fully qualified classname?

javiereguiluz:
Good question. I'd prefer to use the full string because that works perfectly on 2.8 and 3.0 ... but in the future we'll need to change it again to make code more concise.

voronkovich:
@bocharsky-bw, you forgot to update form type's classes too. See https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Form/PostType.php

bocharsky-bw:
@javiereguiluz You're right. I updated with fully qualified classname for now.

bocharsky-bw:
@voronkovich Did you mean this https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Form/PostType.php#L64 ? Seems for now we should use fully qualified classnames, but thanks you pointed it!

voronkovich:
@bocharsky-bw, ok.

stof:
You should also switch to FQCN rather than legacy type names inside the existing form types to avoid deprecation warnings there

stof:
@javiereguiluz this requires bumping the Symfony requirement to ``~2.8`` though. Currently, the code still says ``~2.7``

bocharsky-bw:
Switched. Thanks @stof !

app/config/services.yml
```diff
@@ -31,7 +31,7 @@ services:
     app.form.type.datetimepicker:
```
voronkovich: @bocharsky-bw now we needn't this service. You can remove it.

bocharsky-bw: But why? It still use in https://github.com/bocharsky-bw/symfony-demo/blob/form-type/src/AppBundle/Form/PostType.php#L52-L54

voronkovich: @bocharsky-bw, because in the PostType we use FQCN.

bocharsky-bw: Hm, it's true. Thanks!

yceruto:
This comment is not related to this PR. But IMHO these changes are a throwback for developers. :(

src/AppBundle/Form/PostType.php
```diff
@@ -43,13 +43,13 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'attr' => array('autofocus' => true),
                 'label' => 'label.title',
             ))
-            ->add('summary', 'textarea', array('label' => 'label.summary'))
-            ->add('content', 'textarea', array(
+            ->add('summary', 'Symfony\Component\Form\Extension\Core\Type\TextareaType', array('label' => 'label.summary'))
```
yceruto: This `summary` field is a `string` type, why use textarea? or perhaps we should redefine it as `text`?

src/AppBundle/Form/PostType.php
```diff
@@ -43,13 +43,13 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+            ->add('content', 'Symfony\Component\Form\Extension\Core\Type\TextareaType', array(
```
yceruto: I prefer to leave the guesser to determine the type before using the FQCN.

yceruto: Now be explicit can be a problem. :(

src/AppBundle/Form/PostType.php
bocharsky-bw: I think it doesn't necessary to change it to `text`... Even with `255` length more convenient work with `textarea`.

src/AppBundle/Form/PostType.php
bocharsky-bw: I guess we can change the type to `null` where it appropriate. What do you think, @javiereguiluz ?

src/AppBundle/Form/PostType.php
yceruto: @bocharsky-bw then, when users to type a text with 256 characters or more will be truncated and this could be an unexpected result. I suggest then use `max_length` 255 for this widget.

voronkovich: `max_length' is deprecated option

src/AppBundle/Form/PostType.php
voronkovich: @bocharsky-bw, If you want to use the type guessing you should change the type of summary field to a `text`, as @yceruto suggests

src/AppBundle/Form/PostType.php
yceruto: @voronkovich I meant to the html attribute:
```
'attr' => array('maxlength' => 255)
```

voronkovich: @yceruto, ok.

src/AppBundle/Form/PostType.php
yceruto: that is, change it to `text`, but adding `maxlength` attribute.

bocharsky-bw: Or we can keep FQCN for `summary` but change type to `null` for other :)

voronkovich: @yceruto, I think it's better to add a validation rule for it, instead of using the `maxlength`

javiereguiluz:
@bocharsky-bw thanks for working on this. I'm merging it because I want to release a Symfony 2.8-based demo app soon. Thanks!

PR #254, opened 2015-11-18
Show a template's source code without relying to the "_self" variable

voronkovich:
Not so professional as https://github.com/symfony/symfony-demo/pull/249 :)

### DIFF
--- a/src/CodeExplorerBundle/Twig/SourceCodeExtension.php
+++ b/src/CodeExplorerBundle/Twig/SourceCodeExtension.php
@@ -41,13 +41,16 @@ class SourceCodeExtension extends \Twig_Extension
     public function getFunctions()
     {
         return array(
-            new \Twig_SimpleFunction('show_source_code', array($this, 'showSourceCode'), array('is_safe' => array('html'), 'needs_environment' => true)),
+            new \Twig_SimpleFunction('show_source_code', array($this, 'showSourceCode'), array(
+                'is_safe' => array('html'),
+                'node_class' => 'CodeExplorerBundle\Twig\Node\TemplateAwareFunction'
+            )),
         );
     }
 
-    public function showSourceCode(\Twig_Environment $twig, \Twig_Template $template)
+    public function showSourceCode(\Twig_Template $template)
     {
-        return $twig->render('@CodeExplorer/source_code.html.twig', array(
+        return $template->getEnvironment()->render('@CodeExplorer/source_code.html.twig', array(
             'controller' => $this->getController(),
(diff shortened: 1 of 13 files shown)


javiereguiluz:
@voronkovich this is a pretty amazing feature. I'd like to ask others (e.g. @stof) if you agree with this proposed solution. Thanks!

javiereguiluz:
In https://github.com/symfony/symfony-demo/pull/249 there is an alternative proposal to solve this issue.

src/CodeExplorerBundle/Twig/SourceCodeExtension.php
```diff
@@ -41,13 +41,16 @@ public function setController($controller)
...
+                'node_class' => 'CodeExplorerBundle\Twig\Node\TemplateAwareFunction'
+            )),
         );
     }
 
-    public function showSourceCode(\Twig_Environment $twig, \Twig_Template $template)
+    public function showSourceCode(\Twig_Template $template)
     {
-        return $twig->render('@CodeExplorer/source_code.html.twig', array(
+        return $template->getEnvironment()->render('@CodeExplorer/source_code.html.twig', array(
```
stof: no good, as ``\Twig_Template->getEnvironment()`` is also deprecated

voronkovich: @stof, using the `needs_environment` option isn't possible here, because it requires more complicated code. What about to inject the `Twig_Environment` instance to the extention?

stof: Injecting it is not possible, because the extension itself is injected in the environment. If you want to use a custom node for the function, it should keep supporting the existing features.

voronkovich: @stof, I can pass the `Twig_Environment` as a second argument:
```php
$compiler->raw('($this, $this->env)');
```
Supporting the existing features would require to copy all the code from the `compileArguments` method https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Node/Expression/Call.php#L37. Is it really necessary?

voronkovich: @stof, I've fixed this, now it supports all existing features like `needs_environment` and etc.

javiereguiluz:
I'm closing this because `_self` has been un-deprecated. In any case, thank you @voronkovich for providing different solutions to this issue and for showing your absolute mastery of Twig internals.

PR #257, opened 2015-11-21
Update config.yml

avinsol:
adding catalan language


xabbuh:
Imho this should be done in #258 (there isn't any Catalan translation yet).

javiereguiluz:
Thanks @avinsol.

PR #260, opened 2015-11-25
Fixed demo with PostgreSQL database

lyrixx:
* 'user' table is forbiden
* the model says the column is string(255), but the generated content is too large. (It does not work because postgres is a real database engine)

### DIFF
--- a/src/AppBundle/DataFixtures/ORM/LoadFixtures.php
+++ b/src/AppBundle/DataFixtures/ORM/LoadFixtures.php
@@ -179,7 +179,7 @@ MARKDOWN;
         $numPhrases = rand(6, 12);
         shuffle($phrases);
 
-        return implode(' ', array_slice($phrases, 0, $numPhrases-1));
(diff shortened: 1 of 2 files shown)


src/AppBundle/DataFixtures/ORM/LoadFixtures.php
```diff
@@ -179,7 +179,7 @@ private function getRandomPostSummary()
         $numPhrases = rand(6, 12);
         shuffle($phrases);
 
-        return implode(' ', array_slice($phrases, 0, $numPhrases-1));
+        return substr(implode(' ', array_slice($phrases, 0, $numPhrases-1)), 0, 254);
```
bocharsky-bw: I think we can use from `0` to `255` here

lyrixx: indeed, fixed.

javiereguiluz:
Closing it in favor of #283, which reuses the original commits so @lyrixx will get full credit for his original work. Thanks for letting us know about these errors and for the provided solution.

PR #264, opened 2015-12-01
Update Symfony up to 2.8.0 with dependencies

bocharsky-bw:
Update to `2.8` before `3.0`.


bocharsky-bw:
Should we also update directory structure to the Symfony `3.0` in this PR?

javiereguiluz:
@bocharsky-bw let's do that in a separate PR. In any case, we don't have to update to 3.0 in the coming days. We can stay at 2.8 for some time. Thanks.

stof:
updating to 2.8 first would indeed be better, to give time to solve all deprecations.

but you should update the app to use the new way to load the autoloader, allowing to update SensioDistributionBundle to 5.x

xabbuh:
FYI, the steps @stof is talking about are described [in the upgrade file](https://github.com/symfony/symfony-standard/blob/2.8/UPGRADE-2.8.md) of the Symfony Standard Edition.

bocharsky-bw:
Made changes according to [UPGRADE-2.8.md](https://github.com/symfony/symfony-standard/blob/2.8/UPGRADE-2.8.md), thanks @stof and @xabbuh for help!

composer.json
```diff
@@ -4,7 +4,8 @@
     "type": "project",
     "description": "Symfony Demo Application",
     "autoload": {
-        "psr-0": { "": "src/", "SymfonyStandard": "app/" }
+        "psr-4": { "": "src/" },
+        "files": [ "app/AppKernel.php" ]
```
xabbuh: You should use `classmap` here (see symfony/symfony-standard#897).

bocharsky-bw: Changed.

Thanks!

javiereguiluz:
This looks good, so I'm merging it. Thanks @bocharsky-bw.

PR #265, opened 2015-12-01
Add missing ru translation

bocharsky-bw:
Add missing `label.save_and_create_new` translation in `messages.ru.xliff`


javiereguiluz:
@bocharsky-bw and another merged fix from you! Thanks.

PR #267, opened 2015-12-01
Fix invalid translation key for label name

bocharsky-bw:
We don't need to translate form labels, it happens automatically - otherwise we get a "Missing messages" notice in debug toolbar.


stof:
:+1:

javiereguiluz:
@bocharsky-bw very nice catch! Thanks for this fix.

PR #270, opened 2015-12-05
Added Polish translations

jzawadzki:
Polish translations for `previous`, `next`, and `save and create new`.


javiereguiluz:
Thank you @jzawadzki.

PR #273, opened 2015-12-08
Fix broken URL in Custom Field Type example

alfonsomga:


javiereguiluz:
@alfonsomga very nice catch! Thanks for fixing it.

PR #276, opened 2015-12-15
Remove locale duplicates in array of locales

bocharsky-bw:
We have locale dublicates after default locale unshifted to the array of locales on [line 70](https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php#L70).


voronkovich:
:+1:

javiereguiluz:
@bocharsky-bw thanks for this fix. I'm not fully convinced of this change because it looks like a hack ... but we definitely need your fix to avoid further issues. Thanks!

PR #277, opened 2015-12-15
Ignore requests from referrers with the same HTTP host

bocharsky-bw:
If user select `en` language and then go to the homepage - language changing to the preferred language base on user's browser suggestion (see `RedirectToPreferredLocaleListener`). But users don't need this because they already selected `en` for this application.

With this PR I suggest to prevent changing language to the preferred language for referrer requests with the same HTTP host. Thus locale changing will be occured only for `homepage`'s direct requests (`HTTP_REFERER` equal to `null`) or requests from referrers with different HTTP hosts in HTTP_REFERER URL.

### DIFF
--- a/src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
+++ b/src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
@@ -81,6 +81,11 @@ class RedirectToPreferredLocaleListener
         if (!$event->isMasterRequest() || '/' !== $request->getPathInfo()) {
             return;
         }
+        // Ignore requests from referrers with the same HTTP host in order to prevent
+        // changing language for users who possibly already selected it for this application.
+        if (0 === stripos($request->headers->get('referer'), $request->getSchemeAndHttpHost())) {
+            return;
+        }
 
         $preferredLanguage = $request->getPreferredLanguage($this->locales);
 


src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
```diff
@@ -81,6 +81,10 @@ public function onKernelRequest(GetResponseEvent $event)
         if (!$event->isMasterRequest() || '/' !== $request->getPathInfo()) {
             return;
         }
+        // Ignore requests from referrers with the same HTTP host
```
voronkovich: @bocharsky-bw, I think it's a bad comment, it doesn't explain why we should ignore a request. I would write a comment like this "We must handle only direct requests, because ...."

bocharsky-bw: @voronkovich let's describe it a bit. What about:

> Ignore requests from referrers with the same HTTP host in order to prevent changing language for users who already selected it in this application.

How does it sound for you?

voronkovich: @bocharsky-bw, it sounds good!

bocharsky-bw: Thanks! I just changed it. I agree to change it again if we find a more suitable description.

voronkovich: @bocharsky-bw, ok, let's wait a little bit,  maybe someone will suggest more appropriate description

src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
```diff
@@ -81,6 +81,11 @@ public function onKernelRequest(GetResponseEvent $event)
         if (!$event->isMasterRequest() || '/' !== $request->getPathInfo()) {
             return;
         }
+        // Ignore requests from referrers with the same HTTP host in order to prevent
+        // changing language for users who possibly already selected it for this application.
+        if (0 === stripos($request->server->get('HTTP_REFERER'), $request->getSchemeAndHttpHost())) {
```
voronkovich: @bocharsky-bw, it's better to use `$request->headers->get('REFERER')`. The name `server` is related to the super global php variable `$_SERVER`, not to the HTTP protocol.

bocharsky-bw: It's the same things I think. They both are equal and I also can directly modify both too. But there're no any security issues for this PR, only usage convenience. I prefer to keep SERVER's `HTTP_REFERER` because it well documented in [PHP docs](http://php.net/manual/en/reserved.variables.server.php) (to understand what is the `$request->headers->get('REFERER')` here - more difficult for newcomers). What do you think?

voronkovich: @bocharsky-bw, if the `$_SERVER` is well documented, why didn't you use it? Looks like old good PHP :)
```php
if (0 === stripos($_SERVER['HTTP_REFERER'], $request->getSchemeAndHttpHost()))
```

bocharsky-bw: We should avoid to use `$_SERVER` variable directly in Symfony projects but use `Request` object instead. Moreover, we already have injected Request object in this point.

voronkovich: @bocharsky-bw, it was a joke. I just meant that using `$request->headers->get('REFERER')` is more appropriate for me.

bocharsky-bw: Well, I'm fine to use `$request->headers->get('REFERER')` if others will vote for it too. Let's wait. Thanks!

yceruto: :+1: `$request->headers->get('referer')` for three reasons:
* `$request->headers` taken from the `$_SERVER`
* `HTTP_REFERER` is a request-header field: http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
* Preferred by Symfony core: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/EventListener/RouterListener.php#L119, https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Http/Authentication/DefaultAuthenticationSuccessHandler.php#L122

yceruto: Style Code Preference: As symfony-demo is part of the Symfony project I always try to think first as they do before thinking about how I would do it.

bocharsky-bw: @yceruto Good detective work

javiereguiluz:
@bocharsky-bw thanks for this nice little feature ... and thanks to the reviewers for the "detective work", as you put it :)

PR #278, opened 2015-12-15
Translate customized error pages

bocharsky-bw:
We have a translations in this demo app, but error pages do not translate.

For example, when I get some non-existent blog post (`/uk/blog/posts/non-existent-post`) in another lang (except `en`) I should see translated text.

### DIFF
--- a/app/Resources/TwigBundle/views/Exception/error404.html.twig
+++ b/app/Resources/TwigBundle/views/Exception/error404.html.twig
@@ -13,14 +13,13 @@
 {% block body_id 'error' %}
 
 {% block main %}
-    <h1 class="text-danger">Error 404</h1>
+    <h1 class="text-danger">{{ 'http_error.name'|trans({ '%status_code%': 404 }) }}</h1>
 
     <p class="lead">
-        We couldn't find the page you requested.
+        {{ 'http_error_404.description'|trans }}
     </p>
     <p>
-        Check out any misspelling in the URL or
-        <a href="{{ path('blog_index') }}">go back to the homepage</a>.
+        {{ 'http_error_404.suggestion'|trans({ '%url%': path('blog_index') })|raw }}
     </p>
 {% endblock %}
 
--- a/app/config/routing_dev.yml
+++ b/app/config/routing_dev.yml
@@ -6,5 +6,9 @@ _profiler:
     resource: "@WebProfilerBundle/Resources/config/routing/profiler.xml"
     prefix:   /_profiler
 
+_errors:
+    resource: "@TwigBundle/Resources/config/routing/errors.xml"
+    prefix:   /_error
+
 _main:
     resource: routing.yml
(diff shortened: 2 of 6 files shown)


app/Resources/TwigBundle/views/Exception/error404.html.twig
```diff
@@ -13,14 +13,14 @@
 {% block body_id 'error' %}
 
 {% block main %}
-    <h1 class="text-danger">Error 404</h1>
```
javiereguiluz: This message should be translated with a placeholder for the error code. In some languages we say `Error XXX` but in others they say `XXX Error`.

bocharsky-bw: Done, thanks!

javiereguiluz:
I'm about to merge your nice improvement, but I have doubts with this kind of translations:

```twig
{{ 'http_error.suggestion'|trans }}
<a href="{{ path('blog_index') }}">{{ 'http_error.go_back'|trans }}</a>.
```

```xml
<trans-unit id="http_error.suggestion">
    <source>http_error.suggestion</source>
    <target>Try loading this page again in some minutes or</target>
</trans-unit>
<trans-unit id="http_error.go_back">
    <source>http_error.go_back</source>
    <target>go back to the homepage</target>
</trans-unit>
```

Shouldn't we translate it like this?

```twig
{{ 'http_error.suggestion'|trans({ '%url%': path('blog_index') }) }}
```

```xml
<trans-unit id="http_error.suggestion">
    <source>http_error.suggestion</source>
    <target>Try loading this page again in some minutes or <a href="%url%">go back to homepage</a>.</target>
</trans-unit>
```

Please don't make any change yet. I'm just asking because I don't know which is the common practice in this kind of translations. Thanks!

bocharsky-bw:
I don't like that `http_error.suggestion` message ends with `... or` so I'm :+1: with your suggestion.

javiereguiluz:
@bocharsky-bw nobody said anything contrary to the previous suggestion, so when you have some free time, please update the error message translations to include the URL placeholder. Thanks!

app/config/routing_dev.yml
```diff
@@ -6,5 +6,9 @@ _profiler:
     resource: "@WebProfilerBundle/Resources/config/routing/profiler.xml"
     prefix:   /_profiler
 
+_errors:
```
bocharsky-bw: @javiereguiluz What about to add an additional route in order to debug HTTP errors like in [symfony-standard](https://github.com/symfony/symfony-standard/blob/master/app/config/routing_dev.yml#L9-L11)?

javiereguiluz: I think this is a very good idea! Thanks

javiereguiluz:
@bocharsky-bw thanks for this nice feature. The demo app was only "half translated" and thanks to you is now fully translated.

PR #271, opened 2015-12-07
Enable ESI caching for blog index

sergiu-popa:
We could update the demo app to include by default ESI caching for the blog index and the sidebar.

PR #279, opened 2015-12-16
Tweaked the ESI integration

javiereguiluz:
Minor tweaks following the merge of #271

### DIFF
--- a/app/Resources/views/base.html.twig
+++ b/app/Resources/views/base.html.twig
@@ -152,6 +152,6 @@
             <script src="{{ asset('js/app.js') }}"></script>
         {% endblock %}
 
-        <!-- Page rendered on {{ "now"|date('j/M/Y H:i:s') }} -->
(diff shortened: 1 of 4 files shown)


app/Resources/views/base.html.twig
```diff
@@ -151,5 +151,7 @@
 
             <script src="{{ asset('js/app.js') }}"></script>
         {% endblock %}
+
+        <!-- Page rendered on {{ "now"|date('j/M/Y H:i:s') }} -->
```
bocharsky-bw: What about to use `localizeddate` filter here? Does it meaningless for public cached pages?

javiereguiluz: I don't care much because this is for developers instead of end-users. But considering that this improvement comes "for free" I've implemented your suggestion. Thanks!

stof:
:+1:

PR #280, opened 2015-12-16
Add title filter in admin using LexikFormFilterBundle

sergiu-popa:
Since the admin backend doesn't use KnpPaginator, we could add a filter for title field using LexikFormFilterBundle.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
         "symfony/monolog-bundle"               : "~2.7",
         "symfony/swiftmailer-bundle"           : "~2.3",
         "symfony/symfony"                      : "~2.8",
-        "twig/extensions"                      : "~1.2"
+        "twig/extensions"                      : "~1.2",
+        "lexik/form-filter-bundle": "dev-master"
     },
     "require-dev": {
         "sensio/generator-bundle": "~3.0"
--- a/src/AppBundle/Controller/Admin/BlogController.php
+++ b/src/AppBundle/Controller/Admin/BlogController.php
@@ -11,6 +11,7 @@
 
 namespace AppBundle\Controller\Admin;
 
+use AppBundle\Filter\PostFilterType;
 use AppBundle\Form\PostType;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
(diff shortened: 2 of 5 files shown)


javiereguiluz:
I'm not sure about this feature. The purpose of the Symfony Demo is to show as much built-in Symfony features as possible. We don't want to develop an advanced blog engine.

Before merging or closing this proposal, I'll wait to read more opinions. Thanks!

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -50,12 +51,29 @@ class BlogController extends Controller
      * @Route("/", name="admin_post_index")
      * @Method("GET")
      */
-    public function indexAction()
+    public function indexAction(Request $request)
     {
-        $entityManager = $this->getDoctrine()->getManager();
-        $posts = $entityManager->getRepository('AppBundle:Post')->findAll();
+        $form = $this->get('form.factory')->create(new PostFilterType());
```
stof: don't pass a type instance. This is deprecated

src/AppBundle/Filter/PostFilterType.php
```diff
@@ -0,0 +1,29 @@
...
+    public function getName()
+    {
+        return 'post_filter';
+    }
+
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults(array(
+            'csrf_protection'   => false,
+            'validation_groups' => array('filtering') // avoid NotBlank() constraint-related message
```
stof: which not-blank ? there are none

sergiu-popa: Copied from bundle's installation instructions.

src/AppBundle/Controller/Admin/BlogController.php
sergiu-popa: How it is done now?

bocharsky-bw: @sergiu-popa use FQCN instead: `...->create('AppBundle\Filter\PostFilterType')`.

bocharsky-bw:
I think @javiereguiluz is right. Third-party bundles mostly are developer specific, some developers use this bundle, some of them could use other.

composer.json
```diff
@@ -26,7 +26,8 @@
         "symfony/monolog-bundle"               : "~2.7",
         "symfony/swiftmailer-bundle"           : "~2.3",
         "symfony/symfony"                      : "~2.8",
-        "twig/extensions"                      : "~1.2"
+        "twig/extensions"                      : "~1.2",
+        "lexik/form-filter-bundle": "dev-master"
```
stof: depending on dev-master is a no-go. If the bundle does not have stable releases (which should be considered though), a branch alias should at least be used to avoid getting bumped to a new major version if they start developing it. If they don't support installing it the bundle other than with a ``dev-master`` requirement, I vote against including it in the demo.
And if they don't have a branch alias, you should open a PR adding it.

stof:
And btw, this bundle is used only to perform a ``->andWhere('p.title LIKE :query)->setParameter('query', '%'.$query.'%')`` on the QueryBuilder (well, better if we perform some escaping on ``$query`` in case there is a ``%`` in it). The bundle is way overkill IMO

composer.json
sergiu-popa: composer.phar require lexik/form-filter-bundle ~4.0

sergiu-popa:
Provided types: https://github.com/lexik/LexikFormFilterBundle/blob/master/Resources/doc/provided-types.md. I could add this functionality without the Lexik bundle if we decide it should be in the Demo App.

stof:
@sergiu-popa and all these custom types are necessary only because of all the complex logic the bundle contains just to perform this alteration of the query builder. If you just build it the simple way, using the ``text`` type for the field is enough

sergiu-popa:
@javiereguiluz, I don't consider by adding a filter would make the demo an advanced blog engine. I think we should add a filter for a table because it's not something unusual for an application. If not using LexikBundle, I could create it using Symfony's built in functionality. What do you think?

javiereguiluz:
I'm closing this pull request for being too complex for the purpose of the Symfony Demo app.

However, @sergiu-popa is right and we could add some filter in the backend without using this filter. My question: could we adapt the (still not merged) https://github.com/symfony/symfony-demo/pull/173 to use it also in the backend?

Issue #281, opened 2015-12-18
Getting HTMLPurifier warnings


lfjeff:
I am getting these warnings in the error log:

```
[Thu Dec 17 20:00:37 2015] [error] [client 24.251.53.52] PHP Warning:  Directory 
/home/mydir/symfony_demo/vendor/ezyang/htmlpurifier/library/HTMLPurifier/DefinitionCache/Serializer 
not writable, please chmod to 775 in 
/home/mydir/symfony_demo/vendor/ezyang/htmlpurifier/library/HTMLPurifier/DefinitionCache/Serializer.php on line 278,
```

It appears that the demo is using the HTMLPurifier package instead of the HTMLPurifier Bundle.  If the bundle was used, it should be possible to configure things to set up the cache directory in a better location (not in vendor subdirectory).  See https://github.com/Exercise/HTMLPurifierBundle for details.

voronkovich:
I think we could just disable the cache.

javiereguiluz:
This has been fixed in #282 and it will be available when the next stable version is released.

stof:
@javiereguiluz here is what I have in my config at work to fix this issue (requires HtmlPurifier 4.7+ to work though):

```yaml
exercise_html_purifier:
    default:
        Cache.SerializerPermissions: 0775
```

The default value in HtmlPurifier is ``0755``, but ACLs are masked by the group permission, so the cache must be group-writable to be writable by both the CLI and Apache when using ACLs.

The demo uses the library directly rather than using the Exercise bundle, but this would not forbid to configure the cache permissions too. It would be a better fix than disabling the cache entirely.

PR #282, opened 2015-12-18
Disable HTMLPurifier caching

voronkovich:
Fix an issue #281

### DIFF
--- a/src/AppBundle/Utils/Markdown.php
+++ b/src/AppBundle/Utils/Markdown.php
@@ -34,7 +34,11 @@ class Markdown
     public function __construct()
     {
         $this->parser = new \Parsedown();
-        $this->purifier = new \HTMLPurifier();
+
+        $purifierConfig = \HTMLPurifier_Config::create(array(
+            'Cache.DefinitionImpl' => null, // Disable caching
+        ));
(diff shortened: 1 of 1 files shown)


src/AppBundle/Utils/Markdown.php
```diff
@@ -34,7 +34,11 @@ class Markdown
     public function __construct()
     {
         $this->parser = new \Parsedown();
-        $this->purifier = new \HTMLPurifier();
+
+        $purifierConfig = \HTMLPurifier_Config::create(array(
+            'Cache.DefinitionImpl' => null, // Disable caching
```
bocharsky-bw: What about to cache it to the `/app/cache/{ENV}/html_purifier` instead of disable?

voronkovich: @bocharsky-bw, it will require more code for creating those directories.
See https://github.com/Exercise/HTMLPurifierBundle/blob/master/CacheWarmer/SerializerCacheWarmer.php

javiereguiluz:
I think that in this case, disabling the HTMLPurifier cache is a good compromise to avoid the reported issues. Thanks @voronkovich

PR #283, opened 2015-12-21
Improve our database to allow using other engines (PostgreSQL)

javiereguiluz:
This PR finishes the work started by @lyrixx in #260.


lyrixx:
:+1:

src/AppBundle/Entity/Comment.php
```diff
@@ -16,6 +16,7 @@
 
 /**
  * @ORM\Entity
+ * @ORM\Table(name="symfony_demo_comment")
```
voronkovich: What about to use a table prefix for this? http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/cookbook/sql-table-prefixes.html
Or is it too advanced feature?

javiereguiluz: @voronkovich I didn't know about this. Thanks for sharing that link.

I don't know if others will agree, but I think that solution is too advanced even for experienced developers.

voronkovich:
@javiereguiluz,  you forgot to update the test database's schema

src/AppBundle/Entity/Comment.php
stof: I agree with @javiereguiluz

PR #286, opened 2015-12-21
Add posts featured images

voronkovich:
Each ugly CMS has this feature, I think we should have it too! :) Also I think we need an example of how to upload images and how to use the LiiImagineBundle.

blog_index
post_show


bocharsky-bw:
I'm not sure about this. Symfony Demo isn't considered as a CMS and images increase the complexity of this project.

voronkovich:
@20uf, @bocharsky-bw It's just a simple example showing how to upload files.
http://symfony.com/doc/current/controller/upload_file.html

javiereguiluz:
I was never fond of this feature. Although both @voronkovich and @bocharsky-bw made good arguments in favor/against it, I'm going to close it for now to not complicate the demo application too much. Of course we can reconsider this in the near future. Thanks!

PR #291, opened 2015-12-23
Add ability to test HTTP error pages with different locales

bocharsky-bw:
What about to add `_locale` parameter in HTTP error route prefix in order to test HTTP error pages with different locales?


voronkovich:
:+1:

javiereguiluz:
@bocharsky-bw I like this change and it perfectly matches your previous pull request to translate error pages. Thanks!

PR #294, opened 2016-01-13
Switch to the new Symfony 3 directory structure

bocharsky-bw:


stof:
I'm wondering whether we should do it while we are still on 2.8

xabbuh:
I am :-1: for doing this now (we can switch when upgrading to Symfony 3).

javiereguiluz:
I'm closing this because it looks "too early" to do it. First we need to think well how/when we'll update to Symfony 3. Thanks!

PR #297, opened 2016-01-19
Heroku Button

dzuelke:
- logs to `php://stderr` in prod (sane default IMO, also for other PaaS, Docker etc)
- uses `url` to configure database, which in general makes config simpler too (in prod, it uses `DATABASE_URL` from the env)
- DoctrineFixturesBundle is also needed in prod to populate database with fixtures
- `app.json` and "deploy to Heroku" button take care of the magic (go to https://github.com/dzuelke/symfony-demo/tree/herokubutton and scroll down to see it)

Inclusion on http://symfony.com/download is obviously the next step ;)

Changes reviewed by @stof in December.

Should I also add "manual" deploy instructions to `README` or is that out of scope?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Symfony Demo Application
 The "Symfony Demo Application" is a reference application created to show how
 to develop Symfony applications following the recommended best practices.
 
-[![Build Status](https://travis-ci.org/symfony/symfony-demo.svg?branch=master)](https://travis-ci.org/symfony/symfony-demo)
+[![Deploy](https://www.herokucdn.com/deploy/button.png)](https://heroku.com/deploy) [![Build Status](https://travis-ci.org/symfony/symfony-demo.svg?branch=master)](https://travis-ci.org/symfony/symfony-demo)
 
 Requirements
 ------------
--- a/app.json
+++ b/app.json
@@ -1,0 +1,26 @@
+{
+    "name": "Symfony Demo Application",
+    "description": "The official demo application for the Symfony framework.",
+    "keywords": [
+        "php",
+        "symfony"
+    ],
+    "website": "https://symfony.com/download",
+    "repository": "https://github.com/symfony/symfony-demo",
+    "logo": "https://symfony.com/images/v5/pictos/demoapp.svg?v=4",
+    "success_url": "/",
+    "scripts": {
+        "postdeploy": "php app/console doctrine:schema:create && php app/console doctrine:fixtures:load -n"
+    },
+    "env": {
+        "SYMFONY_ENV": "prod",
+        "SYMFONY_SECRET": {
+            "description": "Extra entropy for %kernel.secret%; used for CSRF tokens, cookies and signed URLs.",
+            "generator": "secret"
+        }
+    },
+    "addons": [
+        "heroku-postgresql"
+    ],
+    "image": "heroku/php"
+}
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -23,6 +23,7 @@ class AppKernel extends Kernel
             new Knp\Bundle\PaginatorBundle\KnpPaginatorBundle(),
             new CodeExplorerBundle\CodeExplorerBundle(),
             new AppBundle\AppBundle(),
+            new Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle(),
         );
 
         // Some bundles are only used while developing the application or during
@@ -34,7 +35,6 @@ class AppKernel extends Kernel
             $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
             $bundles[] = new Sensio\Bundle\DistributionBundle\SensioDistributionBundle();
             $bundles[] = new Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle();
-            $bundles[] = new Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle();
         }
 
         return $bundles;
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -77,16 +77,7 @@ doctrine:
         # them from the app/config/parameters.yml file. The reason is that config.yml
         # stores options that change the application behavior and parameters.yml
         # stores options that change from one server to another
-        driver:   "%database_driver%"
-        host:     "%database_host%"
-        port:     "%database_port%"
-        dbname:   "%database_name%"
-        user:     "%database_user%"
-        password: "%database_password%"
-        charset:  UTF8
-        # if using pdo_sqlite as your database driver, add the path in parameters.yml
-        # e.g. database_path: "%kernel.root_dir%/data/data.db3"
-        path:     "%database_path%"
+        url: "%database_url%"
 
     orm:
         auto_generate_proxy_classes: "%kernel.debug%"
--- a/app/config/config_prod.yml
+++ b/app/config/config_prod.yml
@@ -19,7 +19,7 @@ monolog:
             handler:      nested
         nested:
             type:  stream
-            path:  "%kernel.logs_dir%/%kernel.environment%.log"
+            path:  "php://stderr"
             level: debug
         console:
             type:  console
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -5,23 +5,12 @@
 parameters:
     # this demo application uses an embedded SQLite database to simplify setup.
     # in a real Symfony application you probably will use a MySQL or PostgreSQL database
-    database_driver:   pdo_sqlite
-    database_host:     127.0.0.1
-    database_port:     ~
-    database_name:     ~
-    database_user:     root
-    database_password: ~
-    # the 'database_path' is only used for SQLite type databases
-    database_path:     '%kernel.root_dir%/data/blog.sqlite'
+    # the path must be relative or else it will not work on Windows
+    database_url:      'sqlite:///%kernel.root_dir%/data/blog.sqlite'
 
-    # Uncomment these lines to use a MySQL database instead of SQLite:
+    # Uncomment this line to use a MySQL database instead of SQLite:
     #
-    # database_driver: pdo_mysql
-    # database_host: 127.0.0.1
-    # database_port: ~
-    # database_name: symfony_demo
-    # database_user: root
-    # database_password: ~
+    # database_url: mysql://root:pass@127.0.0.1:3306/symfony_demo
     #
     # You can even create the database and load the sample data from the command line:
     #
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,11 @@
         "symfony-web-dir": "web",
         "symfony-assets-install": "relative",
         "incenteev-parameters": {
-            "file": "app/config/parameters.yml"
+            "file": "app/config/parameters.yml",
+            "env-map": {
+                "database_url": "DATABASE_URL",
+                "secret": "SYMFONY_SECRET"
+            }
         }
     }
 }
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -7,6 +7,11 @@ framework:
         strict_requirements: true
     profiler: { only_exceptions: false }
 
+doctrine:
+    dbal:
+        # temp workaround for https://github.com/doctrine/dbal/issues/1106: define DB path here
+        path: "%kernel.root_dir%/data/blog.sqlite"
+
 web_profiler:
     toolbar: %kernel.debug%
     intercept_redirects: false
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "a41f6ba0108ab7b4a352f6ee652fd314",
-    "content-hash": "e252d4da3f275dbbcdff242ed802d185",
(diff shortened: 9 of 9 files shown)


app.json
```diff
@@ -0,0 +1,22 @@
+{
+	"name": "Symfony Demo Application",
+	"description": "The official demo application for the Symfony framework",
+	"keywords": [
+		"php",
+		"symfony"
+	],
+	"website": "http://symfony.com/download",
```
stof: I suggest using https for the symfony.com website (as it now supports it)

composer.json
```diff
@@ -60,7 +60,10 @@
         "symfony-web-dir": "web",
         "symfony-assets-install": "relative",
         "incenteev-parameters": {
-            "file": "app/config/parameters.yml"
+            "file": "app/config/parameters.yml",
+            "env-map": {
+                "database_url": "DATABASE_URL"
```
stof: you should make the ``secret`` configurable too (a random one should be generated when creating the new application if possible, and should not change for new deployments of the existing app)

app.json
dzuelke: I'll rebase.

composer.json
dzuelke: Great point, will do, thanks!

dzuelke:
There you go @stof!

dzuelke:
Now that AppVeyor build failure is of course not great... :/ Windows path name messes stuff up it seems.

javiereguiluz:
@dzuelke thanks for proposing these changes. We'll gladly help you make this app more Heroku-friendly.

But first, I have some questions for the community and specially for @stof, which was your first reviewer:

  * I don't like to include the DoctrineFixtures bundle in production ... but I can't think of any viable alternative.
  * OK for the `database_url` option, unless someone sees some drawbacks.
  * The only thing I don't like is the change in the production logs. I understand it as: *"let's "break" production logs for everybody so Heroku users can see those logs"*. I'm afraid that logging into `prod.log` is the well-known and documented behavior that users expect from us. Can we solve this differently to make Heroku users happy and the rest of us too?

Thanks!

dzuelke:
One simple approach would be a new environment, e.g. `heroku`, that has that setting, @javiereguiluz.

However, logging to `php://stderr` doesn't really "break" things for others - this way, the logs end up in the web server's error logs. Most other languages' frameworks log to stderr by default, simply because it's easier to consume that log info and direct it elsewhere for central aggregation. It's the Docker standard too.

So it should maybe be(come) the default in the PHP world, too ;) Such a setup is easier to maintain - no worrying about tailing log files, no `logs` folder creation or  no messing around with Docker's filesystem that throws errors on `tail -f`, and so forth. Once you run more than one server, you want to centrally aggregate the logs anyway (since otherwise you need to log into each of your machines until you find the one that served that request), and then writing to a file is an unnecessary intermediate step.

It should also be the default for development, because then errors and other log info would end up where they belong - in the console output for `app/console server:run` (currently, nothing gets displayed there, but that's fixable in Symfony). All Python/Ruby/Node/... web servers I am familiar with do that, too.

dzuelke:
We can also leave the log setting as is, and I drop a `Procfile` into the repo (it currently gets auto-generated by `prepareDeploymentTarget`) that explicitly tails `error.log`, but I figured the `stderr` was simpler...

dzuelke:
Okay so `url` breaks SQLite runs because of https://github.com/doctrine/dbal/issues/1106, investigating a workaround until DBAL 2.5.5 is out...

dzuelke:
AppVeyor fixed. Threw a very basic workaround for the SQLite issue into `config_dev.yml`; can we leave it like that for the moment?

@stof what's your thoughts on the `stderr` situation? IMO Symfony as a whole should move towards that as the default, because it means logs are "free" - e.g. in a Docker run, they will automatically bubble up into a `docker-compose` output etc.

app/config/config.yml
```diff
@@ -77,16 +77,7 @@ doctrine:
...
-        host:     "%database_host%"
-        port:     "%database_port%"
-        dbname:   "%database_name%"
-        user:     "%database_user%"
-        password: "%database_password%"
-        charset:  UTF8
-        # if using pdo_sqlite as your database driver, add the path in parameters.yml
-        # e.g. database_path: "%kernel.root_dir%/data/data.db3"
-        path:     "%database_path%"
+        url:      "%database_url%"
```
fabpot: Even if configuring "just" the URL seems better at first (and I understand why this is needed for Heroku), it breaks how people configure a Symfony app right now: config.yml can contain settings for all environments (like the driver), and config_dev.yml and config_prod.yml (or their parameters.yml counterparts) can "just" change the moving parts like the username or the password.

app/config/config_prod.yml
```diff
@@ -19,7 +19,7 @@ monolog:
             handler:      nested
         nested:
             type:  stream
-            path:  "%kernel.logs_dir%/%kernel.environment%.log"
+            path:  "php://stderr"
```
fabpot: I tend to agree with David here, logging to STDERR by default becomes the de-facto standard and helps a lot with 12-factor style apps and Docker deployments. Of course, this change could also be made on the Symfony Standard Edition and could lead to changes to the documentation if we go down this route.

app/config/parameters.yml.dist
```diff
@@ -5,23 +5,12 @@
...
 
     # Uncomment these lines to use a MySQL database instead of SQLite:
     #
-    # database_driver: pdo_mysql
-    # database_host: 127.0.0.1
-    # database_port: ~
-    # database_name: symfony_demo
-    # database_user: root
-    # database_password: ~
+    # database_url: mysql://root@pass127.0.0.1:3306/symfony_demo
```
fabpot: Now that I see the value, I think my comment above is irrelevant as you can also include parameters in a string. So, using `database_url` seems fine after all.

bocharsky-bw: Why we don't want reuse database parameters here like:
`# database_url: mysql://%database_user%@%database_password%%database_host%:%database_port%/%database_name%`

Is it too complex?

bocharsky-bw: @dzuelke BTW, seems you missed a colon between password and host (`mysql://root@pass:127.0.0.1:3306/symfony_demo`), didn't you?

app/config/config_prod.yml
xabbuh: sounds like a good idea to me

app/config/parameters.yml.dist
dzuelke: Ah yes @bocharsky-bw thanks.

Using URLs for this purpose is not proprietary to Heroku, by the way ;) Doctrine supports them (that's why they work here), and many other languages/frameworks (Rails, Django, ...) because it make management of credentials much less of a hassle especially if you do it through the environment in e.g. Docker or 12-factor apps - `redis://..`, `smtp://user:pass@server:1234/` and so forth, they work for anything.

app/config/config_prod.yml
dzuelke: @fabpot currently, `app/console server:run` will discard stderr output, even with `--verbose` - I had a brief look yesterday to see why but haven't found the problem yet. I've made a reminder to do that; would be useful IMO (including a warning or even `ini_set` call if `error_log` is set to a file for CLI, which really it shouldn't be).

dzuelke: (that's why I only touched the setting for `prod` for now)

dzuelke:
Bump @javiereguiluz @stof @fabpot? :)

README.md
```diff
@@ -4,7 +4,7 @@ Symfony Demo Application
 The "Symfony Demo Application" is a reference application created to show how
 to develop Symfony applications following the recommended best practices.
 
-[![Build Status](https://travis-ci.org/symfony/symfony-demo.svg?branch=master)](https://travis-ci.org/symfony/symfony-demo)
+[![Deploy](https://www.herokucdn.com/deploy/button.png)](https://heroku.com/deploy) [![Build Status](https://travis-ci.org/symfony/symfony-demo.svg?branch=master)](https://travis-ci.org/symfony/symfony-demo)
```
fabpot: This one should be removed and added to the symfony.com download page instead.

app.json
```diff
@@ -0,0 +1,26 @@
...
+		"SYMFONY_SECRET": {
+			"description": "Extra entropy for %kernel.secret%; used for CSRF tokens, cookies and signed URLs.",
+			"generator": "secret"
+		}
+	},
+	"addons": [
+		"heroku-postgresql"
+	],
+	"image": "heroku/php"
+}
```
fabpot: missing newline

README.md
dzuelke: The "original" idea of the button is to live in repos' `README`s. The trick is that with a regular `heroku.com/deploy` link, the Heroku dashboard reads the repo from the `Referer` (sic) header, so if someone forks the "symfony-demo" repo, that fork is also instantly deployable without edits, including branches, which is quite nice for PRs etc, because the corresponding branch can be deployed by anyone as well.

Having it on symfony.com/demo would of course be fantastic ;) but it's a separate thing.

app.json
dzuelke: Will fix!

dzuelke:
I fixed a few formatting things that have been noted by people. I'll squash that into other commits once everyone is happy.

Anything else before you'd like to merge it, @javiereguiluz and @fabpot ? :)

app.json
```diff
@@ -0,0 +1,26 @@
+{
+	"name": "Symfony Demo Application",
```
bocharsky-bw: I think we should use 4 spaces instead of a tab for indentation, don't we? Spaces indentation used everywhere in this project.

app/AppKernel.php
```diff
@@ -23,6 +23,7 @@ public function registerBundles()
             new Knp\Bundle\PaginatorBundle\KnpPaginatorBundle(),
             new CodeExplorerBundle\CodeExplorerBundle(),
             new AppBundle\AppBundle(),
+            new Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle(),
```
bocharsky-bw: Is there any reasons to move `DoctrineFixturesBundle` from `dev` to `prod` environment section?

dzuelke: Yes, otherwise you can't populate the fixtures! It's also already a regular (non-dev) requirement in `composer.json`.

app.json
dzuelke: No problem; pushed a change, will squash later.

app/AppKernel.php
bocharsky-bw: Why do we simply can't use `dev` environment to populate the fixtures instead of allowing it in `prod`?

P.S. Sorry, maybe I don't fully understand Heroku.

stof: The dev environment requires ``sensio/generator-bundle`` to be bootable, which is not a regular requirement and so is not available on heroku (as it runs composer with ``--no-dev``)

dzuelke: If you populate fixtures in production to your production database, then it's not a `dev` requirement. `require-dev` is for stuff like PHPUnit, generator bundles, dev/debug toolbars etc.

dzuelke: This is a demo app, and it has no initial database migration step with e.g. a default admin login - it uses the fixtures bundle instead, so it's fine to use that in production in this special case.

dzuelke: Furthermore, the `dev` environment has different database credentials ;)

dzuelke: (related opinion of mine: even if it's a demo app, it should still not have an SQLite database file in the repo and instead have a one-line instruction on how to populate it, but that's a different discussion :p)

bocharsky-bw: @stof and @dzuelke thanks for an explanation, guys!

stof: @dzuelke I tend to agree with you. given that we have a specific installer for the demo in the symfony-installer, we should be loading fixtures there instead (as done here).

README.md
fabpot: fair enough, let's keep it like this.

fabpot:
LGTM

stof:
:+1:

dzuelke:
Great. Wielding my rebase chainsaws...

dzuelke:
(almost done; rebasing `composer.lock` changes)

dzuelke:
Okay, that should be it. Merge away, @fabpot or @javiereguiluz!

stof:
:+1:

javiereguiluz:
@dzuelke thanks for adding this feature and thanks for your patience and the detailed explanations during the review process. I've squashed your commits before merge, that's why GitHub shows it as "closed" instead of "merged".

bocharsky-bw:
@javiereguiluz Sorry for spam, but could you tell me, Did you squash this commits manually or you use some tool for it? :innocent:

javiereguiluz:
@bocharsky-bw I use a command-line tool which is available to Symfony Core members. I'm afraid that the tool is not public at the moment.

stof:
@bocharsky-bw we have an internal tool doing this for us automatically (but this is a matter of running some ``git rebase`` commands. The tool is just here to make it a one-liner)

dzuelke:
Thanks @javiereguiluz :) I'll happily squash the commits for you next time. I'd have left the `config_dev.yml` workaround in a separate commit, but minor thing (I'll do a PR to remove that change once DBAL 2.5.5 with a fix for doctrine/dbal#1106 is released).

bocharsky-bw:
@javiereguiluz and @stof Thanks! I supposed that should be an internal tool, it provides a nice description of squashed commits. I even suppose you have an internal tool to release projects )

PR #300, opened 2016-01-29
Request::enableHttpMethodParameterOverride for AppCache

### LINKED ISSUE #299: HTTP 405 in production when deleting a post
In production deleting a post in the admin interface gives a 405 error.

The reason is that the AppCache interferes with `http_method_override` being enabled, see notes at http://symfony.com/doc/current/reference/configuration/framework.html#http-method-override

What should the solution be? I'd say adding `Request::enableHttpMethodParameterOverride();` to `web/app.php`, but the alternative of course is disabling `http_method_override` (not supported by most browsers though AFAIK), so I just wanted to double check before sending a pull request :)

dzuelke:
Fixes #299

### DIFF
--- a/web/app.php
+++ b/web/app.php
@@ -38,7 +38,7 @@ $kernel = new AppCache($kernel);
 // If you use HTTP Cache and your application relies on the _method request parameter
 // to get the intended HTTP method, uncomment this line.
 // See http://symfony.com/doc/current/reference/configuration/framework.html#http-method-override
-// Request::enableHttpMethodParameterOverride();
+Request::enableHttpMethodParameterOverride();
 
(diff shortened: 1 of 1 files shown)


dzuelke:
Bump? :)

javiereguiluz:
@dzuelke thanks for this fix. It's merged now and it'll be available in the next stable release.

web/app.php
```diff
@@ -38,7 +38,7 @@
 // If you use HTTP Cache and your application relies on the _method request parameter
 // to get the intended HTTP method, uncomment this line.
```
xabbuh: This comment must now be changed (or removed).

PR #303, opened 2016-02-09
Undo setting Post to null on comment removal

### LINKED ISSUE #302: Null value not valid for setPost function parameter
- File: src\AppBundle\Entity\Post.php
- Line: 171
- Function Code:
```
public function removeComment(Comment $comment)
    {
        $this->comments->removeElement($comment);
        $comment->setPost(null);
    }
```
PHP doesn't accept null value for a type-hinting parameter (you can set default value to null but it seems a workaround not a real solution in my opinion).

Result: ContextErrorException, Catchable Fatal Error: Argument 1 passed to AppBundle\Entity\Comment::setPost() must be an instance of AppBundle\Entity\Post, null given, called in ...

PHP: 5.6.15 using included web server

Excuse my poor english.

bocharsky-bw:
Fix #302.

### DIFF
--- a/src/AppBundle/Entity/Post.php
+++ b/src/AppBundle/Entity/Post.php
@@ -168,7 +168,6 @@ class Post
     public function removeComment(Comment $comment)
     {
         $this->comments->removeElement($comment);
-        $comment->setPost(null);
     }
 
     public function getSummary()


javiereguiluz:
@bocharsky-bw thanks for your quick reaction to fix this reported bug!

PR #304, opened 2016-02-09
Improve routes and fix bug with pagination

voronkovich:
For now the paginator returns an empty array when requested page is out of range. I think we should handle this situation manually and throw the 404 exception. See https://github.com/KnpLabs/knp-components/issues/78
Also, route definition accepts 0 as page parameter and paginator throws an exception.


javiereguiluz:
:+1: these changes look good to me.

The only thing I don't like much is the `requirements={"page": "[1-9]\d*"}` requirement, but I don't know a prettier way to say "page number must be an positive integer different from 0`.

javiereguiluz:
Merged! Thanks @voronkovich

PR #306, opened 2016-02-11

davialexandre:
Translations for the error pages and "save and create new" button


javiereguiluz:
@davialexandre thanks for this nice translation update! *Muito obrigado*

PR #307, opened 2016-02-12
Update Composer before dependencies installation

bocharsky-bw:
Travis use the old Composer's version by default


javiereguiluz:
@bocharsky-bw a quick win! Thank you.

PR #308, opened 2016-02-12
disable Xdebug on Travis CI

xabbuh:


.travis.yml
```diff
@@ -18,6 +18,7 @@ matrix:
     - php: hhvm
 
 before_install:
+  - if [[ "$TRAVIS_PHP_VERSION" != "hhvm" ]; then phpenv config-rm xdebug.ini; fi
```
javiereguiluz: Just asking: why the `hhvm` check?

xabbuh: There's no Xdebug extension on HHVM.

voronkovich: @xabbuh, missing ]
BTW all shells support short syntax:
```sh
[[ "$TRAVIS_PHP_VERSION" != "hhvm" ]] && phpenv config-rm xdebug.ini
```

javiereguiluz: @xabbuh yes ... let me ask differently: why not remove `xdebug.ini` always? After all, HHVM doesn't have xdebug, so the result would be the same.

xabbuh: @javiereguiluz IIRC the command would fail if the file didn't exist.

javiereguiluz: ok, merging then :)

javiereguiluz:
Thanks @xabbuh! We all love fast tests :)

PR #311, opened 2016-02-23
Use non-deprecated APC validator and serializer service names

bocharsky-bw:
According to the https://github.com/symfony/symfony-standard/pull/901 .


javiereguiluz:
The old syntax looked much more human-friendly ... but we must use the new syntax. Thanks @bocharsky-bw.

PR #313, opened 2016-03-06
Wrap MySQL config URL with single quotes

bocharsky-bw:
Quotes allows avoiding YAML's syntax error after uncommenting.


javiereguiluz:
@bocharsky-bw thanks for this fix. This was indeed needed.

Issue #314, opened 2016-03-17
Running composer install leaves the project in a dirty state


stof:
There is a composer hook updating the requirement file, but it also updates ``web/config.php``, while the demo has a customized version.
This makes it harder to contribute as we have to remember to undo the updates all the time.

dunglas:
What do you think about renaming the customized file to something like `web/demo-config.php`?

bocharsky-bw:
@dunglas IIRC, the reason why we didn't rename it was that [app/check.php](https://github.com/symfony/symfony-demo/blob/master/app/check.php#L71) references to the `web/config.php` file, so renaming this file to something different will be confusing.

dunglas:
We can also introduce a `app/check-demo.php`.

bocharsky-bw:
Yes, but then we need to get rid of [ScriptHandler::installRequirementsFile](https://github.com/symfony/symfony-demo/blob/master/composer.json#L40) in Composer's post install and post update commands. In this case, we reject DistributionBundle requirements, so we need to support them manually.

stof:
@bocharsky-bw we would still need to update the requirements file, as we want to have an uptodate one in the demo

javiereguiluz:
I'd say ... let's remove the customized `web/config.php` file and forget about this problem forever.

bocharsky-bw:
Could we keep original `web/config.php` instead of customized?

javiereguiluz:
@bocharsky-bw yes. I meant removing our modifications and leave whatever Symfony provides.

bocharsky-bw:
@javiereguiluz Looks like it's the easiest way :+1:

dunglas:
:+1+: for Javier suggestion too.

PR #315, opened 2016-03-20
[WIP] Use controller autowiring / DunglasActionBundle

dunglas:
Follows symfony/symfony-standard#959


app/config/services.yml
```diff
@@ -1,18 +1,13 @@
 services:
-    # First we define some basic services to make these utilities available in
-    # the entire application
-    slugger:
```
dunglas: This is the interesting part here.

javiereguiluz: In a real application, these services would be used in other services, in console commands, etc.  You cannot remove them.

dunglas: I agree than in a real app *large* app I would explicitly defined at least the `slugger` service. However, it's not necessary in this sample RAD app and it shows how the new controller system can be used (for RAD only of course).

Just some hints:

* those services hey are used in other services here (`@markdown` is used in `@app.twig.app_extension`)
* an advantage of using autowiring is if you change signature of the constructor of the `AppExtension ` or if you rename the `markdown` service, you don't need to update your config
* if the service using `@markdown` is itself autowired the `@markdown` is automatically injected (`@app.twig.app_extension`)
* defining explicitly the service (adding back the `@slugger` definition for instance) doesn't break anything (the named service will be detected and injected everywhere the autowiring service is defined)
* by convention, all services created by the autowiring system are accessible using the `autowired.the\fqn` name (but it's better to name them explicitly if you want to access them). Here there is a `@autowired.AppBundle\Utils\Slugger` service.

Btw, console commands can be autowired the same way than controller, I'm adding an example right now.

app/config/routing.yml
```diff
@@ -5,8 +5,8 @@
 # the routes in YAML, XML or PHP files.
 # See http://symfony.com/doc/current/book/routing.html
 app:
-    resource: '@AppBundle/Controller/'
-    type:     annotation
+    resource: '@AppBundle/Action/'
```
stof: IMO, keeping the controllers in ``Controller`` is better, even if you rely on a special loader aware of the way they are registered as services.

composer.json
```diff
@@ -50,10 +57,7 @@
         ]
     },
     "config": {
-        "bin-dir": "bin",
-        "platform": {
-            "php": "5.3.9"
```
stof: not good. We want to keep the demo compatible with older PHP versions, to ensure that someone running ``symfony demo`` gets a working demo

@javiereguiluz we may switch to 5.4 as min target though, as the installer itself requires 5.4 anyway.

composer.json
```diff
@@ -25,8 +31,9 @@
         "symfony/assetic-bundle"               : "~2.6",
         "symfony/monolog-bundle"               : "~2.7",
         "symfony/swiftmailer-bundle"           : "~2.3",
-        "symfony/symfony"                      : "~2.8",
-        "twig/extensions"                      : "~1.2"
+        "symfony/symfony"                      : "dev-action as 3.1.0",
+        "twig/extensions"                      : "~1.2",
+        "dunglas/action-bundle"                : "dev-master"
```
stof: unbound requirements are a bad idea. Please probably a sever-compatible branch alias in your bundle instead, and use it here

app/config/routing.yml
dunglas: Yes I've opened an alternative PR for that. But I think I'll do it even in the default setup of the bundle. It will be more familiar for Symfony developers (and it seems to be a general concern).

web/config.php
```diff
@@ -1,10 +1,23 @@
 <?php
 
+/*
+ * ************** CAUTION **************
+ *
+ * DO NOT EDIT THIS FILE as it will be overridden by Composer as part of
+ * the installation/update process. The original file resides in the
+ * SensioDistributionBundle.
+ *
+ * ************** CAUTION **************
```
stof: Changes in this file must be reverted, as the demo contains a customized version (see #314 for an issue about not requiring manual reverts each time though)

composer.json
dunglas: It's just because it's a WIP. I'll release a 1.0 version before the merge and I'll update this PR to use it.

composer.json
dunglas: Same here. I'll restore it as is zhen it Zillow be ready. But this PR will require Symfony 3.1 anyway.

stof: Well, this should be changed to ``5.5.9`` for the platform once the demo runs on Symfony 3.

composer.json
stof: well, having the branch alias already set would still allow to write this constraint as ``~1.0@dev``

javiereguiluz:
I'm closing this as "won't merge" because the initial idea of enabling DunglasActionBundle by default on the Symfony Standard edition was abandoned. Symfony Demo must be "Symfony Standard compliant", so we can't make this change by default in the Symfony Demo. Thanks.

dunglas:
@javiereguiluz AFAIK this is not rejected yet (https://github.com/symfony/symfony-standard/pull/960).
But I agree that it will not be for 3.1. (I'll propose an integration path soon).

PR #317, opened 2016-03-27
Update dependencies according to Symfony v2.8.4 release

bocharsky-bw:
Tweak `config.php` file a bit to be more consistent with the original script.

### DIFF
--- a/web/config.php
+++ b/web/config.php
@@ -1,10 +1,13 @@
 <?php
 
 if (!isset($_SERVER['HTTP_HOST'])) {
-    exit('This script cannot be run from the console. Run it from a browser.');
+    exit('This script cannot be run from the CLI. Run it from a browser.');
 }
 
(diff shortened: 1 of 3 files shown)


web/config.php
```diff
@@ -1,10 +1,23 @@
...
 
+/*
+ * ************** CAUTION **************
+ *
+ * DO NOT EDIT THIS FILE as it will be overridden by Composer as part of
+ * the installation/update process. The original file resides in the
+ * SensioDistributionBundle.
+ *
+ * ************** CAUTION **************
+ */
```
stof: This header is wrong, as we are precisely *not* using the original file here

bocharsky-bw: You're right, thanks!

javiereguiluz:
Thanks @bocharsky-bw.

PR #318, opened 2016-03-29
Fix scss compiling with assetic

f-plante:
Update leafo/scssphp in order to make assetic compile sass files.

With the current symfony demo, the sass compiling with assetic does not work well :
It seems to work in console with `app/console assetic:dump --no-debug` but not when commenting out related lines in `base.html.twig`.

Error message in the log :
````
[2016-03-29 15:42:55] request.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\ContextErrorException: "Notice: Undefined offset: 0" at /home/developpeur/Sources/symfony-demo/vendor/leafo/scssphp/src/Compiler.php line 278 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\ContextErrorException(code: 0): Notice: Undefined offset: 0 at /home/developpeur/Sources/symfony-demo/vendor/leafo/scssphp/src/Compiler.php:278)"} []
````
After updating `leafo/scssphp`  assetic compilation in controller works as expected.

Related issue : https://github.com/symfony/symfony-demo/issues/295


javiereguiluz:
I can confirm the reported issue and I've also tested that this PR solves it. @f-plante thank you very much for fixing this. Merging it!

PR #319, opened 2016-03-30
Fix minor assets related issues

f-plante:
This PR address some minor issues :

The bootswatch-flatly theme integration was not properly done. 
The `bootswatch-flatly/_variables.scss` partial is meant to be included _before_ the bootstrap files, and the `bootswatch-flatly/_bootswatch.scss` file _after_. Otherwise, some flatly specific styling are overridden by the standard ones (You can see that with the form field focus, or the table style)

The asset files should not have exec permissions.


javiereguiluz:
@f-plante thank you very much for this! I've tested it and it works great. I'm embarrassed for not having integrated the theme properly :blush:  And you are so right about the form field focus: the difference is like night and day :)

PR #320, opened 2016-03-30
Fix minor flash handling

f-plante:
This PR remove the useless flash partial include in `edit.html.twig` (as it is already included in `base.html.twig`).

I tweaked the flash partial a little to output `<div class="messages">` only if there is really something to display in the flashbag.

### DIFF
--- a/app/Resources/views/default/_flash_messages.html.twig
+++ b/app/Resources/views/default/_flash_messages.html.twig
@@ -7,7 +7,7 @@
    '_flash_messages.html.twig' instead of 'flash_messages.html.twig'
 #}
 
-{% if app.session.started %}
+{% if app.session.started and app.session.flashBag.has('success') %}
     <div class="messages">
         {% for message in app.session.flashBag.get('success') %}
             {# Bootstrap alert, see http://getbootstrap.com/components/#alerts #}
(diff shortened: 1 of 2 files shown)


app/Resources/views/default/_flash_messages.html.twig
```diff
@@ -7,9 +7,9 @@
    '_flash_messages.html.twig' instead of 'flash_messages.html.twig'
 #}
 
-{% if app.session.started %}
+{% if app.session.started and app.session.flashbag.has('success') %}
```
xabbuh: should be `flashBag`

app/Resources/views/default/_flash_messages.html.twig
```diff
@@ -7,9 +7,9 @@
...
     <div class="messages">
-        {% for message in app.session.flashBag.get('success') %}
+        {% for message in app.session.flashbag.get('success') %}
```
xabbuh: This should be reverted.

f-plante:
I reverted the code style issue regarding flashBag vs flashbag. The shame is that I checked first in [the documentation](http://symfony.com/doc/current/book/controller.html) and the twig exemple use `flashbag`.

stof:
@f-plante given that PHP method names are case-insensitive, both work

f-plante:
Yes, I know that. That's why I called it a _code style issue_. This seems important to @xabbuh to keep the case as-is.

javiereguiluz:
@f-plante you are on a roll! Thanks for your contributions. Please keep them coming.

PR #323, opened 2016-03-30
Make the retrieval of template source compatible with Twig 2

### LINKED ISSUE #322: In Twig 2.x, _self is a string instead of a Template instance
In the templates of this application we use the following to display the source code of the current Twig template:

```twig
{{ show_source_code(_self) }}
```

However, Twig 2.x changed the `_self` variable and now it only returns a string of the template path instead of the `Template` object instance.

We should change/improve this feature to make the application compatible with Symfony 3.

stof:
Closes #322


voronkovich:
@stof you're right, this solution simpler than mine :). :+1:

javiereguiluz:
A little but nice change. Thanks!

PR #325, opened 2016-03-30
Fix incorrect 404 response

f-plante:
On the public section, when there are no posts and we are on page 1 we should not return a 404 response.

In fact, that case is already handled in the view.


voronkovich:
Good catch! :+1:

javiereguiluz:
Nice! Thank you.

PR #326, opened 2016-03-31
Iterate over all flash messages in bag

bocharsky-bw:
This PR allows iterating over all message types in the flash bag. It discovers an ability to use different alert types which provided out-of-the-box by Bootstrap and even create a custom one with adding our own `alert-*` CSS classes:

screenshot from 2016-03-31 10-37-54

The code behind this is:
```php
{
    public function indexAction()
    {
        $this->addFlash('success', 'Success message');
        $this->addFlash('info', 'Info message');
        $this->addFlash('warning', 'Warning message');
        $this->addFlash('danger', 'Danger message');
        // ...
    }
}
```

### DIFF
--- a/app/Resources/views/default/_flash_messages.html.twig
+++ b/app/Resources/views/default/_flash_messages.html.twig
@@ -7,17 +7,19 @@
    '_flash_messages.html.twig' instead of 'flash_messages.html.twig'
 #}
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@bocharsky-bw I like it! Even if we don't use all of this in the demo application, it's nice to show people how to do this. Thanks!

PR #331, opened 2016-04-21
Add responsive mobile support by defining viewport

CesarManriqueH:


javiereguiluz:
@CesarManriqueH thanks for reporting this issue in the first place and then for providing the pull request to fix it.

krewetka:
@javiereguiluz When more or less will this be in available in version which installs after using "symfony demo"? 

It took me a while to find out why demo is not working on mobile...and then I found that it's already fixed on master :)

PR #332, opened 2016-05-01
Add translation to pager

krewetka:
Pager was always in English

### DIFF
--- a/app/Resources/translations/messages.ca.xliff
+++ b/app/Resources/translations/messages.ca.xliff
@@ -18,14 +18,6 @@
                 <source>mit_license</source>
                 <target>Llicència MIT</target>
             </trans-unit>
-            <trans-unit id="previous">
-                <source>Previous</source>
-                <target>Anterior</target>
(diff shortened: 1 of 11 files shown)


krewetka:
Removed unnecessary previous & next translations

javiereguiluz:
@krewetka thanks for this contribution! I've tested it locally and it works like a charm. BEfore merging your pull request I've "squashed" your commits. That's why GitHub displays this pull request as "closed" instead of "merged" ... but it's merged :)

krewetka:
You're welcome :)

I knew it's not the best way to make it in separate commits but I haven't find other option on the website and I didn't have time to make it in proper way with downloading it to computer and so on/

javiereguiluz:
@krewetka no worries. I also prefer to make small commits and at the end everything is squashed into a single commit. Luckily the tool we use to merge pull requests does the squashing automatically, so that's why we never ask committers to do anything about that. Thanks!

PR #335, opened 2016-05-10
Added a fallback Twig extension when PHP's intl is not available

### LINKED ISSUE #329: Don't use localizeddate to avoid Intl errors

javiereguiluz:
This fixes #329.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -8,6 +8,7 @@ imports:
     - { resource: parameters.yml }
     - { resource: security.yml }
     - { resource: services.yml }
+    - { resource: services.php }
 
 # These are the configuration parameters that define the application's behavior
 # and which are independent from the underlying technical infrastructure
--- a/app/config/services.php
+++ b/app/config/services.php
@@ -1,0 +1,17 @@
+<?php
+
+/**
+ * This file loads services into the Symfony Container dynamically. This allows
+ * to check if the PHP Intl extension is loaded in the system. If it's loaded,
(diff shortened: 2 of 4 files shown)


javiereguiluz:
@mtsanford @tcicit @WouterJ could you please try the changes proposed in this pull request to see if it solves your problems with Windows + Intl? Thanks!

tcicit:
I can not find the file. In the demo app, there are only the following file: "\symfony_demo\src\AppBundle\Twig\AppExtension.php".

bocharsky-bw:
You should checkout on `fix_329` branch of the https://github.com/javiereguiluz/symfony-demo Javier's fork repo.

javiereguiluz:
@tcicit if you don't want to mess with Git, you can recreate these changes in your own computer. It's just a matter of creating 2 files and modifying 2 other files. See: https://github.com/symfony/symfony-demo/pull/335/files

WouterJ:
While this is probably works fine (didn't test it yet), I think the implementation is very complex (especially for a *demo* app for beginners):

 * Using PHP config to register the service
 * A full-blown polyfill for intl date formatter

What about creating a `render_date()` function? This would allow us to get rid of both problems, makes the implementation quite a lot simpler (as it's just `intl_enabled ? IntlDateFormatter::create(...)->format(new \DateTime()) : date(...);`) and also shows how readable you can make your templates by adding some very tiny functions and filters.

javiereguiluz:
Closed in favor of #344.

PR #336, opened 2016-05-26
+ Added .idea files to gitignore

Elbar:
^ Modified .gitignore to include .idea files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 /bin/
 /composer.phar
 web/bundles/whiteoctoberpagerfanta
+.idea


javiereguiluz:
@Elbar thanks for proposing this improvement. You are right about the need to ignore the `.idea` files ... but there is a problem with the way you did it.

You propose to add this to the project's `.gitignore`. However, the `.idea` files are not related to the project itself but to a particular IDE. These kind of files should be ignored globally in your computer. That way you don't have to repeat this for every project.

This is done with a global `.gitignore` file. GitHub explains it in this guide: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

So, I'm afraid that I must close this issue as "won't fix". I'm sorry and I really hope to merge your next contribution. Thank you!

Elbar:
no problem :+1:

PR #338, opened 2016-05-31
Upgrade the Demo Application to Symfony 3.1

### LINKED ISSUE #337: Upgrading to Symfony 3
I'd like to know the opinion of the community about upgrading the Symfony Demo app to Symfony 3.1.

javiereguiluz:
This fixes #337.


composer.json
```diff
@@ -2,34 +2,35 @@
...
-        "twig/extensions"                      : "~1.2",
+        "doctrine/doctrine-bundle"             : "^1.6",
+        "doctrine/doctrine-fixtures-bundle"    : "^2.2",
+        "doctrine/orm"                         : "^2.5",
+        "erusev/parsedown"                     : "^1.5",
+        "ezyang/htmlpurifier"                  : "^4.7",
+        "incenteev/composer-parameter-handler" : "^2.0",
+        "ircmaxell/password-compat"            : "^1.0",
+        "leafo/scssphp"                        : "^0.4",
+        "patchwork/jsqueeze"                   : "^1.0",
```
stof: I suggest also migrating to version 2

stof:
``var/bootstrap.php.cache`` should not be committed

stof:
the ``bin/console`` file is missing

stof:
Given the 2 above issues, I suspect you did the ``git add`` call before changing the gitignore file...

javiereguiluz:
In eba83d9 you can see that I reverted some logs and DB related changes introduced to make this app more Heroku-friendly:

1) First, errors in `prod` environment are logged again in `var/logs/prod.log` which is what every Symfony developer expects. Otherwise, debugging errors is more complicated than it should.
2) The `database_url` option doesn't work for me. The `prod` environment simply returns `500` for every request. Using the common `path` and `driver` options solved this for me.

stof:
> The database_url option doesn't work for me. The prod environment simply returns 500 for every request.

Have you checked the logs to see what the error is ? It would be great to identify the bug rather than just saying it is buggy and avoiding its usage.

javiereguiluz:
@stof you are right. I apologize! I didn't want to sound harsh. Here are the details.

---

With this config:

```yaml
doctrine:
    dbal:
        driver: "pdo_sqlite"
        path:   "%kernel.root_dir%/../var/data/blog.sqlite"
```

The app works in `prod`.

---

With this other config:

```yaml
parameters:
    database_url: 'sqlite:///%kernel.root_dir%/../var/data/blog.sqlite'

doctrine:
    dbal:
        url: "%database_url%"
```

Every page shows a `500` error because of this:

```
[2016-06-02 11:56:03] request.CRITICAL: Uncaught PHP Exception Twig_Error_Runtime:
"An exception has been thrown during the rendering of a template ("An exception occurred
while executing 'SELECT ... FROM symfony_demo_post s0_ WHERE ...' with params
["2016-06-02 11:56:03"]: 

SQLSTATE[HY000]: General error: 1 no such table: symfony_demo_post")
in "blog/index.html.twig" at line 6."
```

stof:
that looks weird.

voronkovich:
It seems like it creates a new database instead of using an existing one

xabbuh:
You need to update the `app_dev.php` file too.

javiereguiluz:
This is ready for the final review. Thanks!

bin/console
```diff
@@ -14,7 +14,7 @@ set_time_limit(0);
 /**
  * @var Composer\Autoload\ClassLoader $loader
  */
-$loader = require __DIR__.'/autoload.php';
+$loader = require __DIR__.'/../app/autoload.php';
 
 $input = new ArgvInput();
 $env = $input->getParameterOption(array('--env', '-e'), getenv('SYMFONY_ENV') ?: 'dev');
```
voronkovich: I think we should use a new array syntax. This converter can help you https://github.com/thomasbachem/php-short-array-syntax-converter

javiereguiluz: I agree ... but let's do that in a separate PR to not complicate this one too much. In fact, let's push PHP to use as many PHP 5.5 features as we can, not only the short array syntax.

voronkovich: @javiereguiluz, ok

javiereguiluz:
Let's merge this change to test it a bit before releasing the new version of the demo app. Thanks!

PR #340, opened 2016-06-01
Added an example of the dump() function

### LINKED ISSUE #339: Add an example of using dump()
The best way to make this nice function easily discoverable :)

javiereguiluz:
This fixes #339.

### DIFF
--- a/src/AppBundle/Controller/BlogController.php
+++ b/src/AppBundle/Controller/BlogController.php
@@ -56,6 +56,16 @@ class BlogController extends Controller
      */
     public function postShowAction(Post $post)
     {
+        // Symfony provides a function called 'dump()' which is an improved version
+        // of the well-known 'var_dump()' function. It's useful to show the contents
+        // of any variable and check if they contain what you expect.
+        // The 'dump()' function is not available in the 'prod' environment, so use
+        // it only while developing the application, not in production. To make this
+        // app work in both environments, we check the value of the 'kernel.debug' parameter.
+        if (true === $this->getParameter('kernel.debug')) {
+            dump($post, $this->get('security.token_storage')->getToken()->getUser(), new \DateTime());
+        }
+
         return $this->render('blog/post_show.html.twig', array('post' => $post));
     }
 
--- a/src/AppBundle/Controller/BlogController.php
+++ b/src/AppBundle/Controller/BlogController.php
@@ -60,6 +60,8 @@ class BlogController extends Controller
         // of the 'var_dump()' function. It's useful to quickly debug the contents
         // of any variable, but it's not available in the 'prod' environment to
         // prevent any leak of sensitive information.
+        // This function can be used both in PHP files and Twig templates. The only
+        // requirement is to have enabled the DebugBundle.
         if ('dev' === $this->getParameter('kernel.environment')) {
             dump($post, $this->get('security.token_storage')->getToken()->getUser(), new \DateTime());
         }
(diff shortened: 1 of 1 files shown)


src/AppBundle/Controller/BlogController.php
```diff
@@ -56,6 +56,16 @@ public function indexAction($page)
      */
     public function postShowAction(Post $post)
     {
+        // Symfony provides a function called 'dump()' which is an improved version
+        // of the well-known 'var_dump()' function. It's useful to show the contents
+        // of any variable and check if they contain what you expect.
+        // The 'dump()' function is not available in the 'prod' environment, so use
+        // it only while developing the application, not in production. To make this
+        // app work in both environments, we check the value of the 'kernel.debug' parameter.
+        if (true === $this->getParameter('kernel.debug')) {
```
stof: the availability depends on the environment, not on the debug mode. These 2 things can be configured independently.

javiereguiluz: Yes ... but it's always hard to explain things for newcomers in this demo app. Let's wait and see if someone comes up with a better alternative. Thanks!

javiereguiluz: I mean, the obvious alternative is: `function_exists('dump')` but is that better? Some people may thing that `dump()` is not always available or that you have to install something first ?

stof: ``if ('dev' === $this->getParameter('kernel.environment')) {``

voronkovich: I think using `function_exists` is much better. For example, in the client side we detect features instead of browsers. In this case the `dump` is "feature", the environment is "browser".

voronkovich: We could just add a wrapper function:
```php
if (!function_exists('dump')) {
    function dump()
    {
        // calling var_dump
    }
}
```

javiereguiluz: @voronkovich in a real application, that trick is nice. However, in this app it may be confusing:

1) If users see the trick, they might think that they need to do that in their apps.
2) If they don't see the trick, they might think that `dump()` always works but has a random behavior (the output change without reason).

bocharsky-bw: As I understand correctly, we'll get an error if we call `dump()` function in `prod` environment even with enabled `kernel.debug`.

Let's keep this `dump()` function's call only for `dev`? In this case, we don't need any workaround.

javiereguiluz:
At the end I used the solution proposed by @stof (`if ('dev' === $this->getParameter('kernel.environment'))`). This is now ready for review. Thanks!

src/AppBundle/Controller/BlogController.php
```diff
@@ -56,6 +56,14 @@ public function indexAction($page)
      */
     public function postShowAction(Post $post)
     {
+        // Symfony provides a function called 'dump()' which is an improved version
+        // of the 'var_dump()' function. It's useful to quickly debug the contents
+        // of any variable, but it's not available in the 'prod' environment to
+        // prevent any leak of sensitive information.
```
stof: I suggest adding a sentence in the comment explaining that this feature is provided by DebugBundle, and so can only be used in environments where the bundle is enabled.

voronkovich: I agree with @stof. Also I think we should note that the `dump` function can be used in templates and console commands.

javiereguiluz: I've updated the help note with your suggestions. Thanks.

PR #341, opened 2016-06-02
fixes #329: Require ext-intl in the composer.json

### LINKED ISSUE #329: Don't use localizeddate to avoid Intl errors

aminin:


voronkovich:
@aminin, you should add it to the `suggest` section https://getcomposer.org/doc/04-schema.md#suggest
See https://github.com/symfony/symfony-demo/pull/335

javiereguiluz:
@aminin the solution that you suggest is correct. However, it's too hard to comply for some users because the `intl` extension is a bit tricky. That's why we've merged #344 to make this app work without that extension. Therefore, I'm closing this as "won't fix".

PR #344, opened 2016-06-03
Made the application work when PHP's intl is not available

### LINKED ISSUE #329: Don't use localizeddate to avoid Intl errors

javiereguiluz:
This fixes #329.


stof:
:+1:

PR #348, opened 2016-06-19
Fix a PHP version requirenment in composer file

voronkovich:


javiereguiluz:
@voronkovich nice catch and a very important fix! Thank you.

PR #349, opened 2016-06-19
Gitignore: ignore all but a console and a requirements checker

voronkovich:


javiereguiluz:
@voronkovich another nice fix! Thank you very much.

PR #350, opened 2016-06-19
Use new php features

voronkovich:
See https://github.com/symfony/symfony-demo/issues/346

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -11,7 +11,7 @@ class AppKernel extends Kernel
         // application, you must add it in the following array to register it
         // in the application. Otherwise, the bundle won't be enabled and you
         // won't be able to use it.
-        $bundles = array(
+        $bundles = [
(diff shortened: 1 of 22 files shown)


javiereguiluz:
It feels so nice to use non-prehistoric PHP code! Thanks @voronkovich.

If anyone has other ideas to use PHP 5.5 features in the demo app, please share them!

voronkovich:
Should we use:
```php
$entityManager->getRepository(User::class)
```
instead of:
```php
$entityManager->getRepository('AppBundle:User')
```
?

stof:
@voronkovich I would say yes. ``AppBundle:`` shortcuts would probably not be implemented in Doctrine if the project started today.

voronkovich:
@stof, I agree with you. I've changed the code.

javiereguiluz:
@voronkovich thanks for working on this. The next version of the app not only will use Symfony 3.1 but also PHP 5.5!

PR #351, opened 2016-06-19
Fix a datetime picker widget's name

voronkovich:


javiereguiluz:
@voronkovich thanks for the fix ... and for the nice explanation!

PR #353, opened 2016-06-20
Fix console path in the comments

chapay:
We moved console to the new directory, but forgot about changing it in the comments. This PR fixes it.


xabbuh:
👍

javiereguiluz:
I can't believe I forgot to update all this! @chapay thanks for fixing it.

PR #354, opened 2016-06-21
Use same URL for login form and check

WouterJ:
The Security Form login listener will only intercept the request when it's a POST request, otherwise the controller will be executed. Using the same URL makes things a bit easier, removes the need for the weird controller (which is visited, but should not be executed) and makes things similar to other forms.

Related doc tickets:
 * https://github.com/symfony/symfony-docs/issues/6126
 * https://github.com/symfony/symfony-docs/pull/6143

### DIFF
--- a/src/AppBundle/Controller/SecurityController.php
+++ b/src/AppBundle/Controller/SecurityController.php
@@ -25,8 +25,7 @@ use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 class SecurityController extends Controller
 {
     /**
-     * @Route("/login", name="security_login_form")
-     * @Method("GET")
+     * @Route("/login", name="security_login")
      */
     public function loginAction()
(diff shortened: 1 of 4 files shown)


src/AppBundle/Controller/SecurityController.php
```diff
@@ -25,8 +25,7 @@
 class SecurityController extends Controller
 {
     /**
-     * @Route("/login", name="security_login_form")
-     * @Method("GET")
```
WouterJ: This had to be removed, as the router will error when we submit the form. It wasn't very usefull anyways...

javiereguiluz:
👍 I like this change a lot! Thanks Wouter.

xabbuh:
👍

javiereguiluz:
It's merged now! Thanks.

PR #355, opened 2016-06-21
Fixed other localizeddate usages to work with the Intl component

WouterJ:
Seems like these were missed in https://github.com/symfony/symfony-demo/pull/344


javiereguiluz:
A very nice (and important) catch! Thanks @WouterJ

PR #358, opened 2016-07-09
Update doctrine.dbal.path in config.yml

charlycoste:
Symfony demo doesn't seem to work out of the box when using the `prod` configuration because the path used by the `prod` and the `dev` configurations are not the same.

    %kernel.root_dir%/../var/data/blog.sqlite

vs

    %kernel.root_dir%/data/blog.sqlite

Even if it would be better to separate the both two databases in the real life, here is the patch I use to make it work the same way on a local `dev` mode and on a remote shared hosting.


javiereguiluz:
@charlycoste thanks for fixing this bug. I can confirm that the proposed changes fix it. Thank you!

PR #359, opened 2016-07-09
Move PHPUnit tests folder into the root folder

bocharsky-bw:


javiereguiluz:
👍

javiereguiluz:
I agree with this change, but I'd like to read other opinions. Apparently, Symfony doesn't recommend this practice officially.

bocharsky-bw:
I'm wondering to hear other opinions too. BTW, `symfony-standard` already moved `DefaultControllerTest` into the `tests` directory:
https://github.com/symfony/symfony-standard/blob/master/tests/AppBundle/Controller/DefaultControllerTest.php

voronkovich:
I think that holding the folder with tests in the root directory just makes it easier to launch the tests: `phpunit` vs. `phpunit -c app`. :+1:

javiereguiluz:
@voronkovich good point ... but if I'm right, that's achieved by moving the `phpunit.xml.dist` config file to the root directory. That's what Symfony 3.x did and we can run the tests with `phpunit` even if they are located inside `src/`

voronkovich:
@javiereguiluz,  you're right, I mistakenly confused the test directory and phpunit.xml file. So, I don't have any opinion about this PR :(

voronkovich:
@javiereguiluz, there is an explanation why tests folder was moved to the root directory in the symfony-standard. See https://github.com/symfony/symfony-standard/commit/d0311d2eb06309d9fc0b8bacb8cab9e0f29c622f#commitcomment-15653984

xabbuh:
The namespace of the `SluggerTest` class needs to be updated too.

bocharsky-bw:
@xabbuh Yes, I know. It's weird, but it already has correct namespace in [master](https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Tests/Utils/SluggerTest.php#L12).

xabbuh:
@bocharsky-bw But the namespace should be `Tests\AppBundle\Utils` and the location should be updated accordingly.

xabbuh:
And to be consistent with the Symfony Standard Edition we would need to have a `tests/AppBundle` directory.

bocharsky-bw:
@xabbuh Ah, you're totally right, I missed `AppBundle`.

Thanks for notice it! I'll fix it soon.

bocharsky-bw:
Done.

xabbuh:
👍

javiereguiluz:
Thank you @bocharsky-bw.

PR #360, opened 2016-07-09
Clarify patch PHP version in the requirements section

bocharsky-bw:


javiereguiluz:
@bocharsky-bw thanks for this improvement. In 1e2e9e1 I've removed the backticks around version numbers because I think they are not needed.

bocharsky-bw:
@javiereguiluz Fair enough, thanks!

README.md
```diff
@@ -48,7 +48,7 @@ most recent version executing the `symfony self-update` command.
 Usage
 -----
 
-If you have PHP 5.4 or higher, there is no need to configure a virtual host
+If you have PHP `5.4` or higher, there is no need to configure a virtual host
```
xabbuh: We can reword this now that we are on Symfony 3. People will actually already need to have PHP 5.5.9 or higher (as you changed above).

bocharsky-bw: Yes, that's a good idea. I thought about:
> Since PHP 5.4, there is no need to configure a virtual host...

What do you think?

voronkovich: I would write:
```
Since PHP 5.4, there is no need to use an HTTP-server like Apache or Nginx and configure a virtual host for it...
```

xabbuh: I would simply omit the first part of the sentence and start it with "There is […]".

bocharsky-bw: I did in #366

PR #361, opened 2016-07-09
Tweak note about fully-featured web server

bocharsky-bw:
We don't need PHP `5.3` note anymore, Symfony Demo already requires `5.5.9`.


javiereguiluz:
Nice catch! Removing PHP 5.3 support will save us a lot of troubles :)

PR #362, opened 2016-07-09
Tweak Symfony blog URL

bocharsky-bw:
Use `https` schema and add trailing slash for the Symfony blog URL to prevent extra redirection.


javiereguiluz:
Thanks @bocharsky-bw

PR #363, opened 2016-07-09
Fill default username and password only when it's empty

bocharsky-bw:
We lose `Security::LAST_USERNAME` on authorization failure.


javiereguiluz:
👍

javiereguiluz:
@bocharsky-bw in 1a5030e I've added a help note explaining why we do this :)

bocharsky-bw:
👍 Good idea!

PR #364, opened 2016-07-09
Add available locales to the admin path in access control

bocharsky-bw:
As I realized, only `@Security` annotation works in admin `BlogController` for now. Actually, we don't have `/admin` URL, all admin pages are prefixed with a locale.

### DIFF
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -42,4 +42,4 @@ security:
     access_control:
         # this is a catch-all for the admin area
         # additional security lives in the controllers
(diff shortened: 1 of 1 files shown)


javiereguiluz:
👍

javiereguiluz:
@bocharsky-bw thanks for this fix. This is a very nice catch because the current config is really wrong. Luckily we use `@Security` in the admin controller. Otherwise, we wouldn't have any security protection!

PR #365, opened 2016-07-10
Use locked PHPUnit from dependencies

bocharsky-bw:
Why do we not specify PHPUnit version in dependencies? We could use locked PHPUnit version in both Travis CI and Appveyor CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,4 @@ install:
   - composer install
 
 script:
-  - phpunit
+  - ./bin/phpunit
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,8 +23,7 @@ install:
   - cd %APPVEYOR_BUILD_FOLDER%
   - php -r "readfile('http://getcomposer.org/installer');" | php
   - php composer.phar install --no-interaction --no-progress
-  - appveyor DownloadFile https://phar.phpunit.de/phpunit.phar
 
 test_script:
   - cd %APPVEYOR_BUILD_FOLDER%
-  - php phpunit.phar --verbose
+  - php ./bin/phpunit --verbose
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "white-october/pagerfanta-bundle"      : "^1.0"
     },
     "require-dev": {
-        "sensio/generator-bundle"              : "~3.0",
+        "phpunit/phpunit"                      : "^4.8",
+        "sensio/generator-bundle"              : "^3.0",
         "symfony/phpunit-bridge"               : "^3.0"
     },
     "scripts": {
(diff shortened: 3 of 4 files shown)


javiereguiluz:
I'd like to ask the reasons behind this change. Actually I thought it was a bad practice to do this. Thanks.

bocharsky-bw:
I think, locking phpunit version may prevent potential bugs with using different version by users or in continuous integration systems. We will be sure that user uses the same version as we do. As I know, CI systems could use not the latest predefined version.

I'm also wondering to hear another opinions on this.

javiereguiluz:
I'm still not sure about this change ... so I'll wait until we receive more comments.

stof:
The issue is that different PHP versions need different PHPUnit: PHPUnit 4.8 is not officially supporting PHP 7 (and even less PHP 7.1) and PHPUnit 5 does not support PHP 5 anymore

bocharsky-bw:
@stof So what is your point here? Does PHPUnit as a Composer dependency is a bad idea for Symfony Demo?

stof:
given our supported PHP version, I would say yes

andrerom:
> given our supported PHP version, I would say yes

but if it was set to `"phpunit/phpunit"                      : "^4.8 | ^5.0",` that won't be a problem and the project will have some control over which phpunit is used. also testing for contributors becomes simpler as you won't have to make sure you have the right phpunit version.

xabbuh:
@andrerom But that would still require some additional work by the users as the lock file we ship with the repo can only refer to one of the versions.

andrerom:
> @andrerom But that would still require some additional work by the users as the lock file we ship with the repo can only refer to one of the versions.

So? That is a design flaw in composer.lock design affecting *all* packages that think they do anyone a favour by eagerly throwing older php versions under the bus. Guzzle comes to mind, but that is just one out of several. But bottom line being I don't think that argument is relevant.

javiereguiluz:
@bocharsky-bw I'm finishing this PR in #375. I've reused your commit, so you'll get full credit for your work. I also applied @andrerom ideas to allow using PHPUnit 4 or 5.

@xabbuh there should be no problem with our composer.lock because we use `5.5.9` in the `config.platform` option, so we'll always distribute a Symfony Demo app compatible with our minimum PHP requirement.

xabbuh:
@javiereguiluz Yeah, but this means that the project's PHPUnit version is locked to a version that doesn't fully support PHP 7.

bocharsky-bw:
@xabbuh Actually, all dependencies are locked to a version that doesn't fully support PHP 7 due to `config.platform: 5.5.9` option, right? So it's not the only PHPUnit problem.

stof:
@bocharsky-vw lots of packages have a release supporting both PHP 5.5.9 and PHP 7 at the same time, an dthere is no issue for them

PR #367, opened 2016-07-16
Accessibility fixes

mik-laj:
Hello,

Today I wanted to learn this framework, and I discovered a few problems with the availability of this application for people with disabilities.
First problem, fonts awesome icons missing `aria-hidden` attribute. 
You can read more about it on page: http://fontawesome.io/accessibility/
Second problem,  missing `scope` attribute on table headers. 
You  can read more about in on page: http://webaim.org/techniques/tables/data

This application is a role model for other applications, and therefore accessibility for disabled persons should be important.

Thanks in advnace!

### DIFF
--- a/app/Resources/TwigBundle/views/Exception/error403.html.twig
+++ b/app/Resources/TwigBundle/views/Exception/error403.html.twig
@@ -13,7 +13,7 @@
 {% block body_id 'error' %}
 
 {% block main %}
-    <h1 class="text-danger"><i class="fa fa-unlock-alt"></i> {{ 'http_error.name'|trans({ '%status_code%': 403 }) }}</h1>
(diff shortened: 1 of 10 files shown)


mik-laj:
This app also have a other problems:
1. Accessibility text not translated
https://github.com/symfony/symfony-demo/blob/81797b9/app/Resources/views/base.html.twig#L45
2. Missing text description for the text. 
https://github.com/symfony/symfony-demo/blob/81797b9/app/Resources/views/base.html.twig#L79
Example:
````<span class="sr-only">{{ 'choose.language' | trans }}</span>````
Class `sr-only` is a Bootstrap class. You can read more about in on: http://getbootstrap.com/components/#callout-glyphicons-accessibility

javiereguiluz:
@mik-laj thank you very much for improving the accessibility of this application. You are right about this application being a "role model" and we should encourage other developers to care more about accessibility. I'll implement the other two issues that you mentioned in your comment in a separate pull request. Thanks!

javiereguiluz:
@mik-laj the two issues that you reported are now fixed in #368. Thanks!

PR #373, opened 2016-08-25
Formation theodo

fonfonx:


stof:
@fonfonx can you avoid using the symfony repositories for your trainings ? This is sending notification emails to the symfony core team. Please use a theodo fork as target of the PR instead.

PR #375, opened 2016-08-30
Use locked PHPUnit from dependencies

javiereguiluz:
This finishes and rebases #365.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,4 +26,4 @@ install:
 
 test_script:
   - cd %APPVEYOR_BUILD_FOLDER%
-  - php ./bin/phpunit --verbose
+  - php ./vendor/bin/phpunit --verbose
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,8 +23,7 @@ install:
   - cd %APPVEYOR_BUILD_FOLDER%
   - php -r "readfile('http://getcomposer.org/installer');" | php
   - php composer.phar install --no-interaction --no-progress
-  - appveyor DownloadFile https://phar.phpunit.de/phpunit.phar
 
 test_script:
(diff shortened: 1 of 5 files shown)


bocharsky-bw:
👍

composer.json
```diff
@@ -31,7 +31,8 @@
         "white-october/pagerfanta-bundle"      : "^1.0"
     },
     "require-dev": {
-        "sensio/generator-bundle"              : "~3.0",
+        "phpunit/phpunit"                      : "^4.8 | ^5.0",
```
stof: the Composer recommendation is to use ``||`` for the OR operator rather than ``|``, because it is the operator used by other semver-based dependency managers

appveyor.yml
```diff
@@ -23,8 +23,7 @@ install:
   - cd %APPVEYOR_BUILD_FOLDER%
   - php -r "readfile('http://getcomposer.org/installer');" | php
   - php composer.phar install --no-interaction --no-progress
-  - appveyor DownloadFile https://phar.phpunit.de/phpunit.phar
 
 test_script:
   - cd %APPVEYOR_BUILD_FOLDER%
-  - php phpunit.phar --verbose
+  - php ./bin/phpunit --verbose
```
stof: I suggest moving the composer bins back to ``vendor/bin`` to be consistent with symfony-standard-edition and avoid conflicts with the project-specific bin which are also in the bin folder (making exclude rules much easier to write)

bocharsky-bw: This's a good idea 👍

bocharsky-bw: @javiereguiluz you missed tweaking this path to the new `./vendor/bin/phpunit`

javiereguiluz:
@bocharsky-bw it's merged now. Thanks for proposing this improvement.

PR #377, opened 2016-09-02
Fix Deploying to the Heroku

### LINKED ISSUE #371: Deploy to heroku is broken.
Hello,

1. I opened readme and click deploy button;
2. I filled form.
3. I saw error in log. 

Log from heroku:
http://pastebin.com/rFbSNUqv
I want to highlight a line:

> An exception occurred while executing 'CREATE TABLE symfony_demo_user (id I  [39;49m
>  
> [37;41m  NTEGER NOT NULL, username VARCHAR(255) NOT NULL, email VARCHAR(255) NOT NUL  [39;49m
>  
> [37;41m  L, password VARCHAR(255) NOT NULL, roles CLOB NOT NULL, PRIMARY KEY(id))':   [39;49m
>  
> [37;41m  SQLSTATE[HY000]: General error: 1 table symfony_demo_user already exists  

Thanks in advance

bocharsky-bw:
Fix #371 .

Let's use SQLite DB with pre-populated data instead of bootstrapping a new PostgreSQL one:

It works for me: https://symfony-demo.herokuapp.com

### DIFF
--- a/app.json
+++ b/app.json
@@ -9,9 +9,6 @@
     "repository": "https://github.com/symfony/symfony-demo",
     "logo": "https://symfony.com/images/v5/pictos/demoapp.svg?v=4",
     "success_url": "/",
-    "scripts": {
-        "postdeploy": "php bin/console doctrine:schema:create && php bin/console doctrine:fixtures:load -n"
-    },
     "env": {
         "SYMFONY_ENV": "prod",
         "SYMFONY_SECRET": {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@bocharsky-bw thanks for investigating this issue and providing a fix. Let's see if it works for others too. Thanks.

dzuelke:
Let's please reopen and revert this. SQLite is not a proper solution on a container based production system like Heroku, as the local file system is reset between deploys or on periodic restarts, and obviously also not shared between instances, so if you scale to more than one web instance, they won't share the same database.

Issue #223, opened 2015-10-12
Paginator for backend


ccab:
I'd like to add the paginator also in the backend and btw sorting the columns to improve the user experience

bocharsky-bw:
:+1: for paginator since we already have it in the frontend

voronkovich:
:+1:

stof:
I'm :+1: on it. Showing unpaginated lists is a bad idea IMO.

@javiereguiluz does it need to be implemented here or in your admin bundle though ?

javiereguiluz:
I'm :-1: because the demo application is not something you use in the real world ... but something you use to learn how real-world-looking apps work. If we already have a pagination example in the frontend, that's enough to show the newcomer how it's done.

@stof I don't fully understand your comment because this demo app doesn't use any bundle for the admin backend.

stof:
oh, I thought we were using your EasyAdminBundle. I should have checked

voronkovich:
@javiereguiluz, but @ccab wants to add a sorting and etc., no only the pagination

javiereguiluz:
@voronkovich in that case I also think it's a bad idea. The "good practice" here should be: if you need sorting, pagintion and all those common features ... use Sonata or any other backend bundle.

ccab:
Ok so let's keep it in the frontend

javiereguiluz:
Closing it as "won't fix" for the mentioned reasons.

PR #381, opened 2016-09-28
Ensure default Post sort order shows newer Posts first

simensen:
In at least #223 the topic of sorting on the **admin backend** was mentioned. This is the first time I've loaded the demo app in quite awhile and wasn't sure it was working because I didn't realize the new post I had added was going to show up at the bottom of the list. It may not be a problem for everyone but the bottom of the list on my laptop was well past the fold.

I appreciate that we do not need to make this **admin backend** fully completed with pagination and column-by-column sorting and what-not. However, I think it would make sense to have newer posts showing up at the top.

### DIFF
--- a/src/AppBundle/Controller/Admin/BlogController.php
+++ b/src/AppBundle/Controller/Admin/BlogController.php
@@ -54,7 +54,7 @@ class BlogController extends Controller
     public function indexAction()
     {
         $entityManager = $this->getDoctrine()->getManager();
-        $posts = $entityManager->getRepository(Post::class)->findAll();
+        $posts = $entityManager->getRepository(Post::class)->findBy([], ['publishedAt' => 'DESC']);
 
         return $this->render('admin/blog/index.html.twig', ['posts' => $posts]);
     }


src/AppBundle/Repository/PostRepository.php
```diff
@@ -28,6 +28,14 @@
 class PostRepository extends EntityRepository
 {
     /**
+     * @return Post[]
+     */
+    public function findAll()
+    {
+        return $this->findBy(array(), array('publishedAt' => 'DESC'));
```
voronkovich: @simensen, you should use the short array syntax: `[]`.

src/AppBundle/Repository/PostRepository.php
```diff
@@ -28,6 +28,14 @@
 class PostRepository extends EntityRepository
 {
     /**
+     * @return Post[]
+     */
+    public function findAll()
```
bocharsky-bw: Probably we should use a bit different method name like `findAllOrdered()` here instead of `findAll()` as we change its normal behavior of from `EntityRepository`. What do you think?

voronkovich: What about to call the `findBy` method instead of the `findAll` in the controller and don't change the `findAll` behavior at all?

simensen: I'm fine either way and had considered these options. I figured it would be easier to submit a PR and start that discussion if someone objected to overriding `findAll`. :)

voronkovich: @bocharsky-bw,  for now we use the `findAll` only once in the admin controller.
https://github.com/symfony/symfony-demo/search?utf8=%E2%9C%93&q=findAll

stof: Overriding the builtin doctrine methods to change their behavior is a bad practice IMO, so you should use a different name

simensen: Alright, I changed to using `findBy` in the controller.

bocharsky-bw:
I think that's the simplest solution 👍

javiereguiluz:
@simensen thanks! I like this feature and the solution was very simlpe at the end.

PR #383, opened 2016-10-04
Update sl translation

peterkokot:
This patch updates Slovenian (sl_SI) translation according to the English.

Thanks for considering merging it.


javiereguiluz:
@peterkokot thanks for always keeping this translation in sync!

PR #384, opened 2016-10-06
Use .xlf extension instead of .xliff one for translations

bocharsky-bw:
Symfony uses `.xlf` extension in its [XliffFileDumper](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Translation/Dumper/XliffFileDumper.php#L53-L56) instead of `.xliff` which currently uses by Symfony Demo. BTW, the `.xlf` is correct filename extension according to the [wiki](https://en.wikipedia.org/wiki/XLIFF) page about XLIFF format.

What about to change translations' extension for consistency?


javiereguiluz:
@bocharsky-bw it totally makes sense! Thanks for improving this.

PR #386, opened 2016-10-10
Minor performance optimizations

javiereguiluz:
* The purpose of the Symfony Demo is not to build a fast Symfony app, but to showcase how to develop Symfony apps. However, we should optimize the things that make sense and not complicate the application much.
* We're not sending emails, so we can stop loading the Swift bundle.
* We're promoting Assetic less and less ... so we could stop loading Assetic bundle too. Sadly we can't remove Assetic entirely yet because there is no clear alternative (maybe stop using SCSS and just use some simple CSS and JS files?).


bocharsky-bw:
My only argument to left Swiftmailer is to be synchronized with Symfony SE, but I think these changes make sense 👍

PR #387, opened 2016-10-10
Stop using Assetic to manage assets

javiereguiluz:
Related to #386 discussion ... what if we remove Assetic entirely?


stof:
Fine with me, but we should not force contributors to work on the minified CSS file, so we should replace it with another setup for the assets

javiereguiluz:
The diff is now impossible to review, but this is what we've done:

* Removed the source SCSS files and turn them into simple CSS files.
* Link to `*.min.css` and `*.min.js` files for third-party libraries (Bootstrap, jQuery, etc.)
* Link to `main.css` and `main.js` for our app files, so anybody can easily work with them.

PR #389, opened 2016-10-13
Made the slugger compatible with every language

javiereguiluz:
In the Symfony Installer we recently made a change to forbid any non-Latin character in the project names ([see details](https://github.com/symfony/symfony-installer/pull/278)). This is needed because Composer itself requires that for package names.

In the Symfony Demo app we had a similar problem with the slugs of the blog posts. We removed any non-Latin character ... but that caused bugs like #136. I propose to change the slugger completely and accept any character.

For example, you can now create this blog post:

backend_post

And the blog listing no longer displays an error:

blog_post

And the URL is perfectly fine although it contains non-Latin chars:

url-post


src/AppBundle/Utils/Slugger.php
```diff
@@ -27,6 +27,6 @@ class Slugger
      */
     public function slugify($string)
     {
-        return trim(preg_replace('/[^a-z0-9]+/', '-', strtolower(strip_tags($string))), '-');
+        return str_replace(' ', '-', strtolower(trim(strip_tags($string))));
```
stof: you should replace any whitespace char, not only spaces. And you should not put several dashes for consecutive spaces.

stof: btw, if you want to support any language, you should use ``mb_strtolower``

bocharsky-bw: Will `/\s+/` regex pattern be enough here?

javiereguiluz: @bocharsky-bw yes, I did that at the end 👍

javiereguiluz:
There are some failing tests only for PHP 5.5. I don't understand why because we use Symfony's polyfill for mb_strtolower():

```
There were 2 failures:
1) Tests\AppBundle\Utils\SluggerTest::testSlugify with data set #6 ('lorem ру́сский язы... ipsum', 'lorem-ру́сский-язы...-ipsum')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'lorem-ру́сский-язы́к-ipsum'
+'lorem-����������������-����������-ipsum'
/home/travis/build/symfony/symfony-demo/tests/AppBundle/Utils/SluggerTest.php:36

2) Tests\AppBundle\Utils\SluggerTest::testSlugify with data set #7 ('lorem العَرَبِيَّ�... ipsum', 'lorem-العَرَبِيَّ�...-ipsum')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'lorem-العَرَبِيَّة‎‎-ipsum'
+'lorem-������������������������‎‎-ipsum'
/home/travis/build/symfony/symfony-demo/tests/AppBundle/Utils/SluggerTest.php:36
```

voronkovich:
@javiereguiluz, Could we use the slugs without changing their cases?

voronkovich:
@javiereguiluz, I think it doesn't work because you use unsupported symbols like `у́` (it's not Cyrillic). See https://github.com/symfony/polyfill-mbstring/blob/master/Resources/unidata/lowerCase.php

javiereguiluz:
@voronkovich I took that from the Wikipedia 😱  Is it wrong?

wikipedia_russian

src/AppBundle/Utils/Slugger.php
```diff
@@ -27,6 +27,6 @@ class Slugger
      */
     public function slugify($string)
     {
-        return trim(preg_replace('/[^a-z0-9]+/', '-', strtolower(strip_tags($string))), '-');
+        return preg_replace('/\s+/', '-', mb_strtolower(trim(strip_tags($string))));
```
stof: you should force the encoding to UTF-8 for mb_strtolower. The internal encoding of mbstring might be configured to a different one

stof: note that this is very likely, as the default one was *not* UTF-8 on PHP 5 (PHP 7 changed the default encoding to UTF-8)

voronkovich:
@javiereguiluz, it's not wrong but those symbols are used rarely.  Try to use: `русский язык`.

tests/AppBundle/Utils/SluggerTest.php
```diff
@@ -44,7 +44,7 @@ public function getSlugs()
         yield ['!Lorem Ipsum!', '!lorem-ipsum!'];
         yield ['lorem-ipsum', 'lorem-ipsum'];
         yield ['lorem 日本語 ipsum', 'lorem-日本語-ipsum'];
-        yield ['lorem ру́сский язы́к ipsum', 'lorem-ру́сский-язы́к-ipsum'];
+        yield ['lorem русский языќк ipsum', 'lorem-русский язык-ipsum'];
```
voronkovich: @javiereguiluz,  you forgot to delete odd ќ

javiereguiluz:
And ot's working 🎉  Thank you all for your help!

PR #390, opened 2016-10-16
Remove obsolete code

fabpot:
Now that Doctrine DBAL 2.5.5 is out, I think we can remove this "hack". Also because the config value is different from the one recommended and commented in the main config, which might confuse people.


xabbuh:
👍

javiereguiluz:
Thanks Fabien.

PR #392, opened 2016-10-25
Updated the ide option to use the default open in browser

jeremyFreeAgent:
Related to https://github.com/symfony/symfony/pull/19973

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -19,12 +19,12 @@ parameters:
 
 # Basic configuration for the Symfony framework features
 framework:
-    # Uncomment the 'ide' option to turn all of the file paths in an exception
+    # The 'ide' option to turn all of the file paths in an exception
     # page into clickable links that open the given file using your favorite IDE.
     # Supported values are 'textmate', 'macvim', 'emacs' and 'sublime' shortcuts
     # and any custom configuration string, such as: "phpstorm://open?file=%%f&line=%%l"
     # See http://symfony.com/doc/current/reference/configuration/framework.html#ide
-    ide: sublime
+    ide: ~
 
     # When using the HTTP Cache, ESI allows to render page fragments separately
     # and with different cache configurations for each fragment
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I like this proposal because this is the only value that will work for everyone. Thanks @jeremyFreeAgent

jeremyFreeAgent:
You should thank @nicolas-grekas because he is the one who told me there was the `ide` option set in the symfony-demo ;)

app/config/config.yml
```diff
@@ -19,12 +19,11 @@ parameters:
 
 # Basic configuration for the Symfony framework features
 framework:
-    # Uncomment the 'ide' option to turn all of the file paths in an exception
-    # page into clickable links that open the given file using your favorite IDE.
-    # Supported values are 'textmate', 'macvim', 'emacs' and 'sublime' shortcuts
-    # and any custom configuration string, such as: "phpstorm://open?file=%%f&line=%%l"
+    # The 'ide' option turns all of the file paths in an exception page
+    # into clickable links that open the given file using your favorite IDE.
+    # Supported values are 'phpstorm', 'sublime', 'textmate', 'macvim' and 'emacs'.
```
fabpot: You should probably explain the null value here.

sargath:
Bare in mind that the user should have configured a link handler in his browser to open up those -
http://davidbu.ch/mann/blog/2014-05-06/open-php-stack-trace-links-phpstorm.html

Not every OS can handle this by its own.

bocharsky-bw:
I'd like to promote PhpStorm here since it's cross-platform and it has a nice Symfony plugin. Obviously, it's simply the best IDE for PHP apps and especially for Symfony apps.

app/config/config.yml
jeremyFreeAgent: Done

jeremyFreeAgent: Thanks :)

javiereguiluz:
Let's merge this because Symfony 3.2 has already been released and using "null" in this option is now the default value. Thanks @jeremyFreeAgent!

PR #393, opened 2016-10-27
Updated to Symfony 3.2 BETA1

javiereguiluz:
The beta1 of Symfony 3.2 has just been released, so let's upgrade the Symfony Demo to catch bugs and deprecations before the final release.


javiereguiluz:
First deprecation notice discovered after the upgrade:

```
User Deprecated: The __TwigTemplate_7fa0af8977e9ad6022a4bc7d6829cb425210c6ac6e9a4fbe673c55e6030063ab::getSource method is deprecated since version 1.27 and will be removed in 2.0. Use getSourceContext() instead.
```

PR #394, opened 2016-11-02
Add annotation with proper types for entity properties

bocharsky-bw:
The most IDEs start showing returned types on getter calls after these changes.

### DIFF
--- a/src/AppBundle/Entity/Post.php
+++ b/src/AppBundle/Entity/Post.php
@@ -103,17 +103,11 @@ class Post
         $this->comments = new ArrayCollection();
     }
 
-    /**
-     * @return int
-     */
     public function getId()
     {
         return $this->id;
     }
 
-    /**
-     * @return string
-     */
     public function getTitle()
     {
         return $this->title;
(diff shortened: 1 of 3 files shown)


bocharsky-bw:
I also wonder about using `@var` annotation for setters and `@return` one for getters in Symfony Demo. What do you think about adding it?

javiereguiluz:
If we add the `@var` for properties, I guess it makes sense it add it to getters/setters too.

src/AppBundle/Entity/Post.php
```diff
@@ -22,56 +22,74 @@
 class Post
 {
     /**
+     * @var int
```
stof: useless for constants. The IDE already reads the value anyway. A constant is not a variable

bocharsky-bw: Reverted.
Thanks, Christophe!

xabbuh:
When the properties are annotated, you shouldn't have to repeat the type for the returned values of the getter methods as IDEs are able to infer the return types then.

bocharsky-bw:
@xabbuh Do you suggest to fully revert commit about annotations for setters / getters?

xabbuh:
@bocharsky-bw For setters where we do not have type hints (i.e. scalar type arguments) we should keep them.

PR #396, opened 2016-11-02
Remove unused namespace

bocharsky-bw:


javiereguiluz:
Thanks @bocharsky-bw.

PR #397, opened 2016-11-07
Add PHP CS fixer for this application

### LINKED ISSUE #395: Add an example .php_cs
Many projects nowadays use php-cs-fixer, and the new version is losing the ability to run `--config=symfony23`, leaving us with the task to manually craft a `.php_cs` file on our own, it would be helpful to have an example here on symfony-demo.

bocharsky-bw:
Resolve #395.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,16 @@
+#!/usr/bin/env php
+<?php
+
+$finder = Symfony\CS\Finder::create()
+    ->ignoreDotFiles(true)
+    ->ignoreVCS(true)
+    ->exclude('app')
+    ->exclude('var')
+    ->exclude('vendor')
+    ->exclude('web')
+    ->in(__DIR__)
+;
+
+return Symfony\CS\Config::create()
+    ->finder($finder)
+;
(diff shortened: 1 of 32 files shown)


javiereguiluz:
@bocharsky-bw thanks for working on this!

I have two questions:

1) Is it recommended to add PHP-CS-Fixer as a project dependency or is it better to install it as a global tool in your computer?

2) Which is the default list of fixers applied when you don't select any config? I've seen some fixers [in the official docs](https://github.com/FriendsOfPHP/PHP-CS-Fixer) that are interesting but I don't know if they are enabled. For example: `phpdoc_order`, `short_array_syntax`, `ordered_use`, etc.

bocharsky-bw:
Hey @javiereguiluz 

1) Actually, I don't know, but I did the same way as we had done with PhpUnit. I wonder others opinion about it.

2) As I see in docs:
> By default, all PSR-2 fixers and some additional ones are run.

So I specified Symfony level explicitly and added a few fixers that you had suggested.

javiereguiluz:
@bocharsky-bw very nice! Thanks. Let's wait for others opinion about adding PHP-CS-Fixer as a dependency or not.

javiereguiluz:
@bocharsky-bw you did a great work here. Thank you! I'm merging it because we can always tweak this config later to add/remove more fixers if needed.

PR #399, opened 2016-11-08
Use dynamic env params for critical settings and restore Heroku deploys

dzuelke:
Database, log destination, and secrets are now configurable via environment variables (via the quasi standard `DATABASE_URL`, and `SYMFONY_LOG`, and `SYMFONY_SECRET`). All three fall back to defaults from `parameters.yml(.dist)`.

For simplicity (one param versus many), a URL is now used again for SQLite, because that works fine since Doctrine DBAL 2.5.5.

Deploys to Heroku are now fixed again; @bocharsky-bw's PR #377 was not a fix for #371, because SQLite does not scale horizontally :p The original problem was @javiereguiluz's 56cfa66 change to Symfony 3.1, which used individual parameters for config. Again, no longer a problem since DBAL 2.5.5. It also uses `php://stderr` for env `prod` on Heroku only, which @javiereguiluz had also removed in 56cfa665cf1235c436c70afa5fc85c2668c14605 (for the future, I'd encourage separate commits for stuff like this, as it makes it much easier to untangle individual parts and reasons for changes).

This demo app ships with sample data in SQLite, but for other databases, the data needs to be seeded, and that can even happen in a prod env. @javiereguiluz' commit e2264b0 removed the fixtures bundle needed for that from the kernel init; it's now back.

This app now works just fine with the usual defaults both in `dev` and `prod` envs, but users who want to try running it on Docker or on a PaaS like Heroku can now trivially adjust relevant settings via environment variables.

/CC @stof @fabpot

### DIFF
--- a/app.json
+++ b/app.json
@@ -14,6 +14,7 @@
     },
     "env": {
         "SYMFONY_ENV": "prod",
+        "SYMFONY_LOG": "php://stderr",
         "SYMFONY_SECRET": {
             "description": "Extra entropy for %kernel.secret%; used for CSRF tokens, cookies and signed URLs.",
             "generator": "secret"
--- a/app.json
+++ b/app.json
@@ -9,6 +9,9 @@
     "repository": "https://github.com/symfony/symfony-demo",
     "logo": "https://symfony.com/images/v5/pictos/demoapp.svg?v=4",
     "success_url": "/",
+    "scripts": {
+        "postdeploy": "php bin/console doctrine:schema:create && php bin/console doctrine:fixtures:load -n"
+    },
     "env": {
         "SYMFONY_ENV": "prod",
         "SYMFONY_SECRET": {
@@ -16,5 +19,8 @@
             "generator": "secret"
         }
     },
+    "addons": [
+        "heroku-postgresql"
+    ],
     "image": "heroku/php"
 }
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -21,6 +21,7 @@ class AppKernel extends Kernel
             new WhiteOctober\PagerfantaBundle\WhiteOctoberPagerfantaBundle(),
             new CodeExplorerBundle\CodeExplorerBundle(),
             new AppBundle\AppBundle(),
+            new Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle(), // used for initial population of non-SQLite databases in production envs
             // uncomment the following line if your application sends emails
             // new Symfony\Bundle\SwiftmailerBundle\SwiftmailerBundle(),
         ];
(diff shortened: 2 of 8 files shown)


app/config/config.yml
```diff
@@ -61,10 +61,10 @@ twig:
...
 doctrine:
     dbal:
-        # if you don't want to use SQLite, comment the two following lines
-        driver: "pdo_sqlite"
-        path: "%kernel.root_dir%/data/blog.sqlite"
-        # uncomment the following lines to use a database different than SQLite
+        # if you don't want to use SQLite, change the URL in parameters.yml or set the DATABASE_URL environment variable
+        url: "%env(DATABASE_URL)%"
+        
+        # instead of using a URL, you may also uncomment the following lines to configure your database
```
stof: you also need to comment the URL (as the URL wins over these params if you configure both)

dzuelke: That's why it says "instead of" ;)

app/AppKernel.php
```diff
@@ -21,6 +21,7 @@ public function registerBundles()
             new WhiteOctober\PagerfantaBundle\WhiteOctoberPagerfantaBundle(),
             new CodeExplorerBundle\CodeExplorerBundle(),
             new AppBundle\AppBundle(),
+            new Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle(), // used for initial population of non-SQLite databases in production envs
```
fabpot: I don't think it makes sense in a "real" application, might make sense as a demo app though.

app.json
```diff
@@ -9,12 +9,19 @@
     "repository": "https://github.com/symfony/symfony-demo",
     "logo": "https://symfony.com/images/v5/pictos/demoapp.svg?v=4",
     "success_url": "/",
+    "scripts": {
+        "postdeploy": "php bin/console doctrine:schema:create && php bin/console doctrine:fixtures:load -n"
```
dzuelke: No. The name `postdeploy` for the script is, unfortunately, misleading; it is run [only after the very first deploy](https://devcenter.heroku.com/articles/app-json-schema#scripts) using the "Deploy to Heroku" button.

app/AppKernel.php
dzuelke: Yes and no. Even a real application, when first deployed somewhere, may need e.g. an initial user inserted into the database ;)

app.json
bocharsky-bw: Now I see, thanks!

dzuelke:
So... merge? ;)

javiereguiluz:
@dzuelke thanks for fixing all the issues that I introduced a while ago. Your solution is exactly what we were looking for: something that works perfectly for cloud and non-cloud users. It's merged now. Thanks!

dzuelke:
Awesome, thank you so much @javiereguiluz! Like I said, if there's ever again any problems related to Heroku deploys, or just generally Twelve Factor, Cloud, PaaS, env vars... just ping me and I can chime in!

PR #401, opened 2016-11-12
Modify default routing parameter on indexAction

hanahiroAze:
I found $page parameter's type on default routing is different from that of "/page/{page}" routing.
This difference seems not to be reasonable, then I fix it.

### DIFF
--- a/src/AppBundle/Controller/BlogController.php
+++ b/src/AppBundle/Controller/BlogController.php
@@ -34,7 +34,7 @@ use AppBundle\Form\CommentType;
 class BlogController extends Controller
 {
     /**
-     * @Route("/", defaults={"page": 1}, name="blog_index")
+     * @Route("/", defaults={"page": "1"}, name="blog_index")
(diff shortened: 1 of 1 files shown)


xabbuh:
I don't understand what you mean. Which other default value are you talking about?

hanahiroAze:
In `* @Route("/", defaults={"page": 1}, name="blog_index")`, the parameter "page" is set as int value by `{"page": 1}`, but in ` @Route("/page/{page}", requirements={"page": "[1-9]\d*"}, name="blog_index_paginated")`  the parameter "page" is set as string value by `{"page": "[1-9]\d*"}`.

When I dump `$page` parameter in `BlogController.indecAction()`, Symfony profiler says like below, and found the parameter's type is different.

Top page -> application


click "Next" button


click "Previous" button

bocharsky-bw:
I think it makes sense for PHP 7 scalar types hinting 👍

javiereguiluz:
I'm merging this because I think it makes sense to treat this routing parameter consistently. IThanks @hanahiroAze

PR #403, opened 2016-11-18
Upgraded the application to Symfony 3.2-RC1 and Twig 1.28

javiereguiluz:


composer.json
```diff
@@ -23,9 +23,9 @@
         "sensio/framework-extra-bundle"        : "^3.0",
         "symfony/monolog-bundle"               : "^2.8",
         "symfony/swiftmailer-bundle"           : "^2.3",
-        "symfony/symfony"                      : "^3.2-beta1",
+        "symfony/symfony"                      : "^3.2-rc1",
         "twig/extensions"                      : "^1.3",
-        "twig/twig"                            : "^1.27",
```
bocharsky-bw: Do we really need to bump Twig version in composer.json? I mean, do we really have strict dependency with Twig v1.28? it would be enough `composer up twig/twig` - Composer will install 1.28 and freeze it in composer.lock anyway since we have `^` in composer.json.

javiereguiluz: Yes and no. As you can see in [the CHANGELOG of 3.2-rc1](https://github.com/symfony/symfony/pull/20552): 

```
bug #20484 bumped min version of Twig to 1.28 (@fabpot)
```

When some minimum version (e.g. `1.28`) is mandatory for some dependency (e.g. `Twig`) I like to set it explicitly.

dzuelke:
Superseded by #405

javiereguiluz:
Closing it in favor of #405.

PR #405, opened 2016-11-27
Upgraded the application to Symfony 3.2.0

javiereguiluz:


javiereguiluz:
After upgrading, I see the following deprecation log in every page:

32rc2_deprecation

jzawadzki:
@javiereguiluz it looks like it's fixed in symfony/monolog-bundle v2.12 (by https://github.com/symfony/monolog-bundle/commit/a55b0a098951dcab5d902726a8b1579053e1b1d0#diff-25076ecfab002a680c691e32a36c5e6bL106) - currently composer installs v2.11.3

javiereguiluz:
@jzawadzki you are right! I've updated all dependencies and the deprecation log is gone. Thanks!

dzuelke:
:+1:

xabbuh:
:+1:

bocharsky-bw:
Let's update to v3.2.0

PR #407, opened 2016-12-09
Updated to php-cs-fixer 2.0

ABM-Dan:


.php_cs
```diff
@@ -1,29 +1,28 @@
...
+    ->setRules([
+        '@Symfony' => true,
+        '@Symfony:risky' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'ordered_imports' => true,
+        'phpdoc_order' => true,
     ])
-    ->finder($finder)
+    ->setFinder(
+        PhpCsFixer\Finder::create()
```
bocharsky-bw: Oh, please, let's use a variable to get rid of this ugly nested sheet of code which passes to the `setFinder()` - it worth really nothing, but increase readability.

ABM-Dan: I am riffing the style of https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/.php_cs.dist

.php_cs
```diff
@@ -1,29 +1,28 @@
...
-        'phpdoc_order',
-        'short_array_syntax',
+return PhpCsFixer\Config::create()
+    ->setRiskyAllowed(true)
+    ->setRules([
+        '@Symfony' => true,
+        '@Symfony:risky' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'ordered_imports' => true,
+        'phpdoc_order' => true,
```
ABM-Dan: should probably add header for 

    /*
     * This file is part of the Symfony package.
     *
     * (c) Fabien Potencier <fabien@symfony.com>
     *
     * For the full copyright and license information, please view the LICENSE
     * file that was distributed with this source code.
     */

keradus: Probably not.
When one want to make project of his own using symfony skeleton he does not want to give sb else the copyright

.php_cs
keradus: IIRC exclude could take an array

ABM-Dan: Checking https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Finder.php I might get away with less config.

ABM-Dan: How about https://gist.github.com/d65c0a1ac759d3c74162a7145e5d91e6 ?

javiereguiluz: I don't have a strong opinion about this ... but I'm slightly in favor of @bocharsky-bw comment. It's true that PHP-CS-Fixer uses a nested finder ... but their finder is minimal and ours is not (even after the last optimization shared by @ABM-Dan via Gist.

.php_cs
ABM-Dan: I only suggested it because that copyright is already present in all files, the idea is not for me to inject some new content, just enforce any updates and checks for consistent messaging.

I do understand that we might not want to have this, but if we don't why do we even have that header in all files?

.php_cs
ABM-Dan: At this point the difference is almost academic, I'm ok either way.

ABM-Dan: Is this too extreme? https://gist.github.com/ABM-Dan/25375022321bbf779ef5e1ccd4eb62c7

javiereguiluz:
@ABM-Dan just asking: will you be able to finish this PR in the coming days? I'd love to have it merged. Thanks.

xabbuh:
I think this PR should be set on hold until the issues we have with the CS fixer in the code repository have been fixed.

javiereguiluz:
This PR stalled, so I'm finishing it in #454. @ABM-Dan I reused your original commits, so you'll get full credits for your work. Thanks!

PR #408, opened 2016-12-09
Fix annotations code style issues for entities

voronkovich:


src/AppBundle/Entity/Post.php
```diff
@@ -95,7 +95,7 @@ class Post
      *      mappedBy="post",
      *      orphanRemoval=true
      * )
-     * @ORM\OrderBy({"publishedAt" = "DESC"})
+     * @ORM\OrderBy({"publishedAt": "DESC"})
```
bocharsky-bw: Good spot! I think `:` is more native since it looks like a JSON object here, so I like it `:` more. But if both works, I'd rather send a PR to Doctrine docs to use `:` instead of `=` and see what they'll say :)

stof: recent versions of doctrine/annotations allow both ``=`` and ``:`` as delimiter between the array key and value. Old versions were allowing only ``=``, which is why the ORM doc uses it.

voronkovich: @javiereguiluz, we already use the `:` in routes: https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Controller/BlogController.php#L38

frastel: The new syntax could confuse Newbies while they are writing their first Symfony app. The annotations from the framework bundle are still using `=` and they might wonder, why the annotations in this case look complete different. They might not recognize the difference of the two seperate projects.

javiereguiluz: I'm sorry. I wasn't aware of this Doctrine feature and I thought this was a bug. Let's keep it then.

stof: @frastel FrameworkBundle does not have annotations. It uses doctrine/annotations.

The difference you see corresponds to 2 different features of the annotation syntax.
Putting a value in a property of the annotation and writing an associative array in a value are already 2 different things.
what could confuse you is that the property name is omitted here (thanks to a feature allowing to do it in some cases). The complete syntax is ``@ORM\OrderBy(value={"publishedAt": "DESC"})``

frastel: @stof You are totally right! I think I didn't finished my thoughts :D. After thinking 3 times about the fact I think `:` is much cleaner. 
I never thought about that the shortcut syntax misses `value=` actually.

And yeah, I meant the FrameworkExtraBundle, not the FrameworkBundle.

javiereguiluz:
Thanks @voronkovich.

PR #410, opened 2016-12-15
Fixed tests on Windows systems

javiereguiluz:
Trying to fix the AppVeyor issues by reusing Symfony's `appveyor.yml` file.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,29 +1,62 @@
 build: false
-platform: x86
+clone_depth: 1
 clone_folder: c:\projects\symfony-demo
 
 cache:
-  - '%LOCALAPPDATA%\Composer\files'
+    - c:\projects\symfony\composer.phar
+    - .phpunit -> phpunit
 
 init:
-  - SET PATH=C:\Program Files\OpenSSL;c:\tools\php;%PATH%
+    - SET PATH=c:\php;%PATH%
+    - SET COMPOSER_NO_INTERACTION=1
+    - SET SYMFONY_DEPRECATIONS_HELPER=strict
+    - SET ANSICON=121x90 (121x90)
+    - SET SYMFONY_PHPUNIT_SKIPPED_TESTS=phpunit.skipped
+    - REG ADD "HKEY_CURRENT_USER\Software\Microsoft\Command Processor" /v DelayedExpansion /t REG_DWORD /d 1 /f
 
 install:
-  - cinst -y OpenSSL.Light
-  - cinst -y php
-  - cd c:\tools\php
-  - copy php.ini-production php.ini /Y
-  - echo date.timezone="UTC" >> php.ini
-  - echo extension_dir=ext >> php.ini
-  - echo extension=php_openssl.dll >> php.ini
-  - echo extension=php_mbstring.dll >> php.ini
-  - echo extension=php_intl.dll >> php.ini
-  - echo extension=php_pdo_sqlite.dll >> php.ini
-  - echo memory_limit=1G >> php.ini
-  - cd %APPVEYOR_BUILD_FOLDER%
-  - php -r "readfile('http://getcomposer.org/installer');" | php
-  - php composer.phar install --no-interaction --no-progress
+    - mkdir c:\php && cd c:\php
+    - appveyor DownloadFile https://raw.githubusercontent.com/symfony/binary-utils/master/cacert.pem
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -1,29 +1,59 @@
 build: false
-platform: x86
+clone_depth: 1
 clone_folder: c:\projects\symfony-demo
 
 cache:
-  - '%LOCALAPPDATA%\Composer\files'
+    - c:\projects\symfony\composer.phar
```
stof: path should be adapter

appveyor.yml
```diff
@@ -1,29 +1,59 @@
...
+    - echo opcache.enable_cli=1 >> php.ini-max
+    - echo extension=php_openssl.dll >> php.ini-max
+    - echo extension=php_apcu.dll >> php.ini-max
+    - echo apc.enable_cli=1 >> php.ini-max
+    - echo extension=php_intl.dll >> php.ini-max
+    - echo extension=php_mbstring.dll >> php.ini-max
+    - echo extension=php_fileinfo.dll >> php.ini-max
+    - echo extension=php_pdo_sqlite.dll >> php.ini-max
+    - echo extension=php_curl.dll >> php.ini-max
+    - echo curl.cainfo=c:\php\cacert.pem >> php.ini-max
```
stof: you don't need the distinction between ini-max and ini-min for the demo

appveyor.yml
```diff
@@ -1,29 +1,59 @@
...
+    - IF %APPVEYOR_REPO_BRANCH%==master (SET COMPOSER_ROOT_VERSION=dev-master) ELSE (SET COMPOSER_ROOT_VERSION=%APPVEYOR_REPO_BRANCH%.x-dev)
+    - php composer.phar update --no-progress --ansi
+    - SET COMPOSER_ROOT_VERSION=
 
 test_script:
-  - cd %APPVEYOR_BUILD_FOLDER%
-  - php ./vendor/bin/phpunit --verbose
+    - cd c:\projects\symfony-demo
+    - SET X=0
+    - copy /Y c:\php\php.ini-min c:\php\php.ini
```
stof: no need to test twice with different php.ini

appveyor.yml
```diff
@@ -1,29 +1,53 @@
 build: false
-platform: x86
+clone_depth: 1
 clone_folder: c:\projects\symfony-demo
 
 cache:
-  - '%LOCALAPPDATA%\Composer\files'
```
stof: keeping the composer cache would still be a good idea though (Symfony does not do it due to the magic it does regarding subtree packages built on the fly, but you can do it)

appveyor.yml
```diff
@@ -1,29 +1,53 @@
...
+    - c:\projects\symfony-demo\composer.phar
+    - .phpunit -> phpunit
```
stof: this one is useless too, as you don't have such special phpunit install

appveyor.yml
```diff
@@ -1,29 +1,53 @@
...
+    - php composer.phar update --no-progress --ansi
+    - SET COMPOSER_ROOT_VERSION=
 
 test_script:
-  - cd %APPVEYOR_BUILD_FOLDER%
-  - php ./vendor/bin/phpunit --verbose
+    - cd c:\projects\symfony-demo
+    - SET X=0
+    - php vendor/bin/phpunit || SET X=!errorlevel!
+    - exit %X%
```
stof: you can remove all the magic regarding ``X`` here, as you don't need to run several test commands

javiereguiluz:
@stof thanks for your detailed review! This file is getting really polished 😁

appveyor.yml
```diff
@@ -1,29 +1,51 @@
...
+    - php composer.phar self-update
+    - IF %APPVEYOR_REPO_BRANCH%==master (SET COMPOSER_ROOT_VERSION=dev-master) ELSE (SET COMPOSER_ROOT_VERSION=%APPVEYOR_REPO_BRANCH%.x-dev)
+    - php composer.phar update --no-progress --ansi
+    - SET COMPOSER_ROOT_VERSION=
 
 test_script:
-  - cd %APPVEYOR_BUILD_FOLDER%
-  - php ./vendor/bin/phpunit --verbose
+    - cd c:\projects\symfony-demo
+    - php vendor/bin/phpunit || SET X=!errorlevel!
```
stof: you forgot to remove the ``|| SET X=!errorlevel!`` part (which is a good way to ensure that the build stay green 😄 as you turn failures into success currently)

PR #412, opened 2016-12-20
Move SQLite DB to the var/ directory

bocharsky-bw:
According to the changes in Symfony SE: https://github.com/symfony/symfony-standard/pull/1032


javiereguiluz:
👍  It totally makes sense.

stof:
it should be excluded from the gitignore rules though to be consistent with the previous setup. Currently, ``var/*`` is ignored (with some whitelist in it)

bocharsky-bw:
@stof What about to add this to `.gitignore`:
```
!/var/data
/var/data/*
!var/data/.gitkeep
```

stof:
Well, we don't want to ignore the ``var/data/blog.sqlite`` file AFAIK

bocharsky-bw:
But we also don't want to encourage other devs to commit their DB into the repository. We added `blog.sqlite` to the repository only for our demo purposes.

stof:
@bocharsky-bw this repo is the demo, not the starting point for new projects (which is what symfony-standard is). So the rules should match the needs of the demo.
A real project should probably not use sqlite for its main storage anyway, as this is incompatible with any deployment involving multiple servers (as they would each have a separate DB) and with any PaaS (as they replace the server when deploying, going back to the point about multiple server, plus they often advocate redundancy so having multiple servers too)

bocharsky-bw:
Thanks @stof, it makes sense. I fixed it in a3bdaf1

stof:
👍

javiereguiluz:
Thanks @bocharsky-bw.

PR #413, opened 2016-12-28
Tweak composer.json according to the latest changes in Symfony SE

bocharsky-bw:
According to the https://github.com/symfony/symfony-standard/blob/3.2/composer.json


javiereguiluz:
Thank you @bocharsky-bw.

PR #416, opened 2016-12-30
Sort new Composer packages automatically

bocharsky-bw:
Related to the https://github.com/symfony/symfony-standard/pull/1036

However, I think it makes even more sense here since we have already sorted packages in composer.json.


javiereguiluz:
@bocharsky-bw thank you for this improvement!

PR #417, opened 2016-12-31
Fix deprecated swiftmailer parameter

bocharsky-bw:
According to the https://github.com/symfony/symfony-standard/pull/1039

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -30,4 +30,4 @@ monolog:
         #    level:  info
 
 #swiftmailer:
-#    delivery_address: me@example.com
+#    delivery_addresses: ['me@example.com']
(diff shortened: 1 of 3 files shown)


voronkovich:
We don't use the swiftmailer at all. What about to just remove it?

javiereguiluz:
I'm divided about this. I'd like to add some feature using SwiftMailer (hopefully with something like this -> https://github.com/swiftmailer/swiftmailer/pull/845) but I'm not sure we'll add that anytime soon ... so maybe we should remove SwiftMailer altogether as suggested by @voronkovich.

javiereguiluz:
Closing it in favor of #421, which will disable the actual delivery of emails and this parameter will be removed.

PR #420, opened 2017-01-05
Removed the unused SwiftMailerBundle

javiereguiluz:
Related to #417 and created to see if the community agrees or not with this change. Thanks!

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -22,8 +22,6 @@ class AppKernel extends Kernel
             new CodeExplorerBundle\CodeExplorerBundle(),
             new AppBundle\AppBundle(),
             new Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle(), // used for initial population of non-SQLite databases in production envs
-            // uncomment the following line if your application sends emails
-            // new Symfony\Bundle\SwiftmailerBundle\SwiftmailerBundle(),
         ];
 
         // Some bundles are only used while developing the application or during
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -76,16 +76,3 @@ doctrine:
     orm:
         auto_generate_proxy_classes: "%kernel.debug%"
         auto_mapping: true
-
-# Swiftmailer Configuration (used to send emails)
-# instead of configuring the mailer credentials in this file, we pull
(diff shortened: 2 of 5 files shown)


bocharsky-bw:
There's some Swiftmailer code in `config_dev.yml` and `config_test.yml` files.

bocharsky-bw:
IMO, since it comes out-of-the-box with Symfony SE I'd rather add a simple example of its usage instead of remove it at all.

javiereguiluz:
One of the main problems is that we want the Symfony Demo fully working ... and you can't do that with SwiftMailer. We can't make SwiftMailer work and send emails for our users, because that means that they need to add some config parameters in the application with their Gmail or mail server credentials.

bocharsky-bw:
I mean using [The Mail Transport](http://swiftmailer.org/docs/sending.html#the-mail-transport), i.e. native PHP mail() function for sending emails, but as I see, it has some issues too, so probably removing Swiftmailer is the best option here 👍

javiereguiluz:
@bocharsky-bw problem is that the mail transport has been deprecated and it will be removed soon: https://github.com/swiftmailer/swiftmailer/commit/905b0bca8acf359dad3dfc2bc298c56de922cd57

voronkovich:
@javiereguiluz, we can disable email sendig in the configuration
```yaml
swiftmailer:
    disable_delivery:  true
```

javiereguiluz:
@voronkovich yes ... but actually I'm looking for the opposite. If we want to showcase the SwiftMailer feature ... then the email sending must work and developers must receive some emails somewhere.

bocharsky-bw:
@javiereguiluz Ah, then it makes even more sense... thanks!

frastel:
Seeing the Mails in the Web Debug Toolbar is not enough?

voronkovich:
@javiereguiluz, developers can view the sent email in the debug toolbar
http://symfony.com/doc/current/email/dev_environment.html#viewing-from-the-web-debug-toolbar

javiereguiluz:
@voronkovich you are right! Maybe some "fake sending" is enough to showcase this feature.

javiereguiluz:
Closing in favor of other PRs that will make use of the SwiftMailer bundle.

PR #421, opened 2017-01-05
Notify post author about new comments by sending email

voronkovich:
See https://github.com/symfony/symfony-demo/pull/420

### DIFF
--- a/app/Resources/translations/messages.en.xlf
+++ b/app/Resources/translations/messages.en.xlf
@@ -276,13 +276,13 @@
                 <source>post.deleted_successfully</source>
                 <target>Post deleted successfully!</target>
             </trans-unit>
-            <trans-unit id="post.recieved_comment">
-                <source>post.recieved_comment</source>
-                <target>Your post recieved a comment!</target>
+            <trans-unit id="post.comment_added">
+                <source>post.comment_added</source>
+                <target>Your post received a comment!</target>
             </trans-unit>
-            <trans-unit id="post.recieved_comment_message">
-                <source>post.recieved_comment_message</source>
-                <target><![CDATA[Your post "%title%" has recieved a new comment. You can read the comment by following this <a href="%link%">link</a>!]]></target>
+            <trans-unit id="post.comment_added.description">
+                <source>post.comment_added.description</source>
+                <target><![CDATA[Your post "%title%" has received a new comment. You can read the comment by following <a href="%link%">this link</a>]]></target>
             </trans-unit>
 
             <trans-unit id="help.app_description">
--- a/app/Resources/translations/messages.en.xlf
+++ b/app/Resources/translations/messages.en.xlf
@@ -276,12 +276,13 @@
                 <source>post.deleted_successfully</source>
                 <target>Post deleted successfully!</target>
             </trans-unit>
-            <trans-unit id="post.comment_added">
-                <source>post.comment_added</source>
+
+            <trans-unit id="notification.comment_created">
+                <source>notification.comment_created</source>
                 <target>Your post received a comment!</target>
             </trans-unit>
-            <trans-unit id="post.comment_added.description">
-                <source>post.comment_added.description</source>
+            <trans-unit id="notification.comment_created.description">
+                <source>notification.comment_created.description</source>
                 <target><![CDATA[Your post "%title%" has received a new comment. You can read the comment by following <a href="%link%">this link</a>]]></target>
             </trans-unit>
 
--- a/app/Resources/translations/messages.en.xlf
+++ b/app/Resources/translations/messages.en.xlf
@@ -276,6 +276,14 @@
                 <source>post.deleted_successfully</source>
                 <target>Post deleted successfully!</target>
             </trans-unit>
+            <trans-unit id="post.recieved_comment">
+                <source>post.recieved_comment</source>
+                <target>Your post recieved a comment!</target>
+            </trans-unit>
+            <trans-unit id="post.recieved_comment_message">
+                <source>post.recieved_comment_message</source>
+                <target><![CDATA[Your post "%title%" has recieved a new comment. You can read the comment by following this <a href="%link%">link</a>!]]></target>
+            </trans-unit>
 
             <trans-unit id="help.app_description">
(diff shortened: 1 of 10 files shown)


javiereguiluz:
I like this feature proposal ... but what about replacing the private method in the controller by an event listener + custom event trigger? That would allow us to show custom events in action.

voronkovich:
@javiereguiluz, I've added an event listener and custom event trigger. BTW, I found that we have an example of using mailer in the `ListUsersCommand` https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Command/ListUsersCommand.php#L131

app/config/services.yml
```diff
@@ -31,9 +31,15 @@ services:
...
             - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest }
 
-    # Event subscribers are similar to event listeners but they don't need to add
-    # a separate tag for each listened event. Instead, the PHP class of the event
-    # subscriber includes a method that returns the list of listened events.
+    app.comment_notification:
+        class: AppBundle\EventListener\CommentNotificationListener
+        arguments: ['@mailer', '@router', '@translator', '%app.notifications.email_sender%']
+        tags:
+            - { name: kernel.event_listener, event: comment.created, method: onCommentCreated }
```
javiereguiluz: I have two questions about the `onCommentCreated` method name:

1) When the method name is `on + camelCasedEventName`, the `method` parameter is optional. Should we add it anyway?
2) Some people like method names that describe the listened event (e.g. `on + EventName`) and others prefer semantic method names that describe their purposes (e.g. `notifyCommentAdded()`). Which is your preference?

src/AppBundle/EventListener/CommentNotificationListener.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * @var UrlGeneratorInterface
+     */
+    private $urlGenerator;
+
+    /**
+     * @var string
+     */
+    private $sender = 'symfony-demo@localhost';
```
bocharsky-bw: What about to make it as a parameter in `config.yml`?

javiereguiluz: Yes. It's already passed to the constructor of the service, so I'd say: let's remove this private `$sender` property and just use the container parameter.

voronkovich: @javiereguiluz, @bocharsky-bw, do you suggest to inject container into the listener to get the `sender` parameter like there https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Command/ListUsersCommand.php#L138
As you can see in the listener definition those parameter (`app.notifications.email_sender`) is already injected.

stof: no, keep the property here, but removes its default value and make it a mandatory constructor argument

app/config/services.yml
```diff
@@ -31,9 +31,15 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest }
 
-    # Event subscribers are similar to event listeners but they don't need to add
-    # a separate tag for each listened event. Instead, the PHP class of the event
-    # subscriber includes a method that returns the list of listened events.
+    app.comment_notification:
+        class: AppBundle\EventListener\CommentNotificationListener
+        arguments: ['@mailer', '@router', '@translator', '%app.notifications.email_sender%']
```
stof: where is the ``app.notifications.email_sender`` parameter initialized ?

stof: OK, it is already part of the codebase although not used for notifications today

src/AppBundle/EventListener/CommentNotificationListener.php
voronkovich: Ok. I've removed the default sender's value.

app/config/services.yml
```diff
@@ -31,9 +31,18 @@ services:
...
+        # There is an optional tag attribute called "method" which defines which method to execute when the event is triggered.
+        # By default the name of the method is on + "camel-cased event name".
+        # If the event is "comment.created" the method executed by default is onCommentCreated().
```
voronkovich: @javiereguiluz,  I've added a note about listener method's default name.

src/AppBundle/EventListener/CommentNotificationListener.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        $comment = $event->getSubject();
+        $post = $comment->getPost();
+
+        $linkToPost = $this->urlGenerator->generate('blog_post', ['slug' => $post->getSlug()], UrlGeneratorInterface::ABSOLUTE_URL);
+
+        $subject = $this->translator->trans('post.recieved_comment');
+        $body = $this->translator->trans('post.recieved_comment_message', [
+            '%title%' => $post->getTitle(),
+            '%link%' => $linkToPost.'#comment_'.$comment->getId(),
```
javiereguiluz: @voronkovich first, thank you for your patience with this code review 😄 

I have a last comment about your PR. What if we use the shinny new [URL fragment support](http://symfony.com/blog/new-in-symfony-3-2-routing-improvements#added-support-for-url-fragments) in Symfony 3.2 to forget about concatenating strings to define fragments?

voronkovich: @javiereguiluz, it's a good idea! I've updated an url generating to use the `_fragment` parameter.

app/Resources/translations/messages.en.xlf
```diff
@@ -276,6 +276,14 @@
                 <source>post.deleted_successfully</source>
                 <target>Post deleted successfully!</target>
             </trans-unit>
+            <trans-unit id="post.comment_added">
```
yceruto: Just for consistency matter (with event name and method name) shouldn't be named `post.comment_created` instead?

voronkovich: @yceruto, I think you're right! I've changed messages names and added a more appropriate prefix for them.

javiereguiluz:
@voronkovich another great feature added to the Demo app! Thank you very much for contributing this.

Issue #418, opened 2017-01-02
Issue with doctrine:fixtures:load


javiereguiluz:
As reported in https://github.com/symfony/symfony/issues/21120  if you don't have SQLite enabled in your PHP, you can't load fixtures and sadly, the Doctrine error message is not very helpful in this case. We should either improve this error message or add some check on our own to detect if SQLite is not enabled.

javiereguiluz:
Fixed by #422.

PR #422, opened 2017-01-05
Added an event subscriber to display better error messages in the console

javiereguiluz:
I propose this PR because:

* It could solve problems like #418
* It allows us to introduce an event subscriber (so far we only have an event listener)

If you agree, I'll polish this and add some helpful comments:

console_exception

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -31,9 +31,9 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest }
 
-    # Event subscribers are similar to event listeners. The main difference is
-    # that the classes define the events they listen to, instead of defining them
-    # in the service definition.
+    # Event subscribers are similar to event listeners but they don't need service tags.
+    # Instead, the PHP class of the event subscriber includes a method that returns
+    # the list of events listened by that class.
     # See http://symfony.com/doc/current/event_dispatcher.html#creating-an-event-subscriber
     app.console_subscriber:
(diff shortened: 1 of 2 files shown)


javiereguiluz:
I've added the help notes, so this PR can now be reviewed. Thanks!

src/AppBundle/EventListener/ConsoleEventSubscriber.php
```diff
@@ -16,16 +16,37 @@
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
+/**
+ * This application uses by default a SQLite database to store its information.
```
bocharsky-bw: I think it should be `an SQLite`

app/config/services.yml
```diff
@@ -22,12 +22,24 @@ services:
...
+    # Event Listeners are classes that listen to one or more specific events.
+    # Those events are defined in the tags added to the service definition.
+    # See http://symfony.com/doc/current/event_dispatcher.html#creating-an-event-listener
     app.redirect_to_preferred_locale_listener:
         class: AppBundle\EventListener\RedirectToPreferredLocaleListener
         arguments: ['@router', '%app_locales%', '%locale%']
         tags:
             - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest }
 
+    # Event subscribers are similar to event listeners but they don't need service tags.
```
xabbuh: This reads confusing as you still need to make use of the `kernel.event_subscriber` tag.

PR #424, opened 2017-01-06
Use font-awesome icon in "form_errors" block

yceruto:
By default [`bootstrap_3_layout.html.twig` template use `glyphicon-` icon when the `form_errors` block is rendered](https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Twig/Resources/views/Form/bootstrap_3_layout.html.twig#L259). In this demo app we use font-awesome (`fa-`) icon library, so this PR try to fix that.

| Before | After |
| ---- | ---- |
| post-comment-error  | post-comment-fixed  |

### DIFF
--- a/app/Resources/views/form/layout.html.twig
+++ b/app/Resources/views/form/layout.html.twig
@@ -1,0 +1,16 @@
+{% extends 'bootstrap_3_layout.html.twig' %}
+
+{# Errors #}
+
+{% block form_errors -%}
+    {% if errors|length > 0 -%}
+        {% if form.parent %}<span class="help-block">{% else %}<div class="alert alert-danger">{% endif %}
+        <ul class="list-unstyled">
+        {%- for error in errors -%}
+            {# use font-awesome icon library #}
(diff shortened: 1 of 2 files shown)


app/Resources/views/form/layout.html.twig
```diff
@@ -0,0 +1,16 @@
...
+
+{% block form_errors -%}
+    {% if errors|length > 0 -%}
+        {% if form.parent %}<span class="help-block">{% else %}<div class="alert alert-danger">{% endif %}
+        <ul class="list-unstyled">
+        {%- for error in errors -%}
+            {# use font-awesome icon library #}
+            <li><span class="fa fa-exclamation-triangle"></span> {{ error.message }}</li>
+        {%- endfor -%}
+    </ul>
```
voronkovich: @yceruto, you are missing indentation here

yceruto: Fixed, thanks!

javiereguiluz:
@yceruto a very nice catch! Thanks for reporting this problem and solving it elegantly.

PR #425, opened 2017-01-09
Add a blog posts RSS channel

yceruto:
This try to implement a simple RSS channel accepted in https://github.com/symfony/symfony-demo/issues/155 since Aug 2015.

Knowledge that brings:
* Custom XML extension to specify the format of the template (`index.xml.twig`).
* How to render a simple XML file.

Any suggestions would be welcome! ;)

### DIFF
--- a/app/Resources/views/base.html.twig
+++ b/app/Resources/views/base.html.twig
@@ -46,12 +46,6 @@
 
                                 {% block header_navigation_links %}
                                     <li>
-                                        <a href="{{ path('blog_rss') }}">
-                                            <i class="fa fa-rss" aria-hidden="true"></i> {{ 'menu.rss'|trans }}
-                                        </a>
-                                    </li>
-
-                                    <li>
                                         <a href="{{ path('blog_index') }}">
                                             <i class="fa fa-home" aria-hidden="true"></i> {{ 'menu.homepage'|trans }}
                                         </a>
--- a/app/Resources/views/base.html.twig
+++ b/app/Resources/views/base.html.twig
@@ -9,6 +9,7 @@
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1"/>
         <title>{% block title %}Symfony Demo application{% endblock %}</title>
+        <link rel="alternate" type="application/rss+xml" href="{{ path('blog_rss') }}">
         {% block stylesheets %}
             <link rel="stylesheet" href="{{ asset('css/bootstrap-flatly-3.3.7.min.css') }}">
             <link rel="stylesheet" href="{{ asset('css/font-awesome-4.6.3.min.css') }}">
@@ -44,6 +45,12 @@
                             <ul class="nav navbar-nav navbar-right">
 
                                 {% block header_navigation_links %}
+                                    <li>
+                                        <a href="{{ path('blog_rss') }}">
+                                            <i class="fa fa-rss" aria-hidden="true"></i> {{ 'menu.rss'|trans }}
+                                        </a>
+                                    </li>
+
                                     <li>
                                         <a href="{{ path('blog_index') }}">
                                             <i class="fa fa-home" aria-hidden="true"></i> {{ 'menu.homepage'|trans }}
--- a/app/Resources/views/base.html.twig
+++ b/app/Resources/views/base.html.twig
@@ -9,7 +9,7 @@
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1"/>
         <title>{% block title %}Symfony Demo application{% endblock %}</title>
-        <link rel="alternate" type="application/rss+xml" href="{{ path('blog_rss') }}">
+        <link rel="alternate" type="application/rss+xml" title="{{ 'rss.title'|trans }}" href="{{ path('blog_rss') }}">
         {% block stylesheets %}
             <link rel="stylesheet" href="{{ asset('css/bootstrap-flatly-3.3.7.min.css') }}">
             <link rel="stylesheet" href="{{ asset('css/font-awesome-4.6.3.min.css') }}">
--- a/src/AppBundle/Controller/BlogController.php
+++ b/src/AppBundle/Controller/BlogController.php
@@ -35,7 +35,7 @@ class BlogController extends Controller
 {
     /**
      * @Route("/", defaults={"page": "1", "_format"="html"}, name="blog_index")
-     * @Route("/rss.{_format}", name="blog_rss", defaults={"page": "1", "_format"="xml"}, requirements={"_format": "xml"})
+     * @Route("/rss.xml", name="blog_rss", defaults={"page": "1", "_format"="xml"})
      * @Route("/page/{page}", defaults={"_format"="html"}, requirements={"page": "[1-9]\d*"}, name="blog_index_paginated")
      * @Method("GET")
      * @Cache(smaxage="10")
--- a/src/AppBundle/Controller/BlogController.php
+++ b/src/AppBundle/Controller/BlogController.php
@@ -34,16 +34,24 @@ use Symfony\Component\HttpFoundation\Response;
 class BlogController extends Controller
 {
     /**
-     * @Route("/", defaults={"page": "1"}, name="blog_index")
-     * @Route("/page/{page}", requirements={"page": "[1-9]\d*"}, name="blog_index_paginated")
+     * @Route("/", defaults={"page": "1", "_format"="html"}, name="blog_index")
+     * @Route("/rss.{_format}", name="blog_rss", defaults={"page": "1", "_format"="xml"}, requirements={"_format": "xml"})
+     * @Route("/page/{page}", defaults={"_format"="html"}, requirements={"page": "[1-9]\d*"}, name="blog_index_paginated")
      * @Method("GET")
      * @Cache(smaxage="10")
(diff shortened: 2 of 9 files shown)


src/AppBundle/Controller/BlogController.php
```diff
@@ -126,4 +127,25 @@ public function commentFormAction(Post $post)
...
+    public function rssAction()
+    {
+        $posts = $this->getDoctrine()->getRepository(Post::class)->findLatest();
+
+        $response = new Response();
+        // Like the Request, the Response object has also a public headers property.
+        // This is a ResponseHeaderBag that has some nice methods for getting and
+        // setting response headers. The header names are normalized so that using
+        // Content-Type is equivalent to content-type or even content_type.
+        $response->headers->set('Content-Type', 'application/xhtml+xml');
```
javiereguiluz: I'm not sure about this. It's great to show how this works in Symfony ... but it may look "horrible" to newcomers because they might think they always need to do that (but it's optional, because `$this->render()` takes care of that).

app/Resources/views/blog/rss.xml.twig
```diff
@@ -0,0 +1,23 @@
+{% spaceless %}
```
javiereguiluz: Just asking: why this tag?

app/Resources/views/blog/rss.xml.twig
```diff
@@ -0,0 +1,23 @@
...
+<?xml version="1.0" encoding="UTF-8" ?>
+<rss version="2.0">
+    <channel>
```
javiereguiluz: What if we add these two tags to the `<channel>` tag?

```twig
<pubDate>{{ 'now'|date('r') }}</pubDate>
<lastBuildDate>{{ (posts|last).publishedAt|default('now')|date('r') }}</lastBuildDate>
```

* pubDate: The original publication date for the channel or item. (optional)
* lastBuildDate: The most recent time the content of the channel was modified. (optional)

They are totally optional, so if you don't agree that's OK!

app/Resources/views/blog/rss.xml.twig
```diff
@@ -0,0 +1,23 @@
...
+        <language>{{ app.request.locale }}</language>
+
+        {% set RSS = constant('RSS', date()) %}
+        {% for post in posts %}
+            <item>
+                <title>{{ post.title }}</title>
+                <description>{{ post.summary }}</description>
+                <link>{{ url('blog_post', {'slug': post.slug}) }}</link>
+                <guid>{{ url('blog_post', {'slug': post.slug}) }}</guid>
+                <pubDate>{{ post.publishedAt|date(format=RSS, timezone='GMT') }}</pubDate>
```
javiereguiluz: Here I also have doubts about:

```twig
{% set RSS = constant('RSS', date()) %}
<pubDate>{{ post.publishedAt|date(format=RSS, timezone='GMT') }}</pubDate>
```

Why not using this?

```twig
<pubDate>{{ post.publishedAt|date('r', timezone='GMT') }}</pubDate>
```

Showing the Twig constants and all it's great ... but we should be careful. If newcomers see this, they'll think that Symfony is this complicated and you must do it like this.

app/Resources/views/base.html.twig
```diff
@@ -45,6 +45,12 @@
 
                                 {% block header_navigation_links %}
```
javiereguiluz: Maybe we should also add the following to the `<head>` part of the page:

```twig
<link rel="alternate" type="application/rss+xml" href="{{ path('blog_rss') }}">
```

src/AppBundle/Controller/BlogController.php
```diff
@@ -126,4 +127,25 @@ public function commentFormAction(Post $post)
             'form' => $form->createView(),
         ]);
     }
+
+    /**
+     * @Route("/rss.xml", name="blog_rss")
+     * @Method("GET")
+     */
+    public function rssAction()
```
voronkovich: @yceruto, I think it's not needed to create a particular action. You could use the `_format` parameter in the `indexAction`. See http://symfony.com/doc/current/quick_tour/the_controller.html#using-formats

src/AppBundle/Controller/BlogController.php
yceruto: :+1:  Simplified by using `_format`.

app/Resources/views/blog/rss.xml.twig
yceruto: Just for remove space between tags, but this detail can be ignored for simplicity :+1: Removed!

app/Resources/views/blog/rss.xml.twig
yceruto: :+1:  Added!

app/Resources/views/blog/rss.xml.twig
yceruto: You are right, RFC822 (`r`) and RSS constants have the same format. Fixed!

app/Resources/views/base.html.twig
yceruto: :+1: Added!

src/AppBundle/Controller/BlogController.php
yceruto: :+1: Updated!

yceruto:
@javiereguiluz, @voronkovich Thanks for the review. I have addressed your comments.

src/AppBundle/Controller/BlogController.php
```diff
@@ -34,16 +34,24 @@
 class BlogController extends Controller
 {
     /**
-     * @Route("/", defaults={"page": "1"}, name="blog_index")
-     * @Route("/page/{page}", requirements={"page": "[1-9]\d*"}, name="blog_index_paginated")
+     * @Route("/", defaults={"page": "1", "_format"="html"}, name="blog_index")
+     * @Route("/rss.{_format}", name="blog_rss", defaults={"page": "1", "_format"="xml"}, requirements={"_format": "xml"})
```
javiereguiluz: If we decide to have a different `@Route` for the RSS, maybe we can hardcode some things to simplify:

```
@Route("/rss.xml", name="blog_rss", defaults={"page": "1", "_format"="xml"})
```

yceruto: Totally agree!

app/Resources/views/base.html.twig
```diff
@@ -9,6 +9,7 @@
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1"/>
         <title>{% block title %}Symfony Demo application{% endblock %}</title>
+        <link rel="alternate" type="application/rss+xml" href="{{ path('blog_rss') }}">
```
yceruto: The `Content-Type` that we return here is `text/xml` (by default), is it a contradiction? should change this `type` to `text/xml` then?

javiereguiluz: To be honest I don't know. It makes sense to only use `text/xml` ... but then, will browsers recognize that as a RSS link?

yceruto: According to http://www.rssboard.org/rss-autodiscovery#element-link-type
> The type attribute MUST contain the feed's MIME type, which is "application/rss+xml" for RSS 1.0 or RSS 2.0 feeds.

src/AppBundle/Controller/BlogController.php
```diff
@@ -34,16 +34,24 @@
 class BlogController extends Controller
 {
     /**
-     * @Route("/", defaults={"page": "1"}, name="blog_index")
-     * @Route("/page/{page}", requirements={"page": "[1-9]\d*"}, name="blog_index_paginated")
+     * @Route("/", defaults={"page": "1", "_format"="html"}, name="blog_index")
+     * @Route("/rss.xml", name="blog_rss", defaults={"page": "1", "_format"="xml"})
```
bocharsky-bw: Could we keep the order of arguments, i.e. move `name="blog_rss"` to the end like in others? 😇

src/AppBundle/Controller/BlogController.php
```diff
@@ -34,16 +34,24 @@
...
+     * @Route("/page/{page}", defaults={"_format"="html"}, requirements={"page": "[1-9]\d*"}, name="blog_index_paginated")
      * @Method("GET")
      * @Cache(smaxage="10")
+     *
+     * NOTE: For standard formats, Symfony will also automatically choose the best
+     * Content-Type header for the response.
+     * See http://symfony.com/doc/current/quick_tour/the_controller.html#using-formats
      */
-    public function indexAction($page)
+    public function indexAction($page, $_format)
```
xabbuh: I would still name this variable `$format`.

yceruto: We lost then the auto `Content-Type` header response in `blog_rss` route, remember we need the special variable `_format` into `defaults` option. ;)

xabbuh: Sure, you're right.

javiereguiluz:
@yceruto thanks to the help of @iltar and @fbourigault in the Symfony Slack channel, I've managed to add a commit to your PR. The change I propose is to remove the RSS link from the header, which makes it look crowded:

before

---

And instead display it in the sidebar of blog/index and blog/show:

after

javiereguiluz:
@yceruto thanks for implementing this feature. I truly love it. Thanks!

PR #427, opened 2017-01-10
Do not use a Symfony Form to delete an article

lyrixx:
Because it's simpler without a SF Form

### DIFF
--- a/app/Resources/views/admin/blog/show.html.twig
+++ b/app/Resources/views/admin/blog/show.html.twig
@@ -38,12 +38,7 @@
     </div>
 
     <div class="section">
-        {{ include('admin/blog/_form.html.twig', {
-            form: delete_form,
-            button_label: 'action.delete_post'|trans,
-            button_css: 'btn btn-lg btn-block btn-danger',
-            show_confirmation: true,
-        }, with_context = false) }}
+        {{ include('admin/blog/_delete_form.html.twig', { post: post }, with_context = false) }}
     </div>
 
     {{ show_source_code(_self) }}
--- a/src/AppBundle/Controller/Admin/BlogController.php
+++ b/src/AppBundle/Controller/Admin/BlogController.php
@@ -126,11 +126,8 @@ class BlogController extends Controller
             throw $this->createAccessDeniedException('Posts can only be shown to their authors.');
         }
 
-        $deleteForm = $this->createDeleteForm($post);
-
         return $this->render('admin/blog/show.html.twig', [
             'post' => $post,
-            'delete_form' => $deleteForm->createView(),
         ]);
     }
 
@@ -148,12 +145,11 @@ class BlogController extends Controller
 
         $entityManager = $this->getDoctrine()->getManager();
 
-        $editForm = $this->createForm(PostType::class, $post);
-        $deleteForm = $this->createDeleteForm($post);
+        $form = $this->createForm(PostType::class, $post);
 
-        $editForm->handleRequest($request);
+        $form->handleRequest($request);
 
-        if ($editForm->isSubmitted() && $editForm->isValid()) {
+        if ($form->isSubmitted() && $form->isValid()) {
             $post->setSlug($this->get('slugger')->slugify($post->getTitle()));
             $entityManager->flush();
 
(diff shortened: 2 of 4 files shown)


src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -164,16 +160,15 @@ public function editAction(Post $post, Request $request)
...
+            'form' => $form->createView(),
         ]);
     }
 
     /**
      * Deletes a Post entity.
      *
      * @Route("/{id}", name="admin_post_delete")
-     * @Method("DELETE")
+     * @Method("POST")
```
voronkovich: @lyrixx, you could use `_method` field in the form instead of changing the required HTTP method to POST. http://symfony.com/doc/current/form/action_method.html

lyrixx: I know, but I always found this totally useless ;)

stof: if you change it to POST (which is something I do all the time to avoid having to enable the HttpMethodOverride which is disabled by default), I suggest changing the URL to ``/{id}/delete`` though

javiereguiluz: Why don't you like the `DELETE` method? Is because browsers don't support it and we must simulate it with the `_method` trick? Thanks!

javiereguiluz: So you don't use the `DELETE` method either in your apps? Why don't you want to enable `HttpMethodOverride`? Security reasons? Performance reasons? Thanks!

lyrixx: @stof good catch, I will change the URL
@javiereguiluz It just simulate a verb. But what does it bring? IMHO nothing except more code for no real value.

app/Resources/views/admin/blog/show.html.twig
```diff
@@ -38,12 +38,7 @@
     </div>
 
     <div class="section">
-        {{ include('admin/blog/_form.html.twig', {
-            form: delete_form,
-            button_label: 'action.delete_post'|trans,
-            button_css: 'btn btn-lg btn-block btn-danger',
-            show_confirmation: true,
-        }, with_context = false) }}
+        {{ include('admin/blog/_delete_form.html.twig', { post: post}, with_context = false) }}
```
yceruto: Missing space to end `{ post: post }` or remove extra space to beginning `{post: post}` ? same in other files.

javiereguiluz: Nice catch! The style we use adds a space before and after the braces, so it should be `{ post: post }`

lyrixx: Thanks. Fixed.

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -182,40 +177,17 @@ public function editAction(Post $post, Request $request)
      */
     public function deleteAction(Request $request, Post $post)
     {
-        $form = $this->createDeleteForm($post);
-        $form->handleRequest($request);
+        if (!$this->isCsrfTokenValid('delete', $request->request->get('token'))) {
+            return $this->redirectToRoute('admin_post_index');
```
yceruto: The request could come from "edit page", is right to redirect to index page? shouldn't be `admin_post_edit`?

In the other hand, could you add a flash messages (error) when this happen? Similar to:
```
The CSRF token is invalid. Please try to resubmit the form.
```

src/AppBundle/Controller/Admin/BlogController.php
stof: @javiereguiluz enabling it make it easier to create CSRF attacks (well, not for this action as it has a CSRF token). This is why it is disabled by default in Symfony.

and it brings no value (the browser is not doing a DELETE request) and can become a nightmare when dealing with cache proxies (because it is a POST request, and you cannot add a Vary on this override to use the right cache.

src/AppBundle/Controller/Admin/BlogController.php
lyrixx: Actually I keep the same behavior than before my changes...

javiereguiluz:
@lyrixx thanks for simplifying this feature!

PR #428, opened 2017-01-10
Simplify the post list (admin) displaying your posts only

yceruto:
We are currently unable to show a post if the logged user is not the author (i.e. in backend).

| Before | After |
| ---- | ---- |
| posts-list-before | posts-list-after |

### DIFF
--- a/app/Resources/views/admin/blog/index.html.twig
+++ b/app/Resources/views/admin/blog/index.html.twig
@@ -9,16 +9,14 @@
         <thead>
             <tr>
                 <th scope="col">{{ 'label.title'|trans }}</th>
-                <th scope="col"><i class="fa fa-user" aria-hidden="true"></i> {{ 'label.author'|trans }}</th>
                 <th scope="col"><i class="fa fa-calendar" aria-hidden="true"></i> {{ 'label.published_at'|trans }}</th>
-                <th scope="col"><i class="fa fa-cogs" aria-hidden="true"></i> {{ 'label.actions'|trans }}</th>
+                <th scope="col" class="col-xs-1"><i class="fa fa-cogs" aria-hidden="true"></i> {{ 'label.actions'|trans }}</th>
             </tr>
         </thead>
         <tbody>
         {% for post in posts %}
             <tr>
                 <td>{{ post.title }}</td>
-                <td>{{ post.authorEmail }}</td>
                 {# it's not mandatory to set the timezone in localizeddate(). This is done to
                    avoid errors when the 'intl' PHP extension is not available and the application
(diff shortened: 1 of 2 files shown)


bocharsky-bw:
I think you can show posts even if you're not the author of it.

yceruto:
Make sense show a post which we cannot act?

bocharsky-bw:
It's a controversial question since all posts are public on frontend and anyone can read it... so why we should deny reading them on backend?

yceruto:
@bocharsky-bw I agree with you, so there are two options:
* Allow show a post and hide "edit", "delete" actions inside?
* Show only the posts of the authenticated user (author) in this list?

I go for the second.

voronkovich:
I think that showing post only for it's author is very strange behavior :) I would allow all authenticated users to view any post.

yceruto:
@voronkovich

> I would allow all authenticated users to view any post.

IMHO this is the goal of the frontend.

If I enter to backend I'll want to see my posts only, to review/edit it or create a new one.

voronkovich:
@yceruto, thanks for the explanation. It makes sense.  So, I would choose the second option: `Show only the posts of the authenticated user (author) in this list`.

javiereguiluz:
This might be trickier than I thought. For example, if I browse the backend of a WordPress blog managed by several users, I can see (and most of the times, edit) the articles written by all authors, not only me.

If you agree, we could do this instead:

* In the "author" column, display "You" is you are the author of the article. Otherwise, keep displaying the author's email.
* Keep the `Show` button for all articles
* Keep the `Edit` button for all articles ... but disable it for articles that you are not the author. Bonus: display a hover message in those buttons saying: "You can't edit this article because you are not the author".

Does it make sense?

yceruto:
@javiereguiluz I think it's fine, but what do we do with the "show" button? we allow show the post and we hide the edit/delete actions inside it? or we disable it too?

javiereguiluz:
@yceruto if we do this, then yes, we should display but disable the edit and delete buttons.

But there's something I don't like about this proposal and I don't know what is. It doesn't feel right to me :(

yceruto:
@javiereguiluz to be honest, me neither. IMO this is a multi-user app that has two scenarios, frontend (public) and backend (private), so for me, just to filter the posts list by the logged user would simplify the things. What do you think about this alternative?

javiereguiluz:
Yes ... maybe it's better to only display your posts and see if we like the result 👍

yceruto:
Updated!

posts-list

yceruto:
Rebased with master!

javiereguiluz:
@yceruto much simpler and better now! Thanks.

PR #430, opened 2017-01-11
Display missing "about" section in admin blog show page

yceruto:
Also I've added in this PR some minor simplifications ;)

### DIFF
--- a/app/Resources/views/admin/blog/show.html.twig
+++ b/app/Resources/views/admin/blog/show.html.twig
@@ -46,5 +46,7 @@
         }, with_context = false) }}
     </div>
 
+    {{ parent() }}
+
     {{ show_source_code(_self) }}
 {% endblock %}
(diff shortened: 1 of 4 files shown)


app/Resources/views/admin/blog/show.html.twig
```diff
@@ -46,5 +46,7 @@
         }, with_context = false) }}
     </div>
 
+    {{ parent() }}
```
bocharsky-bw: Probably, the original idea was to show that we can override content of parent block, but I can be wrong here...

javiereguiluz: I don't remember ... but the rest of admin/* templates include this `parent()` ... so this is probably a bug.

src/AppBundle/DataFixtures/ORM/LoadFixtures.php
```diff
@@ -34,8 +34,7 @@
  */
 class LoadFixtures implements FixtureInterface, ContainerAwareInterface
```
javiereguiluz: `ContainerAwareInterface` should be removed here too

stof: no. The class is still container aware (but it uses the trait to implement it instead of doing it explicitly)

javiereguiluz: I'm sorry! I'm not very familiar with traits.

javiereguiluz:
Thanks @yceruto.

Issue #423, opened 2017-01-06
Driver error


Thibault34:
Hello,

When I go in http://localhost:8000/fr/blog/ I have error :
An exception has been thrown during the rendering of a template ("An exception occured in driver: could not find driver").

 Can you help me please ?

voronkovich:
@Thibault34, see https://github.com/symfony/symfony/issues/21120

stof:
you need to enable the sqlite extension, as the demo uses sqlite by default

Thibault34:
Tanks you @stof

javiereguiluz:
@Thibault34 let me reopen this issue because maybe we can apply an idea similar to #422 and show a better error message in the browser when you don't have SQLite enabled.

javiereguiluz:
Closing it as fixed by #452.

PR #431, opened 2017-01-11
Show better message when SQLite is not enabled

uikolas:
As mentioned in #423 
I added to show better error message in the browser, if SQLite is not enabled. (I think that message should contain about sqllite3 and pdo_sqlite to enable)
But i have have doubts, how to show it? As a exception or simple text response?

error image


voronkovich:
@uikolas, IMO if we already have the [ConsoleEventSubscriber](https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/EventListener/ConsoleEventSubscriber.php) you should add your code there. Just rename it to `CheckSQLiteSubscriber` and move there your `onKernelException` method.

src/AppBundle/EventListener/CheckSQLiteEventSubscriber.php
```diff
@@ -57,4 +61,16 @@ public function handleDatabaseExceptions(ConsoleExceptionEvent $event)
...
     }
+
+    /**
+     * This method is triggered when kernel exception occurs. And checks if sqlite extension is enabled.
+     *
+     * @param GetResponseForExceptionEvent $event
+     */
+    public function onKernelException(GetResponseForExceptionEvent $event)
+    {
+        if (!extension_loaded('sqlite3')) {
```
stof: should be done only if the exception is a DBAL exception IMO. Otherwise, it could hide other issues happening earlier

uikolas: Yes, i agree with that. But the problem is that from $event->getException() i get _Twig_Error_Runtime_ and don't know why. Maybe is documented somewhere?

stof: Well, any exception thrown during a Twig template rendering is wrapped in a Twig_Error_Runtime. You can check the chain of previous exceptions

uikolas: Oh, didn't know that. Ok :)

src/AppBundle/EventListener/CheckSQLiteEventSubscriber.php
```diff
@@ -69,7 +70,12 @@ public function handleDatabaseExceptions(ConsoleExceptionEvent $event)
      */
     public function onKernelException(GetResponseForExceptionEvent $event)
     {
-        if (!extension_loaded('sqlite3')) {
+        $exception = $event->getException();
+        // Since any exception thrown during a Twig template rendering is wrapped in a Twig_Error_Runtime.
+        // We must get the original exception.
+        $previousException = $exception->getPrevious();
+
+        if ($previousException instanceof DriverException && !extension_loaded('sqlite3')) {
```
stof: the driver exception may still be the main exception depending on where it happens (it can happen in the controller)

src/AppBundle/EventListener/CheckSQLiteEventSubscriber.php
```diff
@@ -69,7 +70,12 @@ public function handleDatabaseExceptions(ConsoleExceptionEvent $event)
      */
     public function onKernelException(GetResponseForExceptionEvent $event)
     {
-        if (!extension_loaded('sqlite3')) {
+        $exception = $event->getException();
+        // Since any exception thrown during a Twig template rendering is wrapped in a Twig_Error_Runtime.
+        // We must get the original exception.
+        $previousException = $exception->getPrevious();
```
yceruto: can be in one line `$previousException = $event->getException()->getPrevious();`

src/AppBundle/EventListener/CheckSQLiteEventSubscriber.php
```diff
@@ -57,4 +62,24 @@ public function handleDatabaseExceptions(ConsoleExceptionEvent $event)
...
+     */
+    public function onKernelException(GetResponseForExceptionEvent $event)
+    {
+        $exception = $event->getException();
+        // Since any exception thrown during a Twig template rendering is wrapped in a Twig_Error_Runtime.
+        // We must get the original exception.
+        $previousException = $exception->getPrevious();
+
+        // Driver exception may happen in controller or in twig template rendering
+        $isDriverException = ($exception instanceof DriverException || $previousException instanceof DriverException);
```
voronkovich: I think we should make shure that a connection is configured to use an `sqlite` driver because it can be an error with another one (`mysql`, 'postgresql` and etc.).

uikolas: Yes, you are totally  correct :)

voronkovich: @uikolas, you could use a platform name:
```php
$entityManager->getConnection()->getDatabasePlatform()->getName() === 'sqlite'
```
See https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php#L511

javiereguiluz:
@uikolas thanks for improving the detection of the SQLite requirement!

PR #432, opened 2017-01-12
Design tweaks for small screens

javiereguiluz:
A typical problem I see with lots of Bootstrap-based templates is that they look awful in a very specific range of page widths. When the browser window is not too small and not too big, the template constraints the container width too much. This PR tries to fix this.

| Before | After
| --- | ---
| before_container | after_container


stof:
👍

PR #434, opened 2017-01-14
Add user relationship to Post and Comment entities

yceruto:
Actually, we are related Post/User and Comment/User indirectly throught `authorEmail` field which IMHO makes things more complicated to understand for the beginner, because really we need this relationship.

Related discussion in https://github.com/symfony/symfony-demo/issues/429.


voronkovich:
@yceruto, I think we don't need the relation between `User` and `Comment`. It's better to allow anonymous users create comments.

yceruto:
@voronkovich before this PR to comment on Post is needed to be logged, just I keep the current behavior for now :)

voronkovich:
@yceruto, ok, let's keep it as is for now.

src/AppBundle/Entity/Post.php
```diff
@@ -149,35 +149,38 @@ public function setContent($content)
...
     }
 
-    public function setPublishedAt(\DateTime $publishedAt)
+    /**
+     * Is the given User the author of this Post?
+     */
+    public function isAuthor(User $user = null)
     {
-        $this->publishedAt = $publishedAt;
+        return $user === $this->author;
```
bocharsky-bw: I think we need an extra check here since the user variable might be `null`:
```php
if (null === $user) {
    return false;
}
```

yceruto: For extra readability? when the user is `null` anyway it already returns `false`.

voronkovich: @yceruto, both variables `$user` and `$this->author` can be null at the same time.

yceruto: it was necessary before since we check `$user->getEmail()` ;)

yceruto: @voronkovich It should not happen I think, as [each `Post` is initialized with its author](https://github.com/yceruto/symfony-demo/blob/a32f77632325311176442d8b0ca10b67b918cfb0/src/AppBundle/Controller/Admin/BlogController.php#L76) and [is required](https://github.com/yceruto/symfony-demo/blob/a32f77632325311176442d8b0ca10b67b918cfb0/src/AppBundle/Entity/Post.php#L86). I missing something?

voronkovich: @yceruto, you're right! So, we don't need to check the user variable for null value here.

bocharsky-bw: As Oleg said, both variables $user and $this->author can be null at the same time - I just want to prevent weird bugs in the future. I think it's unsafe just rely on `nullable` option which could be changed.

yceruto:
Rebase with master and added more tests.

src/AppBundle/Entity/Post.php
yceruto: @bocharsky-bw Sorry for the delay, I understand your preoccupation, but apparently that case is not possible. If a post exists this have a author and this field is not nullable in the database either. Thanks for the discussion guys. :)

yceruto:
In the other hand, it fix a security issue too, related with author email checking:
```
$user->getEmail() === $post->getAuthorEmail()
```
I could create a new admin user and set the same email that another user and see/edit/delete thier posts!

src/AppBundle/Entity/Post.php
javiereguiluz: All of you have given good reasons ... so let's keep this as it is to move on ... and we'll add the extra check if needed in the future. Thanks!

javiereguiluz:
This was a much needed improvement. @yceruto thnks for contributing this feature.

PR #435, opened 2017-01-16
By default redirect to post index page after login

yceruto:
This PR change the default target path `/` to <del>`admin_index`</del> `blog_index` route, useful when we access to login page directly (i.e. through URL).

### DIFF
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -32,6 +32,8 @@ security:
                 # Secure the login form against CSRF
                 # Reference: http://symfony.com/doc/current/cookbook/security/csrf_in_login_form.html
                 csrf_token_generator: security.csrf.token_manager
+                # Redirect to post index page after login (by default)
+                default_target_path: blog_index
 
             logout:
                 # The route name the user can go to in order to logout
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
Looks like in this case, users, which don't have access to the admin page, will have an "Access denied" error.

yceruto:
@bocharsky-bw you're right, but actually already happening, at least the most of times, the users accessing to login page through "browser backend" button and they get a "Access Denied".

Something like "redirect to post index for users" or "redirect to admin index for admins" would be the best approach.

Any suggestion to improving it?

stof:
@yceruto if you want to do it, you need to create a dedicated route (used as target path) for which the controller would do redirections conditionally.

yceruto:
> Something like "redirect to post index for users" or "redirect to admin index for admins" would be the best approach.

I'm agree that @stof's and @javiereguiluz's idea may be a bit overkill for this demo app and I abandon this idea.

But, my initial goal here was to find a better page that `/` for "the user" (admin or not) after login **only in this case**:

**When the user access to login page directly (i.e. through URL) then this option (`default_target_path`) is used and not through "Browser Backend" button from home page, which use the target path saved in session.**

So, to make it simpler and to avoid "Access denied" for users, I've changed again the `default_target_path` option to `blog_index` instead now.

What do you think now?

PR #437, opened 2017-01-17
Tweaked the admin blog index

javiereguiluz:
| Before | After
| --- | ---
| before-list | after-list

### DIFF
--- a/app/Resources/views/admin/blog/index.html.twig
+++ b/app/Resources/views/admin/blog/index.html.twig
@@ -5,12 +5,12 @@
 {% block main %}
     <h1>{{ 'title.post_list'|trans }}</h1>
 
-    <table class="table table-striped">
+    <table class="table table-striped table-middle-aligned">
         <thead>
             <tr>
                 <th scope="col">{{ 'label.title'|trans }}</th>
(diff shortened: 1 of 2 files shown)


app/Resources/views/admin/blog/index.html.twig
```diff
@@ -20,8 +20,8 @@
                 {# it's not mandatory to set the timezone in localizeddate(). This is done to
                    avoid errors when the 'intl' PHP extension is not available and the application
                    is forced to use the limited "intl polyfill", which only supports UTC and GMT #}
-                <td>{% if post.publishedAt %}{{ post.publishedAt|localizeddate('short', 'short', null, 'UTC') }}{% endif %}</td>
-                <td>
+                <td>{{ post.publishedAt|localizeddate('medium', 'short', null, 'UTC') }}</td>
```
stof: what if the post is not published yet ? Passing ``null`` to the loalizeddate filter would display the date of today, because this is what ``new DateTime(null)`` does.

javiereguiluz: Apparently that case is not possible. If a post exists, it's published (the date can be in the future) and this field is not nullable in the database either.

stof: Oh, I was not aware that this is the way this date was handled in the project. Fine then

PR #438, opened 2017-01-17
Tweaked the admin form views

javiereguiluz:
* Removed the author email from the form (you can only create posts for yourself)
* Added some help messages for form fields (they are not displayed to the end-user yet)
* Tweaks to the design of the error messages:

| Before | After
| --- | ---
| before-error | after-error

### DIFF
--- a/src/AppBundle/Form/PostType.php
+++ b/src/AppBundle/Form/PostType.php
@@ -49,13 +49,16 @@ class PostType extends AbstractType
                 'label' => 'label.summary',
             ])
             ->add('content', null, [
-                'attr' => ['rows' => 20],
+                'attr' => [
+                    'rows' => 20,
+                    'help' => 'Markdown is allowed.',
+                ],
                 'label' => 'label.content',
             ])
-            ->add('authorEmail', null, [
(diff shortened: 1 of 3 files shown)


src/AppBundle/Form/PostType.php
```diff
@@ -49,13 +49,16 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'label' => 'label.summary',
             ])
             ->add('content', null, [
-                'attr' => ['rows' => 20],
+                'attr' => [
+                    'rows' => 20,
+                    'help' => 'Markdown is allowed.',
```
stof: should it realy be added as a ``help`` attribute ? This is not valid HTML.

If you plan to display it in JS, you should use ``data-help``.
If you plan to display it in Twig, you should use a form type extension to add a new option (and then a new variable in FormView based on it).

Btw, the help message should be translated too

yceruto:
> Removed the author email from the form (you can only create posts for yourself)

This fix the initial goal of https://github.com/symfony/symfony-demo/issues/429

PR #439, opened 2017-01-17
Tweaked the admin blog post show page

javiereguiluz:
And the last PR about tweaking admin pages. This time, the blog post show page (the "author" is gone because you can only see your own blog posts in the backend):

| Before | After
| --- | ---
| before-show | after-show


stof:
should we keep the author name ?

javiereguiluz:
My reasoning was that you can only see your own posts in the backend (of course you can play with the URL parameters and see other posts ... but should we care about that?)

app/Resources/views/admin/blog/show.html.twig
```diff
@@ -4,30 +4,13 @@
 
 {% block main %}
     <h1>{{ post.title }}</h1>
+    <p class="text-muted">{{ post.publishedAt|localizeddate('long', 'medium', null, 'UTC') }}</p>
```
yceruto: Maybe this could be displayed in blog index/show as well? https://github.com/symfony/symfony-demo/issues/105#issuecomment-271021837

javiereguiluz: Wow! I totally missed this comment ... and I almost copied your idea here!  Yes, we could display that information ... but for the users I'd like to display people names instead of emails (so we might need to add a field in the user entity).

yceruto: > but for the users I'd like to display people names instead of emails (so we might need to add a field in the user entity).

I'm totally agree with that!

stof:
> of course you can play with the URL parameters and see other posts ... but should we care about that?

if you should only see your own posts, but changing the URL allows to see any post, this looks like a flaw in the security checks of the backend. It should send a 403. We should use proper security rules in the demo.

javiereguiluz:
@stof yes ... but it's a bit tricky. This demo app is a simplified blog, so we made some decisions such as: all posts are public and published, everybody can see all posts (even anonymous users visiting the blog), etc. The only restriction is that posts can only be edited by their authors.

PR #442, opened 2017-01-17
Add a security Post Voter

yceruto:
Implement https://github.com/symfony/symfony-demo/issues/440 feature.

- [x] Add more code comments and doc references

Following some references and recommendations in:
- https://symfony.com/doc/current/components/security/authorization.html
- http://symfony.com/doc/current/security/voters.html
- https://stovepipe.systems/post/symfony-security-roles-vs-voters (@iltar blog)

Any suggestion is welcome!

### DIFF
--- a/src/AppBundle/Controller/Admin/BlogController.php
+++ b/src/AppBundle/Controller/Admin/BlogController.php
@@ -119,12 +119,9 @@ class BlogController extends Controller
      */
     public function showAction(Post $post)
     {
-        // This security check can also be performed:
-        //   1. Using an annotation: @Security("post.isAuthor(user)")
-        //   2. Using a "voter" (see http://symfony.com/doc/current/cookbook/security/voters_data_permission.html)
-        if (!$post->isAuthor($this->getUser())) {
-            throw $this->createAccessDeniedException('Posts can only be shown to their authors.');
-        }
+        // This security check can also be performed
+        // using an annotation: @Security("is_granted('show', post)")
+        $this->denyAccessUnlessGranted('show', $post, 'Posts can only be shown to their authors.');
 
         return $this->render('admin/blog/show.html.twig', [
             'post' => $post,
@@ -139,9 +136,7 @@ class BlogController extends Controller
      */
     public function editAction(Post $post, Request $request)
     {
-        if (!$post->isAuthor($this->getUser())) {
-            throw $this->createAccessDeniedException('Posts can only be edited by their authors.');
-        }
+        $this->denyAccessUnlessGranted('edit', $post, 'Posts can only be edited by their authors.');
 
         $entityManager = $this->getDoctrine()->getManager();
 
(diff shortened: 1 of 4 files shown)


src/AppBundle/Security/PostVoter.php
```diff
@@ -0,0 +1,61 @@
...
+use AppBundle\Entity\User;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Authorization\Voter\Voter;
+
+/**
+ * @author Yonel Ceruto <yonelceruto@gmail.com>
+ */
+class PostVoter extends Voter
+{
+    const VIEW = 'show';
```
yceruto: I have doubts about this permission, since in the real world any user should see any Post. This is restricted in admin only but is a bit confusing.

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -139,9 +136,7 @@ public function showAction(Post $post)
      */
     public function editAction(Post $post, Request $request)
     {
-        if (null === $this->getUser() || !$post->isAuthor($this->getUser())) {
-            throw $this->createAccessDeniedException('Posts can only be edited by their authors.');
-        }
+        $this->denyAccessUnlessGranted('edit', $post, 'Posts can only be edited by their authors.');
```
yceruto: We should use `PostVoter::EDIT` instead?

src/AppBundle/Entity/Post.php
```diff
@@ -162,14 +162,6 @@ public function setAuthorEmail($authorEmail)
         $this->authorEmail = $authorEmail;
     }
 
-    /**
-     * Is the given User the author of this Post?
-     */
-    public function isAuthor(User $user)
```
yceruto: We should keep it to implement both ways?

src/AppBundle/Security/PostVoter.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        $user = $token->getUser();
+
+        if (!$user instanceof User) {
+            // the user must be logged in; if not, deny access
+            return false;
+        }
+
+        // you know $post is a Post object, thanks to supports
+        return $user->getEmail() === $post->getAuthorEmail();
```
yceruto: This is unsafe checking see https://github.com/symfony/symfony-demo/pull/434#issuecomment-273264016

yceruto:
Rebase with master!

src/AppBundle/Security/PostVoter.php
```diff
@@ -0,0 +1,61 @@
...
+    const EDIT = 'edit';
+    const DELETE = 'delete';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function supports($attribute, $subject)
+    {
+        // this voter is only executed for three specific permissions on Post objects
+        return $subject instanceof Post && in_array($attribute, [self::SHOW, self::EDIT, self::DELETE]);
```
javiereguiluz: I've simplified this method to a 1-liner to reduce the perceived complexity of voters. They have been traditionally criticized for the required boilerplate ... and thanks to the new abstract voter, it's code can be super simple.

src/AppBundle/Security/PostVoter.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function voteOnAttribute($attribute, $post, TokenInterface $token)
```
yceruto: This implementation can be simplified too, just:
```
return $token->getUser() === $post->getAuthor();
```
?

javiereguiluz:
@yceruto I wanted to add a security voter since day one. Thanks for making it happen!

PR #443, opened 2017-01-17
Validate user data before save

yceruto:
1. I've added a `validateUserData` method to check for existent user email (this improves a better messages exception) and validate plainPassword and email in non-interactive mode (i.e. `$ app:add-user username password email`).
2. Renamed the option `--is-admin` to just `--admin`, Imho `is-admin` feels more like a question rather than an instruction. (now `$ app:add-user username password email --admin`)

Any suggestion is welcome!

### DIFF
--- a/src/AppBundle/Command/AddUserCommand.php
+++ b/src/AppBundle/Command/AddUserCommand.php
@@ -35,6 +35,7 @@ use Symfony\Component\Console\Question\Question;
  * See http://symfony.com/doc/current/cookbook/console/console_command.html
  *
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ * @author Yonel Ceruto <yonelceruto@gmail.com>
  */
 class AddUserCommand extends ContainerAwareCommand
 {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@yceruto thanks for these improvements. I like them. The only question I have is about the `--is-admin` to `--admin` reword.

It's true that "is admin" can be understood as a question: "john-doe is admin?" but also as a non-question sentence: "john-doe is admin."  In any case, `--is-admin` is not perfect. However, using just `--admin` is not perfectly clearly either (it can be understood in several ways in addition to the intended one).

So, let's wait for other opinions before merging it. Thanks!

yceruto:
As I see it in others commands with similar option, this flag is directly related to the command name rather than its related options, i.e. `app:add-user --admin`, it's like "I want: to add an admin user".

This approach in [FOSUserBundle commands] (https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Command) for example:
* [`fos:user:create --super-admin --inactive`](https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Command/CreateUserCommand.php#L40-L41 )
* [`fos:user:promote --super`](https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Command/RoleCommand.php#L36)

Other Symfony commands also follow this approach for your flag options, but I'm fine with `--is-admin` approach too :)

javiereguiluz:
That's a nice argument. Reusing the existing practices is always a good idea.

javiereguiluz:
Merged! Thanks for improving this command!

PR #444, opened 2017-01-19
Added more tests for the backend

javiereguiluz:
This complements the new tests added in #434.

### DIFF
--- a/tests/AppBundle/Controller/Admin/BlogControllerTest.php
+++ b/tests/AppBundle/Controller/Admin/BlogControllerTest.php
@@ -27,35 +27,55 @@ use Symfony\Component\HttpFoundation\Response;
  * Execute the application tests using this command (requires PHPUnit to be installed):
  *
  *     $ cd your-symfony-project/
- *     $ phpunit -c app
+ *     $ ./vendor/bin/phpunit
  */
 class BlogControllerTest extends WebTestCase
 {
-    public function testRegularUsersCannotAccessToTheBackend()
+    /**
+     * @dataProvider getUrlsForRegularUsers
+     */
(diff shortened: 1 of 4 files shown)


tests/AppBundle/Controller/Admin/BlogControllerTest.php
```diff
@@ -27,35 +27,55 @@
...
+        $this->assertEquals($statusCode, $client->getResponse()->getStatusCode());
+    }
 
-        $this->assertEquals(Response::HTTP_OK, $client->getResponse()->getStatusCode());
+    public function getUrlsForAdminUsers()
+    {
+        yield ['GET', '/en/admin/post/', Response::HTTP_OK];
+        yield ['GET', '/en/admin/post/1', Response::HTTP_OK];
+        yield ['GET', '/en/admin/post/1/edit', Response::HTTP_OK];
+        yield ['POST', '/en/admin/post/1/delete', Response::HTTP_FOUND];
```
stof: wouldn't this break when running tests several times (or when changing the order of tests), due to changing the state of the DB ?

javiereguiluz: Nice catch. I've opened #445 to improve this and check if we have any other related issue. Thanks!

yceruto: This test no change the DB status for now, because this redirect come from [here](https://github.com/javiereguiluz/symfony-demo/blob/4cfce12ffd912e31fcffde65737c37de1c3f8fb3/src/AppBundle/Controller/Admin/BlogController.php#L181)

yceruto: Only we checking here the security constraint for your author.

PR #447, opened 2017-01-26
Add tags to Post

yceruto:
This implements #44 and it's a complete variant of https://github.com/symfony/symfony-demo/pull/192 (On hold since Ago 2016).

| Blog Post Index | Post Show | 
| ---- | --- |
| tags-index  |  tags-show | 

| Post New/Edit |
| --- |
| tagsinput |

Also this splits the only one fixture class into three classes (`UserFixtures`, `PostFixtures` and `TagFixtures`):
 - [x] To prevent mess and complexity.
 - [x] To avoid post title duplicated (slug issue).
 - [x] To fix example and explanation of: Sharing objects between fixtures (https://github.com/symfony/symfony-demo/pull/192#r39252419).
 - [x] To add a new example about: Fixture ordering and `DependentFixtureInterface` (dependence between fixtures).

Add Translation:
 - [x] English
 - [x] Español

Updated `blog.sqlite` and `blog_test.sqlite` DBs.

---

### Handling Tags (Bootstrap-tagsinput):

There is many options to do that:

 * Form collection to add new tags?
 * [`select2`](https://select2.github.io/examples.html) or [`chosen`](https://github.com/harvesthq/chosen) js plugins?
 * [Bootstrap tagsinput](http://bootstrap-tagsinput.github.io/bootstrap-tagsinput/examples/) js plugin? <-- :+1: 
* [Tokenfield for Bootstrap](http://sliptree.github.io/bootstrap-tokenfield/) js plugin?

### TODO:
- [x] Create TagsInputType to handle the post tags collection (Added DataTransformer example)
- [x] Add typeaheadjs option to show tags hint.


src/AppBundle/DataFixtures/ORM/LoadPostFixtures.php
```diff
@@ -99,6 +75,30 @@ private function loadPosts(ObjectManager $manager)
...
+    {
+        return [
+            LoadTagFixtures::class,
+            LoadUserFixtures::class,
+        ];
+    }
+
+    private function addRandomTags(Post $post)
+    {
+        $indexes = (array) array_rand(LoadTagFixtures::$names, mt_rand(1, 3));
```
stof: Can we use 0 as the min number of tags ? Tags are optional, so it makes sense to have untagged posts too

src/AppBundle/DataFixtures/ORM/LoadPostFixtures.php
```diff
@@ -31,8 +32,9 @@
  *
  * @author Ryan Weaver <weaverryan@gmail.com>
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ * @author Yonel Ceruto <yonelceruto@gmail.com>
  */
-class LoadFixtures extends AbstractFixture implements ContainerAwareInterface
+class LoadPostFixtures extends AbstractFixture implements DependentFixtureInterface, ContainerAwareInterface
```
stof: IMO, you can remove the ``Load`` part of the name, now that there is an entity name

yceruto: Fixed!

src/AppBundle/DataFixtures/ORM/LoadPostFixtures.php
yceruto: Fixed!

yceruto:
Next form tags approach ( coming soon :sweat_smile: )
tagsinput

yceruto:
tagsinput-closeup

Ready!

Status: Need Reviews

src/AppBundle/Form/DataTransformer/ArrayToStringTransformer.php
```diff
@@ -0,0 +1,68 @@
...
+            $names = explode(',', $string);
+        }
+
+        $tags = $this->manager->getRepository(Tag::class)->findBy([
+            'name' => $names,
+        ]);
+
+        $newNames = array_diff($names, $tags);
+        foreach ($newNames as $name) {
+            $tag = new Tag();
```
yceruto: :question: I'm no sure if it's the best place to do that.

src/AppBundle/DataFixtures/ORM/PostFixtures.php
```diff
@@ -157,16 +164,24 @@ private function getPhrases()
...
+            'Diatrias tolerare tanquam noster caesium',
+            'Teres talis orgias saepe tractare de camerarius flavum sensorem',
+            'Silva de secundus galatae demitto quadra',
+            'Sunt accentores vitare salvus flavum parses',
+            'Potus sensim ducunt ad ferox abnoba',
+            'Sunt seculaes transferre talis camerarius fluctuies',
+            'Era brevis ratione est',
+            'Sunt torquises imitari velox mirabilis medicinaes',
+            'Cum mineralis persuadere omnes finises desiderium bi-color',
+            'Bassus fatalis classiss virtualiter transferre de flavum',
```
yceruto: Added 15 phrases more to avoid duplicate post title and slug issue.

src/AppBundle/Form/DataTransformer/ArrayToStringTransformer.php
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * This data transformer is used to translate the array of tags into a comma separated format
+ * that can be displayed and managed by Bootstrap-tagsinput js plugin (and back on submit).
+ *
+ * See http://symfony.com/doc/current/form/data_transformers.html
+ *
+ * @author Yonel Ceruto <yonelceruto@gmail.com>
+ */
+class ArrayToStringTransformer implements DataTransformerInterface
```
voronkovich: @yceruto,  I would name this class like `TagArrayToStringTransformer`, because the current name sounds very common.

src/AppBundle/Form/DataTransformer/ArrayToStringTransformer.php
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public function reverseTransform($string)
+    {
+        if ('' === $string || null === $string) {
+            $names = [];
+        } else {
+            $names = explode(',', $string);
+        }
+
+        $tags = $this->manager->getRepository(Tag::class)->findBy([
```
voronkovich: @yceruto, what about to define tags repository as service and inject it instead of the object manager?

voronkovich:
@yceruto, you forgot to call `EntityManager::persist` for post in `BlogController::editAction` to save the changes in tags.
https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Controller/Admin/BlogController.php#L137

src/AppBundle/Form/DataTransformer/ArrayToStringTransformer.php
yceruto: Yes, it makes sense, thanks!

src/AppBundle/Form/DataTransformer/ArrayToStringTransformer.php
yceruto: I don't see the benefit :/

yceruto:
> @yceruto, you forgot to call EntityManager::persist for post in BlogController::editAction to save the changes in tags.

AFAIK it's not necessary, Doctrine is able to detect the changes on M2M `tags` property and do it. Likewise for new tags through `cascade={"persist"}` option.

src/AppBundle/Form/Type/TagsInputType.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->addModelTransformer(new CollectionToArrayTransformer())
+            ->addViewTransformer(new TagArrayToStringTransformer($this->manager))
```
stof: this should be a model transformer, not a view transformer. The TextType expects its normalized data to be a string. Changing it to be an array of tags could lead to issues (and such issues could appear in a future version of Symfony if some changes are made, even if there is no such issue today).

Changing the type of the normalized data when extending a type is generally not a supported usage of the type.

the code should look like this:

```php
$builder
    ->addViewTransformer(new TagArrayToStringTransformer($this->manager))
    ->addModelTransformer(new CollectionToArrayTransformer())
;
```

web/js/main.js
```diff
@@ -26,6 +26,23 @@
                 close: 'fa fa-remove'
             }
         });
+
+        // Bootstrap-tagsinput initialization
+        // http://bootstrap-tagsinput.github.io/bootstrap-tagsinput/examples/
+        var $input = $('input[data-toggle="tagsinput"]');
+        var source = new Bloodhound({
```
stof: this should be done only when ``$input.length`` is not 0 (i.e. at least one such input has been found on the page)

src/AppBundle/Form/Type/TagsInputType.php
yceruto: You are right, fixed thanks!

yceruto:
@stof, @voronkovich Thanks for the review. I have addressed your comments.

javiereguiluz:
@yceruto I'm really sorry it took me so long to merge this. I finally did it and I'll propose some tweaks in a separate PR. My only regret is that the solution looks too complex. Don't get me wrong, you did a great work and the feature it's fantastically implemented ... but thinking as a potential Symfony user looking at this demo application, I don't like such a complex solution for such a common feature.

yceruto:
@javiereguiluz I would be happy to improve those parts where the solution is complex, please just tell me which ones and I'll try to propose a better one or comment the code for a better understanding of it. Thanks!

javiereguiluz:
@yceruto my comment was more about the complexity imposed by the Form component, such as creating a custom data transformer, enabling two data transformers in the custom form type, calling to the always confusing `->setByReference(false)` inside the form type, etc.

PR #448, opened 2017-02-06
Tweaked the new blog post tags feature

javiereguiluz:
This continues #447 and introduces some minor changes.

---

The only relevant changes for end users is the minor redesign of the tags. For example, in the blog index, before tags stand out too much and they look like buttons:

index-before

Now they are faded because, after all, they are a secondary information:

index-after

In the blog post show, the tags are bigger:

show-after

And in the admin, I haven't modified anything, so they still look like this:

admin-after


src/AppBundle/DataFixtures/ORM/PostFixtures.php
```diff
@@ -21,10 +21,8 @@
 use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 
 /**
- * Defines the sample data to load in the database when running the unit and
- * functional tests.
- *
- * Execute this command to load the data:
+ * Defines the sample blog posts to load in the database before running the unit
+ * and functional tests. Execute this command to load the data:
```
yceruto: CS expect an endpoint here, maybe should be reverted :/

javiereguiluz: Once we update the PHPCsFixer to 2.0, we can disable that rule because `:` should be valid as the ending of a phrase.

PR #450, opened 2017-02-06
Cleanup and comments about Tags feature implementation

yceruto:
* Removed unused old `isAuthor()` method from `Post`
* Removed useless PHPDoc
* Add code comments about Tags feature implementation
* Removed setByReference option which is unnecessary now.

### DIFF
--- a/src/AppBundle/Entity/Post.php
+++ b/src/AppBundle/Entity/Post.php
@@ -186,14 +186,6 @@ class Post
         $this->author = $author;
     }
 
-    /**
-     * Is the given User the author of this Post?
-     */
-    public function isAuthor(User $user = null)
-    {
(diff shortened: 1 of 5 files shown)


yceruto:
Thanks to new instance of `ArrayCollection` into [CollectionToArrayTransformer](https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Doctrine/Form/DataTransformer/CollectionToArrayTransformer.php#L67) now we can remove the `setByReference(false)` option, so the form data mapper should call setter always.

javiereguiluz:
@yceruto thanks for these nice improvements!

yceruto:
@javiereguiluz thanks to you for improving it!

PR #451, opened 2017-02-06
Delete all post/tag entries on delete a Post

yceruto:
We can't use cascade delete constraint in this case because Foreign key support is not enabled in SQLite by default. (https://www.sqlite.org/foreignkeys.html#fk_enable)

More information about it:
- https://github.com/doctrine/dbal/issues/2104
- https://github.com/doctrine/dbal/issues/1204

So my only one solution was to clear this relations manually.

### DIFF
--- a/src/AppBundle/Controller/Admin/BlogController.php
+++ b/src/AppBundle/Controller/Admin/BlogController.php
@@ -177,6 +177,10 @@ class BlogController extends Controller
 
         $entityManager = $this->getDoctrine()->getManager();
 
+        // Delete related symfony_demo_post_tag entries. This is necessary for SQLite platform only,
+        // you can remove it for another platform.
+        $post->getTags()->clear();
+
         $entityManager->remove($post);
         $entityManager->flush();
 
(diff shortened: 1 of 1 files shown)


voronkovich:
@yceruto , IMO it is a platform-specific issue, so we shouldn't fix it in the demo application.

javiereguiluz:
I tend to agree with @voronkovich ... it may be even confusing for newcomers and they might think that they always need to do this.

yceruto:
I'm agree that it's platform-specific issue, but it's the default one, so currently this works wrong for newcomers. We can insert a comment about this then? Something like:
```
// This is necessary for SQLite platform only, you can remove it for another platform.
$post->getTags()->clear();
```
What do you think?

yceruto:
I'm also thinking when we need to test the Post "Delete/New" actions, if won't fix that, then the SQLite DB will have an invalid state.

javiereguiluz:
Let's merge this with the help note. Thanks @yceruto!

PR #452, opened 2017-02-07
Improved the event subscriber that listens to database exceptions

javiereguiluz:
This finishes #431 by doing some renaming and minor tweaks.

### DIFF
--- a/src/AppBundle/EventListener/CheckRequirementsSubscriber.php
+++ b/src/AppBundle/EventListener/CheckRequirementsSubscriber.php
@@ -28,16 +28,12 @@ use Symfony\Component\HttpKernel\KernelEvents;
  *
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  */
-class CheckSQLiteEventSubscriber implements EventSubscriberInterface
+class CheckRequirementsSubscriber implements EventSubscriberInterface
 {
-    /**
-     * @var EntityManager
-     */
+    /** @var EntityManager */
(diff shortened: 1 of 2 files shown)


src/AppBundle/EventListener/CheckRequirementsSubscriber.php
```diff
@@ -54,9 +50,9 @@ public static function getSubscribedEvents()
         return [
             // Exceptions are one of the events defined by the Console. See the
             // rest here: http://symfony.com/doc/current/components/console/events.html
-            ConsoleEvents::EXCEPTION => 'handleDatabaseExceptions',
-            // See: http://api.symfony.com/3.2/Symfony/Component/HttpKernel/KernelEvents.html
-            KernelEvents::EXCEPTION => 'onKernelException',
+            ConsoleEvents::EXCEPTION => 'handleConsoleException',
+            // See: http://api.symfony.com/master/Symfony/Component/HttpKernel/KernelEvents.html
```
stof: can we use ``current`` on the API doc or no ?

stof:
👍

src/AppBundle/EventListener/CheckRequirementsSubscriber.php
javiereguiluz: It was my first try, but ...

api_current

stof: then fine with using master

PR #454, opened 2017-02-07
Updated PHP CS Fixer to 2.0

javiereguiluz:
This finishes #407.


.php_cs
```diff
@@ -1,29 +1,28 @@
...
     ->exclude('var')
     ->exclude('vendor')
-    ->exclude('web/bundles')
-    ->exclude('web/css')
-    ->exclude('web/fonts')
-    ->exclude('web/js')
     ->notPath('web/config.php')
-    ->in(__DIR__)
+    // Doctrine fixtures are excluded to avoid these PHP-CS-Fixer errors:
+    // "Files were not fixed due to errors reported during linting before fixing"
```
keradus: why doctrine fixtures are invalid PHP code ?

.php_cs
```diff
@@ -1,29 +1,28 @@
...
+    ->setRules([
+        '@Symfony' => true,
+        '@Symfony:risky' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'no_useless_else' => true,
+        'no_useless_return' => true,
+        'ordered_imports' => true,
+        'phpdoc_order' => true,
+        'php_unit_strict' => true,
+        'strict_comparison' => true,
```
keradus: I am totally for this rule, but it's not from Symfony ruleset. Why include it in demo ? That would mean anyone would include here anything with time passing, right ? I would either keep Symfony rulesets only or propose those changes to be included in the ruleset in the first place.

Problematic parts are rules like `ordered_imports`, which Symfony wants to follow in new code, but don't want to apply it for whole codebase to avoid merging conflicts

composer.lock
```diff
@@ -73,7 +73,7 @@
                 "docblock",
                 "parser"
             ],
-            "time": "2015-08-31 12:32:49"
+            "time": "2015-08-31T12:32:49+00:00"
```
keradus: btw, is the lock file really needed for demo project ?

javiereguiluz:
@keradus in 0aa68ef I've removed everything except the default Symfony rules. Is that what you were proposing?

composer.lock
stof: I would say yes, as it showcases best practices for projects, and it ensures that people cloning the demo gets working dependencies, which is a better experience for a first try

.php_cs
```diff
@@ -1,29 +1,22 @@
...
-    ->exclude('app/data')
-    ->exclude('app/Resources')
+$finder = PhpCsFixer\Finder::create()
+    ->in(__DIR__)
     ->exclude('var')
     ->exclude('vendor')
-    ->exclude('web/bundles')
-    ->exclude('web/css')
-    ->exclude('web/fonts')
-    ->exclude('web/js')
```
stof: I would keep excluding them (no need to look for them to discover they are unsupported, while we can skip entering these folders entirely)

.php_cs
stof: IMO, we must indeed rather fix the issue, which is a bad git conflict resolution: https://github.com/symfony/symfony-demo/blob/eba7c9f21bbd2dca312c6f767b77aafc60d47d33/src/AppBundle/DataFixtures/ORM/PostFixtures.php#L80

.php_cs
```diff
@@ -1,29 +1,22 @@
...
-$finder = Symfony\CS\Finder::create()
-    ->ignoreDotFiles(true)
-    ->ignoreVCS(true)
-    ->exclude('app/config')
-    ->exclude('app/data')
-    ->exclude('app/Resources')
+$finder = PhpCsFixer\Finder::create()
+    ->in(__DIR__)
     ->exclude('var')
     ->exclude('vendor')
```
stof: this is already excluded by ``PhpCsFixer\Finder::create()``

composer.lock
keradus: the issue with it that usually the user won't update it, which mean he will use outdated once

keradus:
> @keradus in 0aa68ef I've removed everything except the default Symfony rules. Is that what you were proposing?

no, i proposed to evaluate the rules and maybe include them into `@Symfony` ruleset itself.
or, if Symfony project will refuse to apply the rules on codebase but expect them for new code, create sth like `@SymfonyWishList` (with better name) to track that kind of rules that should apply for new code only.

src/AppBundle/DataFixtures/ORM/PostFixtures.php
```diff
@@ -107,8 +106,6 @@ private function addRandomTags(Post $post)
         }
     }
 
-=======
->>>>>>> pull/431:src/AppBundle/DataFixtures/ORM/LoadFixtures.php
```
keradus: btw, this conflict could be resolved in standalone PR ;)
to keep PRs atomic and clean

.php_cs
keradus: also, sadly, I met few JS packages that are distributed with... **PHP** files (as a demo usage)

composer.lock
javiereguiluz: @keradus you are right, butt keep in mind that this demo application is a bit special. First, we download it as a ZIP file with all the dependencies installed ... so 99% of users never use Composer with this app. Second, the demo app is designed as a temporary utility (to learn Symfony, to showcase some feature, to do a quick benchmark) so you usually install the demo and delete it pretty quickly ... and then download it again. So 99% of people won't get old dependencies or don't care about them.

stof: thus, if people want an updated version of the demo, they should not only update the vendors, but also the demo itself (and so they would get our new lock file and run ``composer install``)

.php_cs
stof: well, Symfony itself decided to order new use statements only, due to maintenance constraints. But this does not forbid this demo project to be strict about them.
The same is true for the short array syntax. This demo decided to use it.

regarding ``no_useless_else``, if it does what I think it does according to its name, it should be part of the ruleset, as Symfony indeed mandates that we don't use ``else`` when the ``if`` quits the execution path (through ``return``, ``throw``, ``continue`` or ``break``

composer.lock
keradus: ok, thus :+1:
(on a side not, that approach would require this demo project to update lock file on regular basis)

keradus:
great to see this happen ;)

PR #455, opened 2017-02-07
Order tags by ascending

voronkovich:


javiereguiluz:
@voronkovich nice! A small change ... but a very useful feature!

PR #457, opened 2017-02-08
Add fullName property to User entity

yceruto:
Implements https://github.com/symfony/symfony-demo/issues/446

Others changes:
* Show full author name in all templates (https://github.com/symfony/symfony-demo/pull/439#discussion_r96436908)
* Use "Jane Doe" instead of "Anna" [RFC] !
* Updated fixtures and user commands.

| Posts Index | Post Comments |
| --- | --- |
| index | comments |


javiereguiluz:
Looks nice!   Just asking: why change the name of Anna to Jane ?

yceruto:
>why change the name of Anna to Jane ?

I haven't preference really, but researching a little about that, "Jane Doe" or "Jane Roe" for women are widely used more than "Anna" which is not a common placeholder name. But I'm fine with "Anna Doe" however if you prefer.

Refs: https://en.wikipedia.org/wiki/John_Doe, https://en.wikipedia.org/wiki/Placeholder_name

javiereguiluz:
@yceruto thanks for contributing this feature. It was on my TODO list sing long ago ... so I'm glad this is finally merged!

PR #458, opened 2017-02-09
An example to improved command as service

RomainMarecat:
There is no reason to extends any command class with ContainerAwareCommand.

### DIFF
--- a/src/AppBundle/Command/AddUserCommand.php
+++ b/src/AppBundle/Command/AddUserCommand.php
@@ -13,12 +13,13 @@ namespace AppBundle\Command;
 
 use AppBundle\Entity\User;
 use Doctrine\Common\Persistence\ObjectManager;
-use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\Question;
+use Symfony\Component\Security\Core\Encoder\UserPasswordEncoder;
 
 /**
  * A command console that creates users and stores them in the database.
@@ -36,15 +37,21 @@ use Symfony\Component\Console\Question\Question;
  *
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  * @author Yonel Ceruto <yonelceruto@gmail.com>
+ * @author Romain Marecat <romain.marecat@gmail.com>
  */
-class AddUserCommand extends ContainerAwareCommand
+class AddUserCommand extends Command
 {
     const MAX_ATTEMPTS = 5;
 
     /**
-     * @var ObjectManager
+     * @var ObjectManager $entityManager
      */
     private $entityManager;
+
+    /**
+     * @var UserPasswordEncoder $passwordEncoder
+     */
+    private $passwordEncoder;
 
     /**
      * {@inheritdoc}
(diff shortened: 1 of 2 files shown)


uikolas:
Whats wrong with extending?

RomainMarecat:
Because abstract ContainerAwareCommand class extends Command class. ContainerAwareCommand just instanciate container attribute. And with DIC, why find Container ??.
[`](https://symfony.com/doc/current/components/dependency_injection.html)

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -298,4 +292,66 @@ private function getCommandHelp()
...
+
+        return $this;
+    }
+
+    /**
+     * Gets the value of passwordEncoder.
+     *
+     * @return UserPasswordEncoder $passwordEncoder
+     */
+    public function getPasswordEncoder()
```
stof: Adding a public getter for the dependency of the command does not make any sense. There is no reason for a service to expose its dependencies to the outside

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -298,4 +292,66 @@ private function getCommandHelp()
...
+    }
+
+    /**
+     * Sets the value of entityManager.
+     *
+     * @param ObjectManager $entityManager $entityManager the entity manager
+     *
+     * @return self
+     */
+    private function _setEntityManager($entityManager)
```
stof: why such method with an _ name ?

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -298,4 +292,66 @@ private function getCommandHelp()
 
 HELP;
     }
+
+    /**
+     * Gets the value of entityManager.
+     *
+     * @return ObjectManager
+     */
+    public function getEntityManager()
```
stof: should not be there

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -36,17 +37,23 @@
...
+ * @author Romain Marecat <romain.marecat@gmail.com>
  */
-class AddUserCommand extends ContainerAwareCommand
+class AddUserCommand extends Command
 {
     const MAX_ATTEMPTS = 5;
 
     /**
-     * @var ObjectManager
+     * @var ObjectManager $entityManager
```
stof: addig the property name does not make sense. We are documenting a property definition here

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -66,19 +73,6 @@ protected function configure()
     }
 
     /**
-     * This method is executed before the interact() and the execute() methods.
-     * It's main purpose is to initialize the variables used in the rest of the
-     * command methods.
-     *
-     * Beware that the input options and arguments are validated after executing
-     * the interact() method, so you can't blindly trust their values in this method.
```
stof: you should copy this comment in the DeleteUserCommand then, otherwise we don't have the complete explanation about ``interact`` anymore

src/AppBundle/Command/AddUserCommand.php
```diff
@@ -298,4 +292,66 @@ private function getCommandHelp()
...
+    }
+
+    /**
+     * Sets the value of entityManager.
+     *
+     * @param ObjectManager $entityManager the entity manager
+     *
+     * @return self
+     */
+    public function setEntityManager(ObjectManager $entityManager)
```
stof: IMO, you should use constructor injection here (or the new experimental getter overriding once migrating to 3.3 if we want to keep the dependency lazy, as this is a good use case)

RomainMarecat:
see the difference with memory usage when you get Container with master and after with my PR 

on execute() function 

me  :
int(10485760)

master:
int(18874368)

stof:
I don't understand what you measure here

javiereguiluz:
I'm not a big fan of making this change. It complicates everything too much :(

javiereguiluz:
@RomainMarecat thanks for your contribution and for helping us improve the Symfony Demo app. However, I must close your pull request as "won't fix". Don't get me wrong: your proposal would be perfectly fine in any other app, but for this particular demo, we think it's too advanced.

That's why we prefer to keep using the normal Symfony commands and then add a mention to "commands as services" (see #475). I hope you understand this decision. Thanks!

PR #460, opened 2017-02-12
Reduce DB query number in post index action

yceruto:
**Performance Issue**

After tags feature the blog index page performs one extra query for each post to fetches its related tags. This PR avoids that.

Other changes:
* Refactoring the Post repository methods to reuse paginator creation in the future.

### DIFF
--- a/src/AppBundle/Repository/PostRepository.php
+++ b/src/AppBundle/Repository/PostRepository.php
@@ -52,7 +52,7 @@ class PostRepository extends EntityRepository
 
     private function createPaginator(Query $query, $page)
     {
-        $paginator = new Pagerfanta(new DoctrineORMAdapter($query, false));
+        $paginator = new Pagerfanta(new DoctrineORMAdapter($query));
         $paginator->setMaxPerPage(Post::NUM_ITEMS);
         $paginator->setCurrentPage($page);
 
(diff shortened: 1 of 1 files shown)


Grafikart:
I noticed this issue too but I'm not sure avout this fix since the LIMIT is then applied to the tags. If the first post has 10 tags, then it will be the only one in the listing right ?

javiereguiluz:
@Grafikart I'm not sure I understand your comment. This is how join queries are supposed to work in Doctrine. I always use a code similar to this one. I've just tested this PR locally and it seems to be working as expected. The results are the same, but the database queries:

| Before | After
| --- | ---
| db-before | db-after

javiereguiluz:
Thanks Yonel.

Grafikart:
@javiereguiluz sorry it works with PagerFanta indeed. It didn't worked for me cause I tried this PR without the paginator :  

```
        $query = $this->getEntityManager()
            ->createQuery('
                SELECT p, t
                FROM AppBundle:Post p
                LEFT JOIN p.tags t
                WHERE p.publishedAt <= :now
                ORDER BY p.publishedAt DESC
            ')
            ->setMaxResults(5)
            ->setParameter('now', new \DateTime())
```

But doing a first query to get the posts ids (like PagerFanta does) solve this issue. Sorry for the false issue.

PR #461, opened 2017-02-12
Minor refactor to new comment action

yceruto:
I couldn't find a text where this was explicitly mentioned, but the code above seems unconventional with respect to what the documentation shows and the rest of the application code.

Refs:
* http://symfony.com/doc/current/best_practices/forms.html#building-forms
* http://symfony.com/doc/current/best_practices/forms.html#handling-form-submits
* http://symfony.com/doc/current/forms.html#building-the-form
* http://symfony.com/doc/current/forms.html#handling-form-submissions
* https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/Controller/Admin/BlogController.php#L75-L93

### DIFF
--- a/src/AppBundle/Controller/BlogController.php
+++ b/src/AppBundle/Controller/BlogController.php
@@ -92,16 +92,15 @@ class BlogController extends Controller
      */
     public function commentNewAction(Request $request, Post $post)
     {
-        $form = $this->createForm(CommentType::class);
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I personally use this style too in my apps ... so I like this proposal! Thanks @yceruto.

PR #463, opened 2017-02-13
Save CS Fixer's cache file into the "var/" directory

bocharsky-bw:
Otherwise, my project leaves in dirty state after running the fixer.


javiereguiluz:
@bocharsky-bw much better! Thanks.

PR #466, opened 2017-02-13
Fix translating "limit" variable: add extra spaces

bocharsky-bw:
`{{limit}}` not translated at all.

### DIFF
--- a/app/Resources/translations/validators.uk.xlf
+++ b/app/Resources/translations/validators.uk.xlf
@@ -12,7 +12,7 @@
             </trans-unit>
             <trans-unit id="post.too_short_content">
                 <source>post.too_short_content</source>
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@bocharsky-bw good catch! I was wondering if this could be added to the proposed Xliff linter (https://github.com/symfony/symfony/pull/21578) but since the placeholder wrapper is configurable (it can be {{ ... }}, or % ... %, etc.) this won't be easy .... or even impossible!

stof:
> (it can be {{ ... }}, or % ... %, etc.) this won't be easy .... or even impossible!

you don't even need a delimiter (even though it would make maintaining translations a huge pain). Symfony just replaces the string passed a placeholder, without caring about its content. You could use ``$count``

PR #467, opened 2017-02-14
Drop support for PHP 5.3 callbacks

javiereguiluz:
We now require PHP 5.5.9 in this app, so there's no need to maintain this code.


stof:
this is broken. Private functions can be used as callable when passing them to built-in functions like ``array_map``, which are special. But here, it is passed to a userland function. So ``array($command, 'fullNameValidator')`` must be callable from the userland scope calling the callable, which require them to be public.

The comment about people public for PHP 5.3 only was outdated (it may have true for a previous implementation not using the Question system)

PR #469, opened 2017-02-14
Fixed typo in polish translations

Wirone:
There is nothing like "zademonostrowania". It is like english `demonstrate` vs `demonostrate`.


javiereguiluz:
@Wirone nice catch! *Dziękuję bardzo*

PR #470, opened 2017-02-14
Improve fixture generation

javiereguiluz:
* Add at least 1 tag to every blog post, to make the design more consistent
* Don't cut out the phrases of the post summary.


src/AppBundle/DataFixtures/ORM/PostFixtures.php
```diff
@@ -91,11 +91,7 @@ public function getDependencies()
 
     private function addRandomTags(Post $post)
     {
-        if (0 === $count = mt_rand(0, 3)) {
```
stof: this means we don't have any post without tags in the fixtures. Is it expected ? Tagging posts is optional

stof:
> Add at least 1 tag to every blog post, to make the design more consistent

The validation of posts does not require having at least 1 tag. So if the design does not support such case, it means that there is a bug in the code, not that fixtures should be changed

javiereguiluz:
I've changed the fixtures again: posts can have 0 tags ... but the first post now always has 2 tags, to make the application look better.

stof:
👍

PR #471, opened 2017-02-14
Reverted #467 to make PHP callbacks work again

javiereguiluz:
As explained by @stof in https://github.com/symfony/symfony-demo/pull/467#issuecomment-279680926 this change was wrong.

### DIFF
--- a/src/AppBundle/Command/AddUserCommand.php
+++ b/src/AppBundle/Command/AddUserCommand.php
@@ -223,8 +223,6 @@ class AddUserCommand extends ContainerAwareCommand
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/AppBundle/Command/AddUserCommand.php
```diff
@@ -222,7 +222,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
         }
     }
 
-    private function passwordValidator($plainPassword)
+    /**
+     * This internal method should be private, but PHP requires callbacks to be declared public.
```
stof: I don't think it should be private, as you pass it to something else.

If you want to make it private, use a closure as actual validator though

PR #474, opened 2017-02-14
Remove unused tag

Grafikart:
When a tag is detached from a post it stays in the Tag table filling the table over time. When we update (or delete a post) we clean up unused (unlinked tags)

## Improvments ideas 💡 

- I wanted to add this behaviour in an EventSubscriber but I don't know how to detect when a tag relation is persisted
- The query to get unused tag could be improved avoiding the double LEFT JOIN, but I don't know how to ask doctrine to limit the LEFT JOIN to the relation table.

### DIFF
--- a/src/AppBundle/Controller/Admin/BlogController.php
+++ b/src/AppBundle/Controller/Admin/BlogController.php
@@ -12,6 +12,7 @@
 namespace AppBundle\Controller\Admin;
 
 use AppBundle\Entity\Post;
+use AppBundle\Entity\Tag;
 use AppBundle\Form\PostType;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
@@ -147,6 +148,7 @@ class BlogController extends Controller
         if ($form->isSubmitted() && $form->isValid()) {
             $post->setSlug($this->get('slugger')->slugify($post->getTitle()));
             $entityManager->flush();
+            $this->getDoctrine()->getRepository(Tag::class)->cleanUnusedTags();
 
             $this->addFlash('success', 'post.updated_successfully');
 
@@ -184,6 +186,7 @@ class BlogController extends Controller
 
         $entityManager->remove($post);
         $entityManager->flush();
+        $this->getDoctrine()->getRepository(Tag::class)->cleanUnusedTags();
 
         $this->addFlash('success', 'post.deleted_successfully');
 
--- a/src/AppBundle/Controller/Admin/BlogController.php
+++ b/src/AppBundle/Controller/Admin/BlogController.php
@@ -147,8 +147,10 @@ class BlogController extends Controller
 
         if ($form->isSubmitted() && $form->isValid()) {
             $post->setSlug($this->get('slugger')->slugify($post->getTitle()));
-            $entityManager->flush();
-            $this->getDoctrine()->getRepository(Tag::class)->cleanUnusedTags();
+            $entityManager->transactional(function ($entityManager) {
+                $entityManager->flush();
+                $this->getDoctrine()->getRepository(Tag::class)->cleanUnusedTags();
+            });
 
             $this->addFlash('success', 'post.updated_successfully');
 
@@ -178,15 +180,14 @@ class BlogController extends Controller
         }
 
         $entityManager = $this->getDoctrine()->getManager();
-
-        // Delete the tags associated with this blog post. This is done automatically
-        // by Doctrine, except for SQLite (the database used in this application)
-        // because foreign key support is not enabled by default in SQLite
-        $post->getTags()->clear();
-
-        $entityManager->remove($post);
-        $entityManager->flush();
-        $this->getDoctrine()->getRepository(Tag::class)->cleanUnusedTags();
+        $entityManager->transactional(function ($entityManager) use ($post) {
+            // Delete the tags associated with this blog post. This is done automatically
+            // by Doctrine, except for SQLite (the database used in this application)
+            // because foreign key support is not enabled by default in SQLite
+            $post->getTags()->clear();
+            $entityManager->remove($post);
+            $this->getDoctrine()->getRepository(Tag::class)->cleanUnusedTags();
+        });
 
         $this->addFlash('success', 'post.deleted_successfully');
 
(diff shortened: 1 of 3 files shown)


src/AppBundle/Entity/Tag.php
```diff
@@ -26,6 +26,13 @@ class Tag implements \JsonSerializable
     private $id;
 
     /**
+     * @var array
+     *
+     * @ORM\ManyToMany(targetEntity="AppBundle\Entity\Post", mappedBy="tags")
+     */
+    private $posts;
```
voronkovich: I agree with @yceruto.  There is a Doctrine best practice:

> Avoid bidirectional relationships. 

http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/best-practices.html#constrain-relationships-as-much-as-possible

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -147,6 +148,7 @@ public function editAction(Post $post, Request $request)
         if ($form->isSubmitted() && $form->isValid()) {
             $post->setSlug($this->get('slugger')->slugify($post->getTitle()));
             $entityManager->flush();
+            $this->getDoctrine()->getRepository(Tag::class)->cleanUnusedTags();
```
Grafikart: Personal choice mainly (I understand if you close the issue).

Having tags attached to nothing could create some problem if you choose to create a tagcloud later for instance.

src/AppBundle/Entity/Tag.php
Grafikart: I put this relation to make the query in the TagRepository

voronkovich: @Grafikart, you can delete unused tags using just one SQL query.

stof: and even in DQL, you **don't** need to have a relation to be able to do a join

src/AppBundle/Controller/Admin/BlogController.php
stof: as you do it using a separate query, you should wrap the flush and this query in a transaction, to be atomic.

src/AppBundle/Entity/Tag.php
```diff
@@ -66,4 +73,45 @@ public function __toString()
...
+    }
+
+    /**
+     * Add post
+     *
+     * @param \AppBundle\Entity\Post $post
+     *
+     * @return Tag
+     */
+    public function addPost(\AppBundle\Entity\Post $post)
```
stof: this adder is worse than anything else you did, because it will **not** add the post (as you don't update the owning side of the relation)

src/AppBundle/Repository/TagRepository.php
```diff
@@ -0,0 +1,56 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace AppBundle\Repository;
+
+use AppBundle\Entity\Tag;
+use Doctrine\ORM\EntityRepository;
+use Doctrine\ORM\Query;
+use Pagerfanta\Adapter\DoctrineORMAdapter;
```
stof: why pagerfanta ?

src/AppBundle/Repository/TagRepository.php
```diff
@@ -0,0 +1,56 @@
...
+use Pagerfanta\Pagerfanta;
+
+/**
+ * This custom Doctrine repository contains some methods which are useful when
+ * querying for blog post information.
+ *
+ * See http://symfony.com/doc/current/book/doctrine.html#custom-repository-classes
+ *
+ * @author Ryan Weaver <weaverryan@gmail.com>
+ * @author Javier Eguiluz <javier.eguiluz@gmail.com>
```
stof: authors look wrong here

src/AppBundle/Controller/Admin/BlogController.php
stof: you could make this code much more efficient, by getting the old list of tags of the Post in an array before ``handleRequest``, then comparing it to the list of new tags here. The difference will give you the list of tags removed from the post, which are the only ones you need to check for being unused

src/AppBundle/Entity/Tag.php
Grafikart: Removing everything from this entity sorry

src/AppBundle/Repository/TagRepository.php
Grafikart: Typo sorry

src/AppBundle/Controller/Admin/BlogController.php
Grafikart: Agreed, but it would add more code to the controller, is it acceptable ?

Grafikart:
I can't figure how to do this query using DQL https://github.com/symfony/symfony-demo/pull/474/files#diff-2629962eca7be81a3484a3f64ad18074R36 (can't find a way to do a join on the join table without having a ManyToMany between Tag and Post owned by Tag)

voronkovich:
@Grafikart, see http://stackoverflow.com/questions/21513086/how-to-join-without-relational-table-in-symfony-doctrine-with-querybuilder-betwe

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -146,7 +147,10 @@ public function editAction(Post $post, Request $request)
 
         if ($form->isSubmitted() && $form->isValid()) {
             $post->setSlug($this->get('slugger')->slugify($post->getTitle()));
-            $entityManager->flush();
+            $entityManager->transactional(function ($entityManager) {
```
javiereguiluz: I appreciate your contribution ... but using `$entityManager->transactional` is "light years" ahead of what we want to teach in the Symfony Demo app. I'm sorry but we can't use this obscure feature. We'll need to find a much simpler alternative solution. Thanks for understanding it.

src/AppBundle/Repository/TagRepository.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    /**
+     * Remove unused Tag from the Database.
+     */
+    public function cleanUnusedTags()
+    {
+        $joinTable = $this->getEntityManager()->getClassMetadata(Post::class)->getAssociationMapping('tags')['joinTable']['name'];
+        $tagTable = $this->getClassMetadata()->getTableName();
+        $unused_tags = $this->getEntityManager()->createNativeQuery("
+            DELETE FROM `$tagTable` WHERE id IN
```
voronkovich: @Grafikart , try to use:
```sql
DELETE FROM tags t
LEFT JOIN tags_posts j ON j.tag_id = t.id
WHERE j.tag_id IS NULL
```

PR #478, opened 2017-02-17
Add more posts authors

voronkovich:
For now we have only one admin: Jane Doe. Latest posts with more than one author look more naturally.


tests/AppBundle/Controller/Admin/BlogControllerTest.php
```diff
@@ -84,9 +84,9 @@ public function testBackendHomepage()
 
         $crawler = $client->request('GET', '/en/admin/post/');
 
-        $this->assertCount(
-            30,
-            $crawler->filter('body#admin_post_index #main tbody tr'),
+        $this->assertGreaterThanOrEqual(
+            1,
+            $crawler->filter('body#admin_post_index #main tbody tr')->count(),
```
yceruto: Since authors are randomly assigned, each time the fixtures are loaded this test would fail (depending on the authenticated user)

src/AppBundle/Repository/PostRepository.php
```diff
@@ -38,8 +38,9 @@ public function findLatest($page = 1)
     {
         $query = $this->getEntityManager()
             ->createQuery('
-                SELECT p, t
+                SELECT p, a, t
                 FROM AppBundle:Post p
+                JOIN p.author a
```
yceruto: AFAIK this shouldn't improved nothing, Doctrine hydrates the author without an extra query.

tests/AppBundle/Controller/Admin/BlogControllerTest.php
voronkovich: @yceruto, Jane Doe has at least one post, so test should pass.

src/AppBundle/Repository/PostRepository.php
voronkovich: @yceruto, I think you're wrong. See http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/dql-doctrine-query-language.html#joins

> A join (be it an inner or outer join) becomes a “fetch join” as soon as fields of the joined entity appear in the SELECT part of the DQL query outside of an aggregate function.

tests/AppBundle/Controller/Admin/BlogControllerTest.php
yceruto: Ooh! `GreaterThanOrEqual` you're right, this answers my first comment too (`Why do you need that the first Post belongs to Jane`)

src/AppBundle/Repository/PostRepository.php
yceruto: Well, I think it's necessary for multiple users then :+1:

20uf:
Why not just use https://github.com/nelmio/alice.
It would alleviate the code above and make it all more understandable.

javiereguiluz:
I wouldn't add the Alice bundle for such a simple demo application as this one. We want to showcase the main Symfony features ... but not turn this into a too large and too slow app.

javiereguiluz:
It took me a long time, but this is finally merged! Thanks voronkovich

PR #479, opened 2017-02-17
Inject a tag repository into TagsInputType instead of an entity manager

voronkovich:


javiereguiluz:
I'm not sure about this proposal because traditionally it has been recommended to not create services for Doctrine repositories. I'll wait for more comments. Thanks!

app/config/services.yml
```diff
@@ -66,18 +66,18 @@ services:
...
-    # Uncomment the following lines to define a service for the Post Doctrine repository.
+    # Defines a service for the Tag Doctrine repository.
     # It's not mandatory to create these services, but if you use repositories a lot,
     # these services simplify your code:
     #
-    # app.post_repository:
-    #     class:     Doctrine\ORM\EntityRepository
-    #     factory:   ['@doctrine.orm.entity_manager', getRepository]
-    #     arguments: [AppBundle\Entity\Post]
+    app.tag_repository:
```
stof: I'm -1 on this. Defining the Doctrine repositories as services can actually create issues when needing to reset the entity manager (recent versions of Symfony and DoctrineBundle solved this issue for the case of injecting the EntityManager itself when lazy services are supported, but this cannot be solved for repositories, especially given that DoctrineBundle does not know about these services).

Thus, getting the repository triggers the loading of the metadata of the class. So depending where you inject the service, it may cause issues, by loading metadata when you don't expect them to be loaded. Btw, if your DBAL connection uses a version-aware driver (MySQL, Postgresql, SQL Server, etc...) and relies on guessing the server version rather than configuring it explicitly, loading metadata will requires connecting to the database (to guess the platform version), which may break things if a service instantiation leads to this very early (or during cache clearing , as happened in the past to some people).

Injecting the EntityManager and keeping the Doctrine instantiation lazy is safer. (this is fine in the code currently just because the only place using it is a form type which is loaded lazily only when needed)

stof: thus, defining repositories as services encourages people to inject extra dependencies in it, without realizing that Doctrine will not use the container to get the repository, and so may give you the instance without the extra injections being done (which would be a total nightmare to debug, and very hard to avoid if you don't want to force instantiating all repositories in each request)

voronkovich: @stof, Thanks for the explanation! So, it's better to close this PR.

PR #481, opened 2017-02-18
Add more tests to blog post

yceruto:
* Testing new blog post
* Testing show blog post (admin)
* Testing add a comment to post
* Changing some `PostFixtures` methods to static for reusability
* Simplify and reordering some tests

### DIFF
--- a/src/AppBundle/DataFixtures/ORM/PostFixtures.php
+++ b/src/AppBundle/DataFixtures/ORM/PostFixtures.php
@@ -13,12 +13,12 @@ namespace AppBundle\DataFixtures\ORM;
 
 use AppBundle\Entity\Comment;
 use AppBundle\Entity\Post;
-use AppBundle\Entity\Tag;
 use Doctrine\Common\DataFixtures\AbstractFixture;
 use Doctrine\Common\DataFixtures\DependentFixtureInterface;
 use Doctrine\Common\Persistence\ObjectManager;
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+use Tests\FixturesTrait;
 
 /**
  * Defines the sample blog posts to load in the database before running the unit
@@ -35,6 +35,7 @@ use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 class PostFixtures extends AbstractFixture implements DependentFixtureInterface, ContainerAwareInterface
 {
     use ContainerAwareTrait;
+    use FixturesTrait;
 
     /**
      * {@inheritdoc}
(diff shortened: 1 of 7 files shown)


src/AppBundle/DataFixtures/ORM/PostFixtures.php
```diff
@@ -184,9 +183,9 @@ private function getPhrases()
         ];
     }
 
-    private function getRandomPostTiles()
+    public static function getRandomPostTiles()
```
voronkovich: @yceruto,  there is a typo in the method's name here. Should be `getRandomPostTitles`.

yceruto: Thanks, Fixed!

tests/AppBundle/Controller/BlogControllerTest.php
```diff
@@ -54,4 +55,38 @@ public function testRss()
...
+     */
+    public function testNewComment()
+    {
+        $client = self::createClient([], [
+            'PHP_AUTH_USER' => 'john_user',
+            'PHP_AUTH_PW' => 'kitten',
+        ]);
+
+        /** @var Post $post */
+        $post = $client->getContainer()->get('doctrine')->getRepository(Post::class)->find(1);
```
voronkovich: @yceruto,  you could use an `EntityManager#find` method:
```php
 $post = $client->getContainer()->get('doctrine')->find(Post::class, 1);
```

yceruto: Sorry your suggestion does not works.

voronkovich: @yceruto, Sorry for confusing you! I mixed up the `doctrine` service and an entity manager. :(

tests/AppBundle/Controller/Admin/BlogControllerTest.php
```diff
@@ -35,31 +35,31 @@ class BlogControllerTest extends WebTestCase
...
     {
-        $client = static::createClient([], [
+        $client = self::createClient([], [
             'PHP_AUTH_USER' => 'john_user',
             'PHP_AUTH_PW' => 'kitten',
         ]);
 
         $client->request($httpMethod, $url);
-        $this->assertSame($statusCode, $client->getResponse()->getStatusCode());
+        $this->assertTrue($client->getResponse()->isForbidden());
```
bocharsky-bw: You made it less strict, but I don't think it makes sense in tests.

tests/AppBundle/Controller/Admin/BlogControllerTest.php
```diff
@@ -70,7 +70,7 @@ public function testAdminBackendHomePage()
         $client = $this->getAdminClient();
 
         $crawler = $client->request('GET', '/en/admin/post/');
-        $this->assertSame(Response::HTTP_OK, $client->getResponse()->getStatusCode());
+        $this->assertTrue($client->getResponse()->isSuccessful());
```
bocharsky-bw: The same here, I think we want to know that status code is exactly 200 OK

src/AppBundle/DataFixtures/ORM/PostFixtures.php
```diff
@@ -108,7 +107,7 @@ private function getRandomTags($numTags = 0)
         return $tags;
     }
 
-    private function getPostContent()
+    public static function getPostContent()
```
bocharsky-bw: 👎 I'd rather suggest to use PHP Traits instead of making `private` methods `public static` or do simple inheritance. We should always avoid public methods which we have to care about later.

tests/AppBundle/Controller/Admin/BlogControllerTest.php
```diff
@@ -80,21 +80,42 @@ public function testAdminBackendHomePage()
...
+        ]);
+        $client->submit($form);
 
-        $this->assertSame(Response::HTTP_FOUND, $client->getResponse()->getStatusCode());
+        $this->assertTrue($client->getResponse()->isRedirect());
 
-        $post = $client->getContainer()->get('doctrine')->getRepository(Post::class)->find(1);
-        $this->assertNull($post);
+        /** @var Post $post */
+        $post = $client->getContainer()->get('doctrine')->getRepository(Post::class)->find(31);
```
javiereguiluz: The `31` id should never fail ... but maybe a `findByTitle($postTitle)` would make the code more robust?

tests/AppBundle/Controller/Admin/BlogControllerTest.php
javiereguiluz: In this case, `isForbidden()` only checks the `403` status code ... so the code it's technically equivalent...

tests/AppBundle/Controller/Admin/BlogControllerTest.php
javiereguiluz: ... however, here the `isSuccessful()` method is too liberal:

```php
    public function isSuccessful()
    {
        return $this->statusCode >= 200 && $this->statusCode < 300;
    }
```

I agree with @bocharsky-bw and I think we should revert this change and go back to `Response::HTTP_OK`

src/AppBundle/DataFixtures/ORM/PostFixtures.php
yceruto: :+1: for Trait

tests/AppBundle/Controller/Admin/BlogControllerTest.php
yceruto: :+1: sure

tests/AppBundle/Controller/Admin/BlogControllerTest.php
yceruto: Yes, in this case it's technically equivalent.

tests/AppBundle/Controller/Admin/BlogControllerTest.php
yceruto: :+1: reverting....

yceruto:
@javiereguiluz @bocharsky-bw done!

* Now the fixtures and tests using a `FixturesTrait` for common methods.
* and all controller test case now extending from `ControllerTestCase` to use common decriptive methods and strict assertions.

Status: Need Reviews :)

tests/ControllerTestCase.php
```diff
@@ -0,0 +1,67 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Tests;
+
+use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+use Symfony\Component\HttpFoundation\Response;
+
+class ControllerTestCase extends WebTestCase
```
yceruto: Missing description

tests/FixturesTrait.php
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Tests;
+
+trait FixturesTrait
```
yceruto: Missing description

javiereguiluz:
I'm not that sure about `tests/ControllerTestCase.php`. For experienced Symfony developers makes sense ... but for newcomers? They probably see that and think ... *"mmm, it's pretty complex to create just one simple functional test ... first I need to create this special ControllerTestCase, then I need to create my tests, etc."*

I'm not asking to removing this (yet) ... but let's think a bit about it. Thanks!

yceruto:
This `ControllerTestCase` is just an utility class to make tests more descriptive and simple, it's more related to basic POO inheritance and DRY (no Symfony/funtional Tests related).

I think it's good to show to newcomers good practices related to how to simplify their apps too. For newcomers of "Symfony framework" it should be easy to understand I think, anyway a nice class description could clarify possible misinterpretations.

yceruto:
Maybe I could remove the response assertions and leave only the client creation methods, what do you think?

yceruto:
Last changes:
* Moved response assert to tests metods again
* Removed controller inheritance by trait usage
* Added comments

yceruto:
Rebased and resolved conflicts with https://github.com/symfony/symfony-demo/pull/502

javiereguiluz:
@yceruto thanks for all the changes and for rebasing, etc. There are a lot of things that I like in this PR, so I'm going to merge it now. However, there are some things that I don't like because it's not a practice promoted by Symfony (although your proposal works perfectly and it's technically correct). So, I'm going to create a PR soon after to revert some changes. Thanks!

PR #483, opened 2017-02-20
[WIP] make the app/config/services.yml file use the new Symfony autoregistration and autowiring features.

hhamon:
This PR will work as expected when the following PRs will be merged in the main Symfony repository:

* https://github.com/symfony/symfony/pull/21690
* https://github.com/symfony/symfony/pull/21627

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -32,6 +32,7 @@ class AppKernel extends Kernel
         if (in_array($this->getEnvironment(), ['dev', 'test'])) {
             $bundles[] = new Symfony\Bundle\DebugBundle\DebugBundle();
             $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
+            $bundles[] = new Symfony\Bundle\WebServerBundle\WebServerBundle();
             $bundles[] = new Sensio\Bundle\DistributionBundle\SensioDistributionBundle();
             $bundles[] = new Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle();
 
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -32,9 +32,12 @@ class AppKernel extends Kernel
         if (in_array($this->getEnvironment(), ['dev', 'test'])) {
             $bundles[] = new Symfony\Bundle\DebugBundle\DebugBundle();
             $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
-            $bundles[] = new Symfony\Bundle\WebServerBundle\WebServerBundle();
             $bundles[] = new Sensio\Bundle\DistributionBundle\SensioDistributionBundle();
-            $bundles[] = new Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle();
+
+            if ('dev' === $this->getEnvironment()) {
+                $bundles[] = new Symfony\Bundle\WebServerBundle\WebServerBundle();
+                $bundles[] = new Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle();
+            }
 
             if ('test' === $this->getEnvironment()) {
                 // this bundle makes it easier to work with databases in PHPUnit
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -51,18 +51,17 @@ services:
     # In this example, the classes found in the following directories will be
     # automatically registered as services:
     #
+    # * src/AppBundle/EventListener/
     # * src/AppBundle/Form/Type/
     # * src/AppBundle/Security/
     # * src/AppBundle/Twig/
     # * src/AppBundle/Utils/
     #
-    # Thus, only the classes whose corresponding filename ends with the "Subscriber.php"
-    # suffix in the src/AppBundle/EventListener/ directory will be registered as
-    # services in the service container. Other classes of this directory will be simply
-    # ignored.
+    # The "resource" attribute can accept more specific glob patterns like
+    # "{EventListener/*Subscriber.php,Form/Type/*Type.php}".
     AppBundle\:
         # Register all classes in the src/AppBundle directory as services
-        resource: '../../src/AppBundle/{EventListener/*Subscriber.php,Form/Type,Security,Twig,Utils}'
+        resource: '../../src/AppBundle/{EventListener,Form/Type,Security,Twig,Utils}'
 
     # The other section defines a rule to automatically register and autowire the
     # controller classes found in the src/AppBundle/Controller/ directory.
@@ -116,7 +115,7 @@ services:
     # Those events are defined in the tags added to the service definition.
     # See http://symfony.com/doc/current/event_dispatcher.html#creating-an-event-listener
     AppBundle\EventListener\CommentNotificationListener:
-        arguments: ['@mailer', '@router', '@translator', '%app.notifications.email_sender%']
+        $sender: '%app.notifications.email_sender%'
         # The "method" attribute of this tag is optional and defaults to "on + camelCasedEventName"
         # If the event is "comment.created" the method executed by default is "onCommentCreated()".
         tags:
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -111,12 +111,5 @@ services:
     AppBundle\EventListener\RedirectToPreferredLocaleSubscriber:
         $locales: '%app_locales%'
 
-    # Event Listeners are classes that listen to one or more specific events.
-    # Those events are defined in the tags added to the service definition.
-    # See http://symfony.com/doc/current/event_dispatcher.html#creating-an-event-listener
-    AppBundle\EventListener\CommentNotificationListener:
+    AppBundle\EventListener\CommentNotificationSubscriber:
         $sender: '%app.notifications.email_sender%'
-        # The "method" attribute of this tag is optional and defaults to "on + camelCasedEventName"
-        # If the event is "comment.created" the method executed by default is "onCommentCreated()".
-        tags:
-            - { name: kernel.event_listener, event: comment.created, method: onCommentCreated }
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -1,83 +1,123 @@
 services:
-    # First we define some basic services to make these utilities available in
-    # the entire application
-    slugger:
-        class: AppBundle\Utils\Slugger
 
-    markdown:
-        class: AppBundle\Utils\Markdown
-
-    # These are the Twig extensions that create new filters and functions for
-    # using them in the templates
-    app.twig.app_extension:
-        public:    false
-        class:     AppBundle\Twig\AppExtension
-        arguments: ['@markdown', '%app_locales%']
-        tags:
-            - { name: twig.extension }
-
-    app.twig.intl_extension:
+    # The special "_defaults" section configure the default behaviors for registering
+    # new services into the Symfony service container. This section has a local scope
+    # only, which means it only affects the service definitions being registered by
+    # this file.
+    #
+    # The "autowire" section tells Symfony to make its best to autowire new services
+    # by introspecting their constructor.
+    #
+    # The "public" section with the FALSE boolean value forces all new autowired classes
+    # to be declared as private services. Marking services private limits their scope to
+    # the Symfony service container itself only. They're not accessible from the outside
+    # world and thus let the service container optimizes itself when it's being compiled
+    # and dumped to raw PHP code.
+    _defaults:
+        autowire: true
         public: false
-        class:  Twig_Extensions_Extension_Intl
-        tags:
-            - { name: twig.extension }
 
-    # Defining a form type as a service is only required when the form type
-    # needs to use some other services, such as the entity manager.
-    # See http://symfony.com/doc/current/best_practices/forms.html
-    app.form.type.tagsinput:
-        class: AppBundle\Form\Type\TagsInputType
-        arguments: ['@doctrine.orm.entity_manager']
-        tags:
-            - { name: form.type }
+    # The special "_instanceof" section appends additional wiring configuration 
+    # to any service definitions that match the listed types.
+    #
+    # For instance, the first rule adds a special "kernel.event_subscriber" tag
+    # to any definitions declared in this file whose class is an instance of the
+    # "Symfony\Component\EventDispatcher\EventSubscriberInterface" class or interface.
+    _instanceof:
+        Symfony\Component\EventDispatcher\EventSubscriberInterface:
+            tags: ['kernel.event_subscriber']
+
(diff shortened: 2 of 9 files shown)


stof:
IMO, we should not merge it until we have a 3.3 release in Symfony, to be able to keep releasing the demo project

hhamon:
Yes that's the plan!

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -189,4 +190,13 @@ public function deleteAction(Request $request, Post $post)
 
         return $this->redirectToRoute('admin_post_index');
     }
+
+    /**
+     * Must be protected to allow Symfony DIC to create a proxy.
+     *
+     */
+    protected function getSlugger(): Slugger
```
stof: this breaks compat with PHP 5.x, while the demo still supports it for now

hhamon: True! I will update the PR.

src/AppBundle/Controller/SecurityController.php
```diff
@@ -13,6 +13,7 @@
```
javiereguiluz: These SecurityController.php changes are the perfect example of why I don't like this new way of developing in Symfony. You remove 1 easy-to-understand line (`$helper = $this->get('security.authentication_utils');`) and replace it with 6 lines of code ... including some "class guessing" to import the `AuthenticationUtils` class 😕

src/AppBundle/Controller/Admin/BlogController.php
javiereguiluz: My guess is that this feature will confuse a lot of users. You add to your class an empty getter ... which during runtime is not empty because somehow Symfony injects the needed code to return what you want to get 😕

src/AppBundle/EventListener/RedirectToPreferredLocaleSubscriber.php
```diff
@@ -23,7 +25,7 @@
  *
  * @author Oleg Voronkovich <oleg-voronkovich@yandex.ru>
  */
-class RedirectToPreferredLocaleListener
+class RedirectToPreferredLocaleSubscriber implements EventSubscriberInterface
```
javiereguiluz: Just asking: is this change needed because of Symfony ... or is it because you configured this: `resource: '../../src/AppBundle/{EventListener/*Subscriber.php,...}'`

In other words ... if the previous config was: `resource: '../../src/AppBundle/{EventListener,...}'` would this change be required?

app/config/services.yml
```diff
@@ -1,83 +1,121 @@
...
+    # To do so, the remaining unwired named arguments must be defined with their
+    # corresponding values. Here, the "$locales" named argument is configured to
+    # receive the value of the global "%app_locales%" parameter defined under the
+    # "parameters" section of the "app/config/config.yml" file.
+    #
+    # Note that the order in which the named arguments are defined below doesn't
+    # matter as they're referenced here by their real names in the PHP code.
+    # Symfony is then smart enough to make the corresponding matching when compiling,
+    # optimizing and dumping the container as a raw PHP class in the cache directory.
+    AppBundle\Twig\AppExtension:
```
javiereguiluz: This may introduce some confusion. The reason is that you already configured this service previously as part of:

```
AppBundle\:
  resource: '../../src/AppBundle/{EventListener/*Subscriber.php,Form/Type,Security,Twig,Utils}'
```

And also:

```
_instanceof:
  Twig_ExtensionInterface:
    tags: ['twig.extension']
```

And now you are partially configuring it ... or completing its configuration:

```
AppBundle\Twig\AppExtension:
  $locales: '%app_locales%'
```

So ... the `AppBundle\Twig\AppExtension` is defined in three different parts of the file 😕

app/config/services.yml
```diff
@@ -1,83 +1,121 @@
...
-    app.redirect_to_preferred_locale_listener:
-        class: AppBundle\EventListener\RedirectToPreferredLocaleListener
-        arguments: ['@router', '%app_locales%', '%locale%']
-        tags:
-            - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest }
-
-    app.comment_notification:
-        class: AppBundle\EventListener\CommentNotificationListener
+    AppBundle\EventListener\CommentNotificationListener:
         arguments: ['@mailer', '@router', '@translator', '%app.notifications.email_sender%']
```
javiereguiluz: Can't these be autowired?

src/AppBundle/Controller/SecurityController.php
hhamon: Yes this is to show that we can easily inject dependencies into controllers. In this example, the `AuthenticationUtils` service is always used.

src/AppBundle/EventListener/RedirectToPreferredLocaleSubscriber.php
hhamon: We may encourage to produce `EventSubscriber`s in favor of regular `EventListener`s. We can indeed simplify the `resource` to regiter all classes in the `EventListener/` directory. Because later we also override the listener class to complete its definition.

app/config/services.yml
hhamon: It can be autowired if we simplify the `AppBundle/` section by allowing registering all classes under the `EventListener/` directory. However we would still have to manually wire the last argument that is scalar and cannot be guessed by Symfony.

app/config/services.yml
hhamon: The first section lets Symfony register the classes from these directories as services. It'll make its best to autowire everything. But some service requires some extra manual configuration when Symfony cannot guess the remaining parts (dependency collision or scalar arguments). This is what the third section does.

The second section called `instanceof` allows to append configuration to every definitions that match the rule type. It has nothing to do with autowiring! This is probably the confusing part that must be well taught to newcomers. It's just a way to apply a group of configuration settings to each service definition matching the rule. It's a way to make configuration not repeat itself. This means you can still manually define your services and use the `instanceof` section to append the DI tags or extra method calls to your sets of configured services.

app/config/services.yml
```diff
@@ -1,83 +1,115 @@
...
+    # * src/AppBundle/Form/Type/
+    # * src/AppBundle/Security/
+    # * src/AppBundle/Twig/
+    # * src/AppBundle/Utils/
+    #
+    # The "resource" attribute can accept more specific glob patterns like
+    # "{EventListener/*Subscriber.php,Form/Type/*Type.php}".
+    AppBundle\:
+        # Register all classes in the src/AppBundle directory as services
+        resource: '../../src/AppBundle/{EventListener,Form/Type,Security,Twig,Utils}'
```
stof: most form types don't need to be registered as services at all (they need to become services only when they have dependencies).
And as this is broken currently because you define private services, you could avoid doing it for all of them

hhamon: Yes the current implementation is broken. I'm working on a PR to make form type services privatable.

src/AppBundle/Controller/SecurityController.php
```diff
@@ -23,18 +24,28 @@
  */
 class SecurityController extends Controller
 {
+    private $helper;
+
+    /**
+     * Constructor.
+     *
+     * @param AuthenticationUtils $helper
+     */
```
GuilhemN: this doc block does not bring useful informations, what about removing it?

src/AppBundle/Controller/SecurityController.php
nicolas-grekas: Deciding only on the number of line of code is dangerous. As discussed in https://github.com/symfony/symfony/issues/21623, the bigger issue with `$container->get('foo')` is that you get no guarantee whatsoever about what you'll get back - no contrat. It's working "by luck" in terms of design.

src/AppBundle/EventListener/RedirectToPreferredLocaleSubscriber.php
nicolas-grekas: Now that event subscribers can be made private, the new best practice should be to always use them in place of listeners. Thus, I think also that *not* advertising listeners in symfony-demo is a good thing - even if it works the same as before.

src/AppBundle/Controller/SecurityController.php
hhamon: Fixed thanks.

app/AppKernel.php
```diff
@@ -32,6 +32,7 @@ public function registerBundles()
         if (in_array($this->getEnvironment(), ['dev', 'test'])) {
             $bundles[] = new Symfony\Bundle\DebugBundle\DebugBundle();
             $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
+            $bundles[] = new Symfony\Bundle\WebServerBundle\WebServerBundle();
             $bundles[] = new Sensio\Bundle\DistributionBundle\SensioDistributionBundle();
             $bundles[] = new Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle();
```
yceruto: Could you add `WebServerBundle` and `SensioGeneratorBundle` to `dev` environment only?

hhamon: Done.

app/config/services.yml
Wirone: Wow, @hhamon I didn't know about `_instanceof` config section. Are there any docs about it? Could be appended to comment to let people know what's the magic behind it.

hhamon: Documentation doesn't exist yet. This PR is a PoC to experiment the upcoming new DI features of Symfony 3.3. We'll document everything very soon ;)

app/config/services.yml
```diff
@@ -1,83 +1,115 @@
...
+    # controller classes found in the src/AppBundle/Controller/ directory.
+    #
+    # By default all services are made private according to the global "_defaults"
+    # section at the top of this file.
+    #
+    # However, in Symfony, controllers must always be declared public in order to
+    # be lazy instanciated when they're really needed. This is why the inherited
+    # default "public" attribute is overriden to force the registered controller
+    # services to be marked public.
+    AppBundle\Controller\:
```
dunglas: WYDT about removing the `AppBundle` like in https://github.com/symfony/symfony-standard/pull/1034?

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -189,4 +189,12 @@ public function deleteAction(Request $request, Post $post)
 
         return $this->redirectToRoute('admin_post_index');
     }
+
+    /**
+     * Must be protected to allow Symfony DIC to create a proxy.
+     */
+    protected function getSlugger()
```
dunglas: I would not advertise this feature in the demo (it's for advanced users). Using constructor injection is OK here and works out of the box.
It's also easier to understand for newcomers.

src/AppBundle/Controller/SecurityController.php
dunglas: I also prefer the new way, it looks more like typical PHP. No framework magic.

app/config/services.yml
hhamon: Interesting discussion. I really like the idea of getting rid of the AppBundle.

src/AppBundle/Controller/Admin/BlogController.php
hhamon: Yes I have the same feeling as well.

app/config/services.yml
javiereguiluz: I'm sorry but we are not going to get rid of AppBundle in the Symfony Demo. For this kind of changes, the Symfony Demo must go behind the Symfony Framework itself. If Symfony decides to go bundle-less for Symfony 4.0 (I wish it will) then we'll update the Symfony Demo accordingly. But we can't do that before. I'm sorry.

dunglas: We can still create PR to test this setup without merging it.

app/config/services.yml
Wirone: Ah, I've searched for `_instance` here and it was only in config so I thought it is already supported elsewhere.

app/config/services.yml
```diff
@@ -1,83 +1,115 @@
...
+    # because its class doesn't live in any of the previous defined directories.
     #
-    # app.post_repository:
-    #     class:     Doctrine\ORM\EntityRepository
-    #     factory:   ['@doctrine.orm.entity_manager', getRepository]
-    #     arguments: [AppBundle\Entity\Post]
+    # Thus, its class is not namespaced and cannot be defined in a global rule.
+    # Indeed, registering new classes whose filename matches a glob pattern as
+    # services like in the two previous sections only works for namespaced classes.
+    app.twig.intl_extension:
```
weaverryan: Would we want this to also match the class name?

src/AppBundle/Controller/SecurityController.php
weaverryan: I agree with... everyone! :D 

The lines of code are important - there is a "perceived complexity" when something has extra lines of code. It's superficial, but, in a sense, people "judge a book bike its cover" by subconsciously seeing a lot of lines of code and thinking that something is complex.

On the other hand, controllers have always behaved differently than other services (where you *do* need to inject objects and use their methods directly). That's always been unfortunate, and has increased the later-parts of Symfony's learning curve (the base controller shortcuts make it easy to get started... but then you pay a bit more of a price later when you need to learn about services, DI, etc).

In this case, for me, I'm ok with it. But, it does strike me as odd that we *still* have a mixture of approaches: sometimes (e.g. `$this->render()`), we're relying on the base controller and using its shortcut methods. But in other cases, we say that you should inject the service you need? If we're moving towards controllers as services, shouldn't we *not* extend the base controller anymore? If we didn't, that would be a big problem, as it would increase the boilerplate code for many things. Unless, we added traits with getter injects (which could be used for both controllers and in services). I'm not sure what the best approach is - this just strikes me as walking half-way between 2 worlds.

weaverryan: In fact, does this even work? Since the controller is a service (and I think? that the routing automatically knows to use the service, since it exists), the container is never set, correct? So `$this->render()` is broken.

app/config/services.yml
javiereguiluz: If I do that, I get this error:

```
RuntimeException in CheckDefinitionValidityPass.php line 52:

The definition for "Twig_Extensions_Extension_Intl" has no class attribute, and
appears to reference a class or interface in the global namespace. Leaving out
the "class" attribute is only allowed for namespaced classes. Please specify the
class attribute explicitly to get rid of this error.
```

javiereguiluz:
I'm finishing this PR in #562. Hugo, I'm sorry but I couldn't reuse your original commits because there were too many conflicts (and because Symfony changed the way some of these new features work too). Thanks!

hhamon:
Hi guys! I'll update this PR this week :)

PR #485, opened 2017-02-21
[RFC] Run lint commands on TravisCI

bocharsky-bw:


composer.json
```diff
@@ -25,7 +25,7 @@
         "symfony/monolog-bundle"               : "^3.0",
         "symfony/polyfill-apcu"                : "^1.0",
         "symfony/swiftmailer-bundle"           : "^2.3",
-        "symfony/symfony"                      : "^3.2",
+        "symfony/symfony"                      : "dev-master",
```
stof: depending on dev-master directly is a total no-go.

And depending of ``^3.3@dev`` would make this PR unmergeable, as it would mean that the demo would be in a state where it cannot be released (depending on beta releases is fine, but not on the dev version)

bocharsky-bw: I agree. I think `lint:xliff` command will be available in Symfony 3.3 so let's just wait for it. Otherwise, I can remove `xliff` command to make this PR mergeable for now - we may add xliff linter later when it will be released. Wdyt?

stof: I would do the PR without xliff for now (and without changing dependencies), and add the xliff command once the demo is updated to 3.3.

Keeping the deps update in this PR will just be painful, as it will create conflicts with other PRs moving to Symfony 3.3 (we already have a PR migrating to Symfony 3.3 to use the new DI features)

bocharsky-bw: Great, I updated PR. Thanks!

javiereguiluz:
@bocharsky-bw very nice! I love these little linters (and soon we'll add "lint:xliff" too!)

PR #490, opened 2017-02-27
Fix PSR4 in composer.json, add dump()

nicolas-grekas:
Empty psr-4 prefixes can force the autoloader to scan the `src/` directory for all namespaces. The best practice is to always make the prefix explicit.

`composer update` also, doing:
```
  - Updating twig/twig (v1.31.0 => v1.32.0) Loading from cache
  - Updating sensiolabs/security-checker (v4.0.0 => v4.0.1) Loading from cache
  - Updating phpunit/php-timer (1.0.8 => 1.0.9) Loading from cache
  - Updating phpunit/php-token-stream (1.4.9 => 1.4.11) Loading from cache
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,11 +4,15 @@
     "type": "project",
     "description": "Symfony Demo Application",
     "autoload": {
-        "psr-4": { "": "src/" },
+        "psr-4": {
+            "AppBundle\\": "src/AppBundle/",
+            "CodeExplorerBundle\\": "src/CodeExplorerBundle/"
+        },
         "classmap": [ "app/AppKernel.php", "app/AppCache.php" ]
     },
     "autoload-dev": {
-        "psr-4": { "Tests\\": "tests/" }
+        "psr-4": { "Tests\\": "tests/" },
+        "files": [ "vendor/symfony/symfony/src/Symfony/Component/VarDumper/Resources/functions/dump.php" ]
     },
     "require": {
         "php"                                  : ">=5.5.9",
--- a/composer.lock
+++ b/composer.lock
@@ -1485,16 +1485,16 @@
         },
         {
             "name": "sensiolabs/security-checker",
-            "version": "v4.0.0",
+            "version": "v4.0.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/sensiolabs/security-checker.git",
-                "reference": "116027b57b568ed61b7b1c80eeb4f6ee9e8c599c"
+                "reference": "f2ce0035fc512287978510ca1740cd111d60f89f"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/sensiolabs/security-checker/zipball/116027b57b568ed61b7b1c80eeb4f6ee9e8c599c",
-                "reference": "116027b57b568ed61b7b1c80eeb4f6ee9e8c599c",
+                "url": "https://api.github.com/repos/sensiolabs/security-checker/zipball/f2ce0035fc512287978510ca1740cd111d60f89f",
+                "reference": "f2ce0035fc512287978510ca1740cd111d60f89f",
                 "shasum": ""
             },
             "require": {
@@ -1525,7 +1525,7 @@
                 }
             ],
             "description": "A security checker for your composer.lock",
-            "time": "2016-09-23T18:09:57+00:00"
+            "time": "2017-02-18T17:53:25+00:00"
         },
         {
             "name": "swiftmailer/swiftmailer",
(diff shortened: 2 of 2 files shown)


stof:
👍

composer.json
```diff
@@ -4,11 +4,15 @@
...
+        "psr-4": {
+            "AppBundle\\": "src/AppBundle/",
+            "CodeExplorerBundle\\": "src/CodeExplorerBundle/"
+        },
         "classmap": [ "app/AppKernel.php", "app/AppCache.php" ]
     },
     "autoload-dev": {
-        "psr-4": { "Tests\\": "tests/" }
+        "psr-4": { "Tests\\": "tests/" },
+        "files": [ "vendor/symfony/symfony/src/Symfony/Component/VarDumper/Resources/functions/dump.php" ]
```
theofidry: why is this needed? isn't it supposed to already be included from the VarDumper `composer.json`?

javiereguiluz:
Nice! I think this is important, so I've tweeted about it https://twitter.com/symfony_en/status/836232295329722370

composer.json
```diff
@@ -4,11 +4,15 @@
     "type": "project",
     "description": "Symfony Demo Application",
     "autoload": {
-        "psr-4": { "": "src/" },
+        "psr-4": {
```
theofidry: whilst this may boost performance, it may be harder for new users don't you think?

javiereguiluz: But only if you create a ton of bundles in your app ... which by the way is now considered a bad practice in most of the cases!

theofidry: Well to be honest I don't even see why you need any bundle here in the first place, but it's not limited to bundles. It highly depends on how you manage your `src` bundle. I would personally promote more something like having a `App` namespace for the whole `src`.

composer.json
theofidry: This is more as to ask for clarifications. I seem to be unable to use the `dump` function when the app is not bootstrapped sometimes so I took the habit to add a shortcut to make a FQCN call to avoid that problem, but this still looks weird as I would expect to be able to use `dump` at anytime.

nicolas-grekas: When symfony/symfony is required, the dump function is not autoloaded. It is only when symfony/var-dumper is.

composer.json
stof: @theofidry the symfony-demo project has a AppBundle namespace for its source, because this is what the Symfony best practices are telling (and the demo is also about showing these best practices).

The extra CodeExplorerBundle is more a project-specific vendor bundle actually (we haven't shared it as a real vendor bundle, because this feature simply does not make sense outside the demo project).

and enw users won't find it harder, as they won't need to change the autoload config of symfony-demo.

babaorum:
First of all, nice catch ! 
But after a few minutes on the subject, don't you think this is just a patch and not really correcting the real issue ? 
From my point of view we have composer on one hand which handle the autoloading. 
And we have symfony on the other hand which handle the application. 
But there is no way for symfony to tell composer where to look for our application directly. 
So before your PR, composer was kind of on is own to find our app. 
But now it's to the user to tell composer as well as symfony that we created a new bundle. 

I think the idealist way would be that composer has a way to load this from an external file, which would be generated by symfony on cache warm up for prod env. 

What do you think ? I would really appreciate your opinions on this. 

Thanks by advance. And once more, great job for digging this.

nicolas-grekas:
@babaorum my humble opinion: we can wonder how the world could be better with more things being done out of the box, but that's really hard, many tried already with some success (composer is great!), and that's always improving. If you think you have some ideas that could make all of this even better, PRs are welcomed on both Composer and Symfony. For this very PR, it's unrelated. Just my 2cts :)

juliendufresne:
The issue only exists  in the developper environment because when you push your code in production, you should use ˋcomposer install --optimize-autoloader` Or better `composer install --classmap-authoritativeˋ which will produce a classmap and is the fastest autoload.

composer.json
Pierstoval: Isn't the `dump` function loaded when `Debug::enable()` is called?

stof: @Pierstoval no, it is loaded when the DebugBundle is booted. but when running tests, it is common for it not to be booted, making it painful. This is why we added it as a dev autoloading here for convenience.

stof:
@juliendufresne ``--optimize-autoloader`` will not help for classmap misses. So while it is already better (PSR-4 fallbacks are checked after PSR-4 prefixes but before any PSR-0 config), it is still better to use a prefix.
This optimization is indeed useless when using ``--classmap-authoritative`` as composer will not do any PSR-4 lookup at runtime then.

composer.json
Pierstoval: Then symfony/symfony-standard#1025 should also be merged, shouldn't it? :)

Issue #494, opened 2017-03-04
User not implementing Serializable in symfony-demo?


ptrm04:
According to http://symfony.com/doc/current/security/entity_provider.html#what-do-the-serialize-and-unserialize-methods-do the User class should implement `Serializable` - why does the symfony-demo not implement it on the User class?

voronkovich:
@ptrm04, you're right, the `User` class should implement the `Serializable` interface. Could you please create a PR to fix this issue (just copy and paste code from the Symfony's tutorial)?

ptrm04:
@voronkovich thanks for the confirmation, will do.

Symfony Demo (and any app where you don't need to serialize anything in particular) will actually work just fine without implementing the `Serializable` interface - why is that? And since it is working, do we **_really_** need it?

ptrm04:
Quick check revealed that if you don't implement `Serializable` it will basically put everything in the session i.e. email anna_admin@... et al. can be found in the session. So would the purpose of implementing `Serializable` is to strictly decide what needs to be serialized? Otherwise everything will go in?

javiereguiluz:
This is one of the multiple things that I don't like about the Symfony Security. Why should we do this ourselves? Shouldn't Symfony by default apply some smart serialization and force you to define it yourself only in advanced/complex cases?

stof:
@javiereguiluz serializing a PHP object which is not Serializable works fine in most cases.
but there is one case where it causes issues: serializing a Doctrine Proxy which has not yet been initialized, because the internal PHP serialization logic will not trigger the proxy initialization. But when implementing ``Serializable``, the logic calls the method, which loads the proxy data.

stof:
And note that this is true any time you serialize a Doctrine entity, not only for the Security component (and it is documented in the Doctrine doc IIRC)

javiereguiluz:
Thanks for the explanation. But I wonder why should we document this, implement this, etc.  ... instead of changing this. It doesn't feel right. It's a burden that we impose to our users ... and for something a critical as security.

voronkovich:
@ptrm04, you're right. As mentioned [here](https://dzone.com/articles/how-correctly-work-php):

> However, it follows field references, so if you're referencing an ORM or framework objects from your own serialized ones, you're likely to grab the whole library with it

So, it's better in most cases to specify what to serialize programmatically.

javiereguiluz:
Closing it as fixed by #495.

PR #495, opened 2017-03-07
#494 Implements Serializable interface in the User entity

apetitpa:
As discussed in #494, implements the `Serializable` interface in the `User` entity.

### DIFF
--- a/src/AppBundle/Entity/User.php
+++ b/src/AppBundle/Entity/User.php
@@ -18,7 +18,7 @@ use Symfony\Component\Security\Core\User\UserInterface;
  * @author Ryan Weaver <weaverryan@gmail.com>
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  */
-class User implements UserInterface
+class User implements UserInterface, \Serializable
 {
     /**
      * @var int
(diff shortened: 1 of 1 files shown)


src/AppBundle/Entity/User.php
```diff
@@ -163,4 +163,28 @@ public function eraseCredentials()
...
         // $this->plainPassword = null;
     }
+
+    /** @see \Serializable::serialize() */
+    public function serialize()
+    {
+        return serialize([
+            $this->id,
+            $this->username,
+            $this->password,
```
javiereguiluz: Is adding the password here a recommended practice (from the security point of view)?

apetitpa: I don't think so indeed but I used the example from the documentation http://symfony.com/doc/current/security/entity_provider.html#create-your-user-entity

apetitpa: Quoted from http://symfony.com/doc/current/security/entity_provider.html#understanding-serialize-and-how-a-user-is-saved-in-the-session

> First, the Serializable interface and its serialize() and unserialize() methods have been added to allow the User class to be serialized to the session. This may or may not be needed depending on your setup, but it's probably a good idea. In theory, only the id needs to be serialized, because the refreshUser() method refreshes the user on each request by using the id (as explained above). This gives us a "fresh" User object.

But Symfony also uses the username, salt, and password to verify that the User has not changed between requests (it also calls your AdvancedUserInterface methods if you implement it). Failing to serialize these may cause you to be logged out on each request.

javiereguiluz: This explanation is not very good 😕   Saying *"maybe you need this ... maybe you don't"* without explaining anything else doesn't look good. And then, saying *"you can only serialize "id" ... but maybe it doesn't work because Symfony wants more"* is also confusing.

As usual, this is caused by the massive internal complexity of the security component ... so I don't know what can we do about this.

apetitpa: I have to agree with you, it's quite unclear. I don't know what to do either maybe someone could enlight us on this particular topic ?

apetitpa:
What about this PR ?

javiereguiluz:
@apetitpa I've finally merged this because it's exactly the same as the Symfony Docs, so let's apply our own best practices here. Thanks!

Issue #489, opened 2017-02-24
Add some log


maidmaid:
For example, we could log a message ``comment added`` in ``CommentNotificationListener``, to show how work monolog in Symfony. WDYT?

javiereguiluz:
Closing because the related PR was closed ... but we'll add this soon.

PR #496, opened 2017-03-08
Add a log example as suggested in #489

apetitpa:
Add an example of how to use the logger service in `CommentNotificationListener` as suggested in #489.

### DIFF
--- a/src/AppBundle/EventListener/CommentNotificationListener.php
+++ b/src/AppBundle/EventListener/CommentNotificationListener.php
@@ -98,6 +98,6 @@ class CommentNotificationListener
         // Symfony comes with an outside library - called Monolog - that allows you to create logs
         // See http://symfony.com/doc/current/logging.html#logging-a-message
         $logger = $this->get('logger');
-        $logger->info('Comment created on ' . $linkToPost);
+        $logger->info('Comment created on '.$linkToPost);
     }
 }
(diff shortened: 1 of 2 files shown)


src/AppBundle/EventListener/CommentNotificationListener.php
```diff
@@ -94,5 +94,10 @@ public function onCommentCreated(GenericEvent $event)
         // However, you can inspect the contents of those unsent emails using the debug toolbar.
         // See http://symfony.com/doc/current/email/dev_environment.html#viewing-from-the-web-debug-toolbar
         $this->mailer->send($message);
+
+        // Symfony comes with an outside library - called Monolog - that allows you to create logs
+        // See http://symfony.com/doc/current/logging.html#logging-a-message
+        $logger = $this->get('logger');
```
yceruto: I guess it doesn't work? so you need inject it on the constructor.

apetitpa: You're right, I'll make a commit to fix this. Sorry for the oversight.

javiereguiluz:
@apetitpa I appreciate your contribution ... but I wonder if we need to add this change. As you can see, we already inject a lot of services, so it looks a bit complex if we keep injecting more services.

If the purpose is to show how to log in a service, maybe we can add this to a different (and simpler) service ... and maybe we can show optional dependencies too to allow `null` as the logger.

What do others think about this? Thanks!

apetitpa:
@javiereguiluz I think you're right, I'll make some new commits to improve this PR.

javiereguiluz:
I'm closing this PR because it stalled. @apetitpa we're probably going to update this app to use Symfony Flex very soon, so I propose to wait a bit and re-add this feature once the app is updated. Thanks!

PR #497, opened 2017-03-09
Add missing french translations

apetitpa:
Add missing french translations + backport a translation from FR to EN


javiereguiluz:
@apetitpa thank you very much for completing the French translation with these missing strings!

PR #500, opened 2017-03-13
Fixes a minor php cs error that prevents builds to pass

apetitpa:
This minor php cs error was introduced by commit 774adca218e95d0a68f5ac1d41948dfd927899d2


javiereguiluz:
Thank you @apetitpa.

PR #503, opened 2017-03-14
Removed ControllerTestTrait

javiereguiluz:
In #481, @yceruto introduced some nice features. I like them all ... but one of them is not compatible with the way Symfony promotes to do things. The `ControllerTestTrait` is a PHP trait with some helper methods to reduce the boilerplate code in tests. Even if the solution is technically correct, we don't use it in the official Symfony docs. This can create confusion for new developers studying this app and reading the docs.

I propose to revert this feature in this PR ... and later we can discuss about ways to improve this without using traits. Thanks!


yceruto:
I'm fine with that 👍

javiereguiluz:
@yceruto thanks for understanding it!

PR #508, opened 2017-03-15
ApiDoc

kanghua-wangchanghao:


.gitignore
```diff
@@ -15,3 +15,4 @@
 !var/SymfonyRequirements.php
 /vendor/
 /web/bundles/
+.idea/
```
yceruto: This should be excluded in you `.git/info/exclude` or global

app/config/config.yml
```diff
@@ -89,3 +89,6 @@ swiftmailer:
     username:  "%mailer_user%"
     password:  "%mailer_password%"
     spool:     { type: memory }
+
+#开启ApiDoc配置
+nelmio_api_doc: ~
```
yceruto: Missing newline at end of file

app/config/routing.yml
```diff
@@ -25,3 +25,8 @@ homepage:
         _controller: FrameworkBundle:Template:template
         template:    default/homepage.html.twig
         _locale:     '%locale%'
+
+#配置ApiDoc路由
```
yceruto: The comments should be in English

src/AppBundle/Controller/BlogController.php
```diff
@@ -20,6 +20,7 @@
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
+use Nelmio\ApiDocBundle\Annotation\ApiDoc;//添加ApiDoc依赖
```
yceruto: Unnecessary comments

web/app.php
```diff
@@ -28,7 +28,7 @@
 // $loader->unregister();
 // $apcLoader->register(true);
 
-$kernel = new AppKernel('prod', false);
+$kernel = new AppKernel('prod', true);//因为缓存无法启动ApiDoc
```
yceruto: Should be reverted (no make sense), in prod env `debug` should be `false`.

javiereguiluz:
@kanghua-wangchanghao we appreciate your contribution, but sadly we can't accept this contribution. The main reason is that this bundle is used to generate the documentation of the API ... but we don't have an API in this application. Besides, for this kind of important changes, we prefer to create an Issue first so we can discuss about it and once approved, then submit the pull request.

So, I must close this pull request, but I hope you understand it. Thanks!

Issue #504, opened 2017-03-14
Deploy to heroku fails


fiunchinho:
Yesterday, this was working fine, but today I get the following error when trying to deploy to Heroku

```
ATTENTION: This operation should not be executed in a production environment.



Creating database schema...

Database schema created successfully!

PHP Fatal error:  Trait 'Tests\FixturesTrait' not found in /app/src/AppBundle/DataFixtures/ORM/PostFixtures.php on line 38



[37;41m                                                                   [39;49m

[37;41m  [Symfony\Component\Debug\Exception\ClassNotFoundException]       [39;49m

[37;41m  Attempted to load trait "FixturesTrait" from namespace "Tests".  [39;49m

[37;41m  Did you forget a "use" statement for another namespace?          [39;49m

[37;41m                                                                   [39;49m



[32mdoctrine:fixtures:load [--fixtures [FIXTURES]] [--append] [--em EM] [--shard SHARD] [--purge-with-truncate] [--multiple-transactions] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] [39m[32m<command>[39m
```

apetitpa:
I think this issue is related to the PR #490 and the fact that the namespace where FixturesTrait lives (Tests) is autoloaded in dev envrionment only. A fix might be to move this class from the tests folder to another folder in the AppBundle namespace maybe the dataFixtures folder ?

javiereguiluz:
@fiunchinho thanks for reporting this error. It was fixed in #509.

PR #509, opened 2017-03-15
#504 Move FixturesTrait class from Tests to AppBundle namespace

apetitpa:
Moved the FixturesTrait class to the AppBundle namespace in attempt to fix the heroku deployment failure. This class is needed not only in dev or test environments but also in production because the database is loaded using the doctrine fixtures bundle. Related to #504


javiereguiluz:
Thanks for fixing this bug @apetitpa.

PR #511, opened 2017-03-15
Enabled the PHP CS rule to fix the file header comment

javiereguiluz:
To avoid issues like #510.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,6 +1,17 @@
 #!/usr/bin/env php
 <?php
 
+$fileHeaderComment = <<<COMMENT
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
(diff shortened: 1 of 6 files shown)


javiereguiluz:
Now the tests fail because the PHP files in `app/` and `web/` don't have this header. We won't add it to those files, so how can we solve this?

apetitpa:
Add `->notPath('web/app_dev.php')` etc to the $finder in .php_cs file ?

javiereguiluz:
Yes ... but we want the rest of the rules applied to app/ and web/ files :(

apetitpa:
I don't think there is a way to excluse some checks for a particular path. Maybe this could be a new feature.

yceruto:
@javiereguiluz if we let the license be included in these files?

javiereguiluz:
@yceruto let's do that because I honestly see no other solution.

PR #513, opened 2017-03-17
Remove part related to Assetic which is no longer used

apetitpa:
Since #387 Assetic is no longer used to manage assets.


javiereguiluz:
Good catch, thanks @apetitpa.

PR #516, opened 2017-03-21
Add "show" button on edit post

yceruto:
I propose to add a new button "show post" to review quickly the edited post, it avoid come back to post list for perform this common action.

I think it's more an ergonomic issue rather than a new feature.

**Preview:**

show_btn


javiereguiluz:
It makes sense! Thanks for this contribution.

PR #519, opened 2017-03-29
Conform to Twig coding standards

michaelperrin:
See http://twig.sensiolabs.org/doc/2.x/coding_standards.html:

> Do not put any spaces before and after the opening and the closing of arrays and hashes.

> Put one (and only one) space after the start of a delimiter ({{, {%, and {#) and before the end of a delimiter (}}, %}, and #})

I changed a `{%else%}` to `{% else %}` on this point.


bocharsky-bw:
👍

Something that our PHP CS Fixer doesn't cover

yceruto:
:+1:

javiereguiluz:
I hate this and it looks ugly ... but it's the standard, so let's merge this! Thanks @michaelperrin!

michaelperrin:
Thanks @javiereguiluz! I agree with you, that doesn't really look better, but it's the standard!

PR #525, opened 2017-04-03
Prevent affecting some styles on the right sidebar

bocharsky-bw:
Before, some styles like margin also were applied to the right sidebar elements like `p`, `h3` tags.


javiereguiluz:
Thanks Victor.

PR #527, opened 2017-04-08
Updated the DocBlocks of the Entities

cafferata:
Updated the DocBlocks of the Entities from consistency point of view.

### DIFF
--- a/src/AppBundle/Entity/Comment.php
+++ b/src/AppBundle/Entity/Comment.php
@@ -76,6 +76,9 @@ class Comment
      */
     private $author;
 
+    /**
+     *
+     */
     public function __construct()
     {
         $this->publishedAt = new \DateTime();
@@ -91,11 +94,17 @@ class Comment
         return !$containsInvalidCharacters;
     }
 
+    /**
+     * @return int
+     */
     public function getId()
     {
         return $this->id;
     }
 
+    /**
+     * @return string
+     */
     public function getContent()
     {
         return $this->content;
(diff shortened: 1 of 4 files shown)


javiereguiluz:
@cafferata thanks for this improvement. In 49d6dedbfb27ff49d96614525b14c38c2aad1494 I've removed the empty PHPdoc comments for `__construct()` methods because it's not a Symfony practice to do that. Thanks!

src/AppBundle/Entity/Comment.php
```diff
@@ -91,11 +94,17 @@ public function isLegitComment()
         return !$containsInvalidCharacters;
     }
 
+    /**
+     * @return int
```
bocharsky-bw: @javiereguiluz Sorry, it's probably too late, but we don't need these `@return` statements for getters since we already have it on properties and PhpStorm could resolve them by itself: https://github.com/cafferata/symfony-demo/blob/e039e604a31754e371811b5dba2b0eb738af07fc/src/AppBundle/Entity/Comment.php#L33

But we need them for setters though.

Some discussion about it here: https://github.com/symfony/symfony-demo/pull/394

cafferata:
@bocharsky-bw Thanks for pointing this out! Your feedback is appreciated.
@javiereguiluz What to do with this feedback?

javiereguiluz:
@cafferata yes, we should remove those statements as suggested by @bocharsky-bw.

cafferata:
@javiereguiluz I removed the `@return` statements based on @bocharsky-bw feedback, and push the code to the same branch after an upstream merge. Can you merge this branch again or do I need to do something else?

javiereguiluz:
@cafferata we use a custom tool for merging Symfony PRs ... and I'm afraid I cannot merge the PR again. Maybe you can create a new PR and "cherrypick" your new commit? Thanks!

bocharsky-bw:
I think we can just revert this merged PR, since annotations like `@param \DateTime $publishedAt` we don't need to - we have type hints for objects which make this annotation pointless either

PR #531, opened 2017-04-11
Add a command that checks for missing translations in foreign language files

elkuku:
This will add a command that checks foreign language translation files for missing strings.

Executing the command without arguments checks all language files:

2017-04-11-133404_1366x768_scrot

Providing a locale parameter only files for this locale will be checked and untranslated keys are displayed along with their fallback message.

2017-04-11-133508_1366x768_scrot

If everything is translated a `<success>` message will be displayed.

2017-04-11-134342_1366x768_scrot

This command is using the [`SymfonyStyle`](http://symfony.com/doc/current/console/style.html) class to style the console output using the "Symfony style guide". I believe that also the other console commands should use this class to ensure a uniform display.

### DIFF
--- a/src/AppBundle/Command/CheckTranslationsCommand.php
+++ b/src/AppBundle/Command/CheckTranslationsCommand.php
@@ -1,0 +1,149 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace AppBundle\Command;
+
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@elkuku thanks for this contribution. However, I must close it as "won't merge". Why? Because Symfony already provides a command for this: `debug:translation`.

It's true that your command is more specialized and quickly provides a summary for all locales, but adding it could confuse Symfony developers into thinking that they must develop this command themselves in their projects. That's why I'm closing it. I'm sorry but I hope you understand it. Thanks!

jkufner:
@elkuku I think this pull request should be posted to Symfony itself, rather than this demo app.

PR #533, opened 2017-04-14
Fix typo

Mithrandir0407:


voronkovich:
@Mithrandir0407, I think it's not a typo. See http://grammarist.com/usage/dealt-dealed/

bocharsky-bw:
Anyway, `dealt` is more common form 👍

javiereguiluz:
Thanks for this fix. As @voronkovich said is not a bug strictly speaking ... but your proposal looks much better. Thanks!

PR #534, opened 2017-04-17
Use ::class constants to set entities repostitories

voronkovich:


bocharsky-bw:
Hm, I even didn't know it works. I wonder is it a good practice? Do we have any mention of this feature in Symfony docs? Does Symfony plugin for PhpStorm autocomplete it well?

javiereguiluz:
I'm 👎  because this will confuse some people into thinking that `PostRepository::class` is all they need (without the corresponding `use` import). And for experienced users, this requires more typing, so I don't know if it's worth it.

jkufner:
`::class` is much better than string (because PHP checks whether the class exists). If you don't want `use`, then use full class name with namespace.

stof:
@jkufner PHP does not check for class existence when using `::class`

javiereguiluz:
@jkufner I didn't know what @stof said, but it's true:

```php
echo ThisDoesNotExist::class;
// result = 'ThisDoesNotExist'
```

In any case, Doctrine annotations would show a warning message if you use a string and it corresponds to a class that doesn't exist.

jkufner:
@stof Good to know, Thank you.

> In any case, Doctrine annotations would show a warning message if you use a string and it corresponds to a class that doesn't exist.

@javiereguiluz Why use of `::class` is different? This looks like a bug in Doctrine.

javiereguiluz:
@jkufner if I change this:

```php
    /**
     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\ThisDoesNotExist")
     * @ORM\JoinColumn(nullable=false)
     */
    private $author;
```

I see this error:

doctrine-entity-error

I'd ask the opposite: if the related class doesn't exist, why shouldn't it display an error message?

jkufner:
@javiereguiluz The error message is correct behavior. I understood the comments above that `@ORM\ManyToOne(targetEntity=\AppBundle\Entity\ThisDoesNotExist::class)` does not show the desired error.

voronkovich:
@javiereguiluz, I tend to agree with you! Thanks!

PR #535, opened 2017-04-20
Revert "Updated the DocBlocks of the Entities from consistency pov"

bocharsky-bw:
This reverts #527

The reasons were described earlier in #394


javiereguiluz:
Victor, thanks for taking care of this. One day we'll upgrade to PHP 7.1 and we'll use "string" and "int" as the return type hint; can you imagine that!?

bocharsky-bw:
Oh man, I'm really waiting for it - it's my favorite PHP7 feature :) And then we won't need the remaining annotation for getters/setters at all!

PR #536, opened 2017-04-20
Update security-checker and uncomment this check on Travis CI

bocharsky-bw:
Looks security checker should work now


bocharsky-bw:
It's a temporary workaround as I understand. Somehow it doesn't work on Travis without specifying the endpoint and throw the next error:
> An error occurred: gnutls_handshake() failed: A TLS fatal alert has been received..

You can see it in build logs of my first comment in this PR:
https://travis-ci.org/symfony/symfony-demo/jobs/224110287

javiereguiluz:
I like this as a temporary solution ... but what @stof said is important, so I've created #543 to not forget about this. Thanks!

bocharsky-bw:
Yes, I agree. Thanks!

PR #542, opened 2017-04-21
[Test] add unit test for CommentNotificationListener

20uf:
Hello,

The listeners are not always easy to test.
I propose this version and I am at your disposal to do better if neccessaire.

Regards

### DIFF
--- a/tests/AppBundle/EventListener/CommentNotificationListenerTest.php
+++ b/tests/AppBundle/EventListener/CommentNotificationListenerTest.php
@@ -29,37 +29,40 @@ use Symfony\Component\Translation\TranslatorInterface;
  *
  *     $ cd your-symfony-project/
  *     $ ./vendor/bin/phpunit
+ *
+ * @author Michael COULLERET <michael.coulleret@gmail.com>
  */
 class CommentNotificationListenerTest extends \PHPUnit_Framework_TestCase
 {
-    public function testOnCommentCreated()
+    public function testListenerOnCommentCreated()
     {
-        $mailerProphecy = $this->prophesize(\Swift_Mailer::class);
-        $urlGeneratorProphecy = $this->prophesize(UrlGeneratorInterface::class);
-        $translatorProphecy = $this->prophesize(TranslatorInterface::class);
+        $mailer = $this->prophesize(\Swift_Mailer::class);
+        $urlGenerator = $this->prophesize(UrlGeneratorInterface::class);
+        $translator = $this->prophesize(TranslatorInterface::class);
 
-        $userProphecy = $this->prophesize(User::class);
-        $userProphecy->getEmail('fabien@symfony.com');
+        $user = $this->prophesize(User::class);
+        $user->getEmail('fabien@symfony.com');
 
-        $postProphecy = $this->prophesize(Post::class);
-        $postProphecy->getTitle()->willReturn('bar foo');
-        $postProphecy->getAuthor()->willReturn('fabien');
-        $postProphecy->getSlug()->willReturn('bar-foo');
-        $postProphecy->getAuthor()->willReturn($userProphecy->reveal());
+        $post = $this->prophesize(Post::class);
+        $post->getTitle()->willReturn('bar foo');
+        $post->getAuthor()->willReturn('fabien');
+        $post->getSlug()->willReturn('bar-foo');
+        $post->getAuthor()->willReturn($user->reveal());
 
-        $commentProphecy = $this->prophesize(Comment::class);
-        $commentProphecy->getId()->willReturn(241);
-        $commentProphecy->getPost()->willReturn($postProphecy->reveal());
+        $comment = $this->prophesize(Comment::class);
+        $comment->getId()->willReturn(241);
+        $comment->getPost()->willReturn($post->reveal());
 
-        $event = new GenericEvent($commentProphecy->reveal());
+        $event = new GenericEvent($comment->reveal());
 
-        $urlGeneratorProphecy->generate('blog_post', ['slug' => 'bar-foo', '_fragment' => 'comment_241'], 0)->willReturn('http://foo.sf')->shouldBeCalled();
-        $translatorProphecy->trans('notification.comment_created')->shouldBeCalled();
-        $translatorProphecy->trans('notification.comment_created.description', ['%title%' => 'bar foo', '%link%' => 'http://foo.sf'])->shouldBeCalled();
+        $urlGenerator->generate('blog_post', ['slug' => 'bar-foo', '_fragment' => 'comment_241'], 0)->willReturn('https://foo.sf')->shouldBeCalled();
 
-        $listener = new CommentNotificationListener($mailerProphecy->reveal(), $urlGeneratorProphecy->reveal(), $translatorProphecy->reveal(), 'foo@bar.sf');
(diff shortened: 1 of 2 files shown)


tests/AppBundle/EventListener/CommentNotificationListenerTest.php
```diff
@@ -0,0 +1,60 @@
...
+        $user = new User();
+        $user->setEmail('bar@foo.fr');
+
+        $post = new Post();
+        $post->setTitle('bar foo');
+        $post->setAuthor($user);
+        $post->setSlug('bar-foo');
+
+        $comment = new Comment();
+        $comment->setPost($post);
```
voronkovich: @20uf, you should set a comment id which is used to genarate an url

20uf:
@voronkovich Yes but there is no setter for id. I do not think this is a good way to add it.

tests/AppBundle/EventListener/CommentNotificationListenerTest.php
20uf: @voronkovich Yes but there is no setter for id. I do not think this is a good way to add it.

voronkovich: @20uf, you could create a mock for the comment instance.

20uf: Ok 👍

20uf: Done.

20uf:
Prophecy want the same returns (with a double quote), but phpcs growl because he wants simple quotes.
See https://travis-ci.org/symfony/symfony-demo/builds/224473508

```
Prophecy\Exception\Call\UnexpectedCallException: Method call:
  - trans("notification.comment_created.description", ["%title%" => "bar foo", "%link%" => "http://foo.sf"])
on Double\TranslatorInterface\P5 was not expected, expected calls were:
  - trans(exact("notification.comment_created"))
```

tests/AppBundle/EventListener/CommentNotificationListenerTest.php
```diff
@@ -0,0 +1,62 @@
...
+ * Execute the application tests using this command (requires PHPUnit to be installed):
+ *
+ *     $ cd your-symfony-project/
+ *     $ ./vendor/bin/phpunit
+ */
+class CommentNotificationListenerTest extends \PHPUnit_Framework_TestCase
+{
+    public function testOnCommentCreated()
+    {
+        $mailerProphecy = $this->prophesize(\Swift_Mailer::class);
```
voronkovich: @20uf, you forgot to check that an email was sent.

20uf: It is not possible to have the same Swift_Message instance, the test cannot pass.
Swift_Message should be injected.

```
$message = \Swift_Message::newInstance()
    ...
 ;

 $mailerProphecy->send($message)->shouldBeCalled();
```

voronkovich: @20uf, you could just ensure that the `send` method was called:
```php
// I've never used Prophency, so I don't know whether this code works, but I hope it does :)
$mailerProphecy->send()->shouldBeCalled();
```
See this example: https://github.com/phpspec/prophecy/issues/320

voronkovich:
@javiereguiluz, can we use the Prophency in the tests or we should use the built-in phpunit's mocking framework?

20uf:
Prophecy is supported in PHPUnit directly as of PHPUnit 4.5.
Maybe we can offer other ways to use mocking on other tests.

tests/AppBundle/EventListener/CommentNotificationListenerTest.php
20uf: Nice feedback. The solution is ultimately simple

20uf:
@javiereguiluz A test that interacts with a database is not a unit test, it is an integration test. 

A unit test must be executed as quickly as possible in order to have an almost immediate return. Therefore, access to files, databases or external services in a unit test must be avoided.

javiereguiluz:
@20uf yes ... but a test must ... *test* your code ... here we are "faking" (prophesizing) everything. Maybe we can change this test to be an integration test then?

20uf:
@javiereguiluz Yes the dependencies are mock (falsified), but the objective is not to test what happens behind (on the dependencies) but only to test the behavior of the class itself.

The current tests ensure coherent behavior.

20uf:
I propose instead to add other functional tests, which are also delivered to integrations tests.
We should have its two tests (unitary and functional) on each application.

tests/AppBundle/EventListener/CommentNotificationListenerTest.php
```diff
@@ -0,0 +1,68 @@
...
+ * Execute the application tests using this command (requires PHPUnit to be installed):
+ *
+ *     $ cd your-symfony-project/
+ *     $ ./vendor/bin/phpunit
+ *
+ * @author Michael COULLERET <michael.coulleret@gmail.com>
+ */
+class CommentNotificationListenerTest extends \PHPUnit_Framework_TestCase
+{
+    public function testListenerOnCommentCreated()
```
voronkovich: I'd split this one big test into 4 ones:

1. It should be subscribed to the comment creation event;
2. It should generate a link to created comment;
3. It should send a email to admin;
4. Is should translate a sending message.

20uf: - The registration is done outside the class, it is not possible to test the dispatcher.
- The generator URL is mocked, it is checked that the generation of the link to a comment is executed (here there is no interest to test the dependency, another test already does)
- We test also that the mail is sent, nothing determines that it is an administrator, besides it is precisely: the utility to make functional tests, because here the class it knows only the notion of user (no matter who hides behind)
- Same principle for translation, one already tests its execution

It would also be easy to cut the tests if conditions existed, which is not the case here, the listener is very simple.

voronkovich: @20uf, I mean something like this:
```php
public function testItShouldSubscribeToEvent()
{
    $this->assertEqual([ Events::COMMENT_CREATED ], CommentNotificationListener::getSubscribedEvents());
}
```
IMHO, It's not a good idea to test 3 different behaviours (features) of listener in one test.

20uf: I will add the test on the subscribed.

I do not think it is very relevant to break into 3 pieces because its role and corresponds to a single feature: send an email with a link and a translation.

I think what you're asking will complicate for too much. [Let's be KISS.](https://en.wikipedia.org/wiki/KISS_principle)

20uf:
Hello @stof , Is it possible to have your opinion on the subject? Ty.

javiereguiluz:
@20uf I'm merging or closing all the pending PRs because I want to star clean when upgrading this project to Symfony Flex. I've decided to close this PR for two reasons:

* The underlying code has changed a lot; we don't even have event listeners any more, only event subscribers.
* I still think that we're "mocking too much". Once the app is updated to Symfony Flex, please re-add these tests but without any mock, so tests are as real as possible. Thanks!

PR #546, opened 2017-04-22
[Test] add unit test for RedirectToPreferredLocaleListener

20uf:
Hello,

Here is the addition of a new cover test with Prophecy.

Regards

### DIFF
--- a/src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
+++ b/src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
@@ -52,8 +52,8 @@ class RedirectToPreferredLocaleListener
     public function __construct(UrlGeneratorInterface $urlGenerator, $locales, $defaultLocale = null)
     {
         $this->urlGenerator = $urlGenerator;
+        $this->locales = array_filter(explode('|', trim($locales)));
 
-        $this->locales = explode('|', trim($locales));
         if (empty($this->locales)) {
             throw new \UnexpectedValueException('The list of supported locales must not be empty.');
         }
@@ -82,6 +82,7 @@ class RedirectToPreferredLocaleListener
         if (!$event->isMasterRequest() || '/' !== $request->getPathInfo()) {
             return;
         }
+
         // Ignore requests from referrers with the same HTTP host in order to prevent
         // changing language for users who possibly already selected it for this application.
         if (0 === stripos($request->headers->get('referer'), $request->getSchemeAndHttpHost())) {
--- a/tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
+++ b/tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
@@ -45,7 +45,14 @@ class RedirectToPreferredLocaleListenerTest extends \PHPUnit_Framework_TestCase
         $this->listener = new RedirectToPreferredLocaleListener($this->urlGenerator->reveal(), 'en|fr|es');
     }
 
-    public function testListenerOnKernelRequestUnexpectedValueException()
+    public function testListenerOnKernelRequestUnexpectedValueExceptionWhenLocaleIsEmpty()
+    {
+        $this->setExpectedException(\UnexpectedValueException::class);
+
+        new RedirectToPreferredLocaleListener($this->urlGenerator->reveal(), '', 'fr');
+    }
+
+    public function testListenerOnKernelRequestUnexpectedValueExceptionWhenLocaleIsMissing()
     {
         $this->setExpectedException(\UnexpectedValueException::class);
 
--- a/tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
+++ b/tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
@@ -55,9 +55,7 @@ class RedirectToPreferredLocaleListenerTest extends \PHPUnit_Framework_TestCase
         $this->eventProphecy->getRequest()->willReturn($this->requestProphecy->reveal())->shouldBeCalled();
         $this->eventProphecy->isMasterRequest()->willReturn(false)->shouldBeCalled();
 
-        $return = $this->listener->onKernelRequest($this->eventProphecy->reveal());
-
-        $this->assertNull($return);
+        $this->listener->onKernelRequest($this->eventProphecy->reveal());
     }
 
     public function testOnKernelRequestWhenGetPathInfoIsNotDefault()
@@ -67,9 +65,7 @@ class RedirectToPreferredLocaleListenerTest extends \PHPUnit_Framework_TestCase
         $this->eventProphecy->getRequest()->willReturn($this->requestProphecy->reveal())->shouldBeCalled();
         $this->eventProphecy->isMasterRequest()->willReturn(false)->shouldBeCalled();
 
-        $return = $this->listener->onKernelRequest($this->eventProphecy->reveal());
-
-        $this->assertNull($return);
+        $this->listener->onKernelRequest($this->eventProphecy->reveal());
     }
 
     public function testOnKernelRequestWhenSameHttpHost()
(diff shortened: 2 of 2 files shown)


tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
```diff
@@ -0,0 +1,105 @@
...
+    public function testOnKernelRequestWhenGetPathInfoIsNotDefault()
+    {
+        $this->requestProphecy->getPathInfo()->willReturn('/foo');
+
+        $this->eventProphecy->getRequest()->willReturn($this->requestProphecy->reveal())->shouldBeCalled();
+        $this->eventProphecy->isMasterRequest()->willReturn(false)->shouldBeCalled();
+
+        $return = $this->listener->onKernelRequest($this->eventProphecy->reveal());
+
+        $this->assertNull($return);
```
voronkovich: @20uf, this assertion is useless. Any php function returns `null` when returning value is not specified.
```php
echo (function() { return null; })() === (function() {})(); // True
```

20uf: Fixed

tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
```diff
@@ -0,0 +1,99 @@
...
+class RedirectToPreferredLocaleListenerTest extends \PHPUnit_Framework_TestCase
+{
+    private $requestProphecy;
+    private $urlGeneratorProphecy;
+    private $eventProphecy;
+    private $listener;
+
+    public function setUp()
+    {
+        $this->requestProphecy = $this->prophesize(Request::class);
```
javiereguiluz: I don't use Prophecy so I'm asking: is it common to add the `Prophecy` suffix to the * prophesized* variables?

tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
```diff
@@ -0,0 +1,99 @@
...
+        $this->urlGeneratorProphecy = $this->prophesize(UrlGeneratorInterface::class);
+        $this->eventProphecy = $this->prophesize(GetResponseEvent::class);
+        $this->listener = new RedirectToPreferredLocaleListener($this->urlGeneratorProphecy->reveal(), 'en|fr|es');
+    }
+
+    public function testOnKernelRequestUnexpectedValueException()
```
javiereguiluz: Would it make the test method names more readable if we replace `testOnKernelRequest + ...` by `testListener + ...` ?

tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        $request = new Request();
+        $request->headers->set('referer', 'http://foo.bar');
+        $request->headers->set('HOST', 'bar.fo');
+        $request->headers->set('Accept-Language', 'fr');
+
+        $httpKernelProphecy = $this->prophesize(HttpKernelInterface::class);
+        $event = new GetResponseEvent($httpKernelProphecy->reveal(), $request, HttpKernelInterface::MASTER_REQUEST);
+
+        $this->urlGeneratorProphecy->generate('homepage', ['_locale' => 'fr'])->willReturn('http://foo.sf');
```
javiereguiluz: Shouldn't we test somewhere that the redirected URL is this `http://foo.sf` value?

tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
20uf: Yes it is true that this is not common.

tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
20uf: Done :+1:

tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
20uf: Done :+1:

tests/AppBundle/EventListener/RedirectToPreferredLocaleListenerTest.php
```diff
@@ -0,0 +1,101 @@
...
+    {
+        $this->request = $this->prophesize(Request::class);
+        $this->urlGenerator = $this->prophesize(UrlGeneratorInterface::class);
+        $this->event = $this->prophesize(GetResponseEvent::class);
+        $this->listener = new RedirectToPreferredLocaleListener($this->urlGenerator->reveal(), 'en|fr|es');
+    }
+
+    public function testListenerOnKernelRequestUnexpectedValueException()
+    {
+        $this->setExpectedException(\UnexpectedValueException::class);
```
voronkovich: @20uf, You should check the exception's message, because listener can throw `UnexpectedValueException` in 2 cases:
https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php#L58
https://github.com/symfony/symfony-demo/blob/master/src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php#L64

20uf: Case 1 is not correct because an "explose" always returns an array.
http://php.net/manual/en/function.explode.php
In my opinion it is a logical problem in listening class. Cc @javiereguiluz

voronkovich: @20uf, the `explode` can return an **empty** array.

20uf: See http://sandbox.onlinephpfunctions.com/code/07516ee9e288484bfc8a64f284ac5c489ef0bdeb

20uf: Just case 1 is totally useless.

voronkovich: @20uf, you could fix it by passing -1 as a third argument:
```php
$this->locales = explode('|', trim($locales), -1);
```
See http://sandbox.onlinephpfunctions.com/code/f35071a578b551a0d9cc98c7154b4fe61f59cd41

20uf: I'm not agree, this alter the value in the ```$this->locales``` variable.
If I apply a correction I prefer to use ```array_filter```
- with empty locales : http://sandbox.onlinephpfunctions.com/code/00fd67fc9f3579190315a0e2d95f5a10e2d0f159
- with locales : http://sandbox.onlinephpfunctions.com/code/7ec7a8475191ce656e2b8d1b658a6b78960827a8

voronkovich: @20uf, you're right, using `array_filter` is more flexible in this case.

20uf:
Are we ok on returns?

src/AppBundle/EventListener/RedirectToPreferredLocaleListener.php
```diff
@@ -52,8 +52,8 @@ class RedirectToPreferredLocaleListener
     public function __construct(UrlGeneratorInterface $urlGenerator, $locales, $defaultLocale = null)
     {
         $this->urlGenerator = $urlGenerator;
+        $this->locales = array_filter(explode('|', trim($locales)));
```
dmaicher: why exactly the `array_filter`? Maybe just `$this->locales = explode('|', trim($locales, ' |'));`? Seems unnecessary to iterate the full array.

20uf: What do you suggest?

dmaicher: Well I guess the idea was if someone puts the separator `|` in the beginning or end of the locale list to filter empty strings out?

Then this seems more logical to me:

```php
$this->locales = explode('|', trim($locales, ' |'));

20uf: I think you should consult my exchanges with @voronkovich , which explains why a filter has been added.

dmaicher: Ok I see :wink: Sorry did not check all review comments before.

20uf: No problem ✌️  , you can approve this PR if it suits you.

javiereguiluz:
Closing this for the same reasons explained in https://github.com/symfony/symfony-demo/pull/542#issuecomment-318288240  Let's wait a bit until we re-add these tests to the updated app. Thanks!

PR #550, opened 2017-04-27
Always use https:// when linking to symfony.com

javiereguiluz:


bocharsky-bw:
Wow, a lot of links 👍

PR #555, opened 2017-05-02
Simplify comment creating functional test

voronkovich:
Current test is not truly "functional" because it knows too much about internal structure of an application.

@dmaicher, this test pass only when I disable the `PHPUnitStaticDbConnectionListener`. Could you please look at the code?

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2414,16 +2414,16 @@
     "packages-dev": [
         {
             "name": "dama/doctrine-test-bundle",
-            "version": "v1.0.10",
+            "version": "v1.0.12",
             "source": {
                 "type": "git",
                 "url": "https://github.com/dmaicher/doctrine-test-bundle.git",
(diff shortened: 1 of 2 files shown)


dmaicher:
Thanks for the report :wink: I actually think I found the problem which lies inside `dama/doctrine-test-bundle`.

I will look into this tomorrow and will let you know asap.

dmaicher:
@voronkovich there is a new release that fixes the problem :wink: 

https://github.com/dmaicher/doctrine-test-bundle/releases/tag/v1.0.12

So if you run `composer update dama/doctrine-test-bundle` and commit the changes the build should pass.

voronkovich:
@dmaicher, test passes! Thank you very much!

javiereguiluz:
@voronkovich thanks for simplifying this test. I like it!

PR #557, opened 2017-05-03
Use proper database in a test environment

voronkovich:
See https://github.com/symfony/symfony-demo/issues/556

### DIFF
--- a/app/config/config_test.yml
+++ b/app/config/config_test.yml
@@ -21,7 +21,7 @@ web_profiler:
 # In this case we just need to define a different path for the application database.
 doctrine:
     dbal:
-        path: "%kernel.root_dir%/../var/data/blog_test.sqlite"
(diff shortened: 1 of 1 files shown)


voronkovich:
@stof, I replaced a test database `path` by `url`, but now it throws an exception: `Incompatible use of dynamic environment variables "DATABASE_URL" found in parameters.` in test environment.

dmaicher:
We could just provide the environment variable in the phpunit config:

```xml
<php>
    ...
    <env name="DATABASE_URL" value="sqlite:///%kernel.root_dir%/../var/data/blog_test.sqlite" />
</php>
```
This works.

dmaicher:
@stof @javiereguiluz what do you think? Specifying the environment variable like that seems the easiest to me.

voronkovich:
Closed in favor of #572

PR #561, opened 2017-05-09
Add more PHP CS Fixer rules

bocharsky-bw:


javiereguiluz:
Thank you @bocharsky-bw.

PR #562, opened 2017-05-22
Updated the application to Symfony 3.3.0

javiereguiluz:
This finishes #483 and uses the all the new 3.3 features for DI services.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,5 +36,7 @@ script:
   - ./bin/console lint:yaml @CodeExplorerBundle
   # this checks that the Twig template files contain no syntax errors
   - ./bin/console lint:twig app/Resources @CodeExplorerBundle
+  # this checks that the XLIFF translations contain no syntax errors
+  - ./bin/console lint:xliff app/Resources
   # this checks that the application doesn't use dependencies with known security vulnerabilities
   - ./bin/console security:check --end-point=http://security.sensiolabs.org/check_lock
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -30,50 +30,50 @@ framework:
     # and with different cache configurations for each fragment
     # https://symfony.com/doc/current/book/http_cache.html#edge-side-includes
     esi:             { enabled: true }
-    translator:      { fallback: "%locale%" }
-    secret:          "%env(SYMFONY_SECRET)%"
+    translator:      { fallback: '%locale%' }
+    secret:          '%env(SYMFONY_SECRET)%'
     router:
-        resource: "%kernel.root_dir%/config/routing.yml"
+        resource: '%kernel.root_dir%/config/routing.yml'
         strict_requirements: ~
     form:            ~
     csrf_protection: ~
     validation:      { enable_annotations: true }
     templating:
         engines: ['twig']
-    default_locale:  "%locale%"
+    default_locale:  '%locale%'
     trusted_hosts:   ~
     session:
         handler_id:  session.handler.native_file
-        save_path:   "%kernel.project_dir%/var/sessions/%kernel.environment%"
+        save_path:   '%kernel.project_dir%/var/sessions/%kernel.environment%'
     fragments:       ~
     http_method_override: true
     assets: ~
 
 # Twig Configuration (used for rendering application templates)
 twig:
-    debug:            "%kernel.debug%"
-    strict_variables: "%kernel.debug%"
+    debug:            '%kernel.debug%'
+    strict_variables: '%kernel.debug%'
     form_themes:
-        - "form/layout.html.twig"
-        - "form/fields.html.twig"
+        - 'form/layout.html.twig'
+        - 'form/fields.html.twig'
 
 # Doctrine Configuration (used to access databases and manipulate their information)
 doctrine:
     dbal:
         # if you don't want to use SQLite, change the URL in parameters.yml or set the DATABASE_URL environment variable
-        url: "%env(DATABASE_URL)%"
+        url: '%env(DATABASE_URL)%'
 
         # instead of using a URL, you may also uncomment the following lines to configure your database
         # driver:   pdo_mysql
-        # host:     "%database_host%"
-        # port:     "%database_port%"
-        # dbname:   "%database_name%"
-        # user:     "%database_user%"
-        # password: "%database_password%"
+        # host:     '%database_host%'
+        # port:     '%database_port%'
+        # dbname:   '%database_name%'
+        # user:     '%database_user%'
+        # password: '%database_password%'
         # charset:  UTF8
 
     orm:
-        auto_generate_proxy_classes: "%kernel.debug%"
+        auto_generate_proxy_classes: '%kernel.debug%'
         auto_mapping: true
 
 # Swiftmailer Configuration (used to send emails)
@@ -83,8 +83,8 @@ doctrine:
 # stores options that change from one server to another
 #
 swiftmailer:
-    transport: "%mailer_transport%"
-    host:      "%mailer_host%"
-    username:  "%mailer_user%"
-    password:  "%mailer_password%"
+    transport: '%mailer_transport%'
+    host:      '%mailer_host%'
+    username:  '%mailer_user%'
+    password:  '%mailer_password%'
     spool:     { type: memory }
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -44,7 +44,7 @@ framework:
     trusted_hosts:   ~
     session:
         handler_id:  session.handler.native_file
-        save_path:   "%kernel.root_dir%/../var/sessions/%kernel.environment%"
+        save_path:   "%kernel.project_dir%/var/sessions/%kernel.environment%"
     fragments:       ~
     http_method_override: true
     assets: ~
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -42,7 +42,6 @@ framework:
         engines: ['twig']
     default_locale:  "%locale%"
     trusted_hosts:   ~
-    trusted_proxies: ~
     session:
         handler_id:  session.handler.native_file
         save_path:   "%kernel.root_dir%/../var/sessions/%kernel.environment%"
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -43,5 +43,5 @@ services:
             $locales: '%app_locales%'
             $defaultLocale: '%locale%'
 
-    Twig_Extensions_Extension_Intl:
-        class:  Twig_Extensions_Extension_Intl
+
+    Twig\Extensions\IntlExtension: ~
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -1,83 +1,46 @@
 services:
-    # First we define some basic services to make these utilities available in
-    # the entire application
-    slugger:
-        class: AppBundle\Utils\Slugger
+    # default configuration for services in *this* file
+    _defaults:
+        # automatically injects dependencies in your services
+        autowire: true
+        # automatically registers your services as commands, event subscribers, etc.
+        autoconfigure: true
+        # this means you cannot fetch services directly from the container via $container->get()
+        # if you need to do this, you can override this setting on individual services
+        public: false
 
-    markdown:
-        class: AppBundle\Utils\Markdown
+    # makes classes in src/AppBundle available to be used as services
+    # this creates a service per class whose id is the fully-qualified class name
+    AppBundle\:
+        resource: '../../src/AppBundle/*'
+        # you can exclude directories or files
+        # but if a service is unused, it's removed anyway
+        exclude: '../../src/AppBundle/{Entity,Repository}'
 
-    # These are the Twig extensions that create new filters and functions for
-    # using them in the templates
-    app.twig.app_extension:
-        public:    false
-        class:     AppBundle\Twig\AppExtension
-        arguments: ['@markdown', '%app_locales%']
-        tags:
-            - { name: twig.extension }
+    # controllers are imported separately to make sure they're public
+    # and have a tag that allows actions to type-hint services
+    AppBundle\Controller\:
+        resource: '../../src/AppBundle/Controller'
+        public: true
+        tags: ['controller.service_arguments']
 
-    app.twig.intl_extension:
+    # Autowiring can't guess the constructor arguments that are not type-hinted with
+    # classes (e.g. container parameters) so you must define those arguments explicitly
+    AppBundle\Twig\AppExtension:
+        arguments:
+            $locales: '%app_locales%'
+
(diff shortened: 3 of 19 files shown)


javiereguiluz:
Commands still extend from `ContainerAwareCommand` and get services with `$container->get()` What can we do about that? Thanks!

weaverryan:
@javiereguiluz I'm a bit undecided about commands (what to officially recommend). In the docs, I've left the `$this->getContainer()->get()` in the commands, and just mentioned that you can use DI normally now that these are services. So, the answer would be, if you want, you can absolutely add a `__construct()` method to these commands and use normal DI.

app/config/services.yml
```diff
@@ -1,83 +1,46 @@
...
+        # you can exclude directories or files
+        # but if a service is unused, it's removed anyway
+        exclude: '../../src/AppBundle/{Entity,Repository}'
+
+    # controllers are imported separately to make sure they're public
+    # and have a tag that allows actions to type-hint services
+    AppBundle\Controller\:
+        resource: '../../src/AppBundle/Controller'
+        public: true
+        tags: ['controller.service_arguments']
```
yceruto: IMO this is not needed if our controllers extend from `Controller` so it's already auto-tagged.

weaverryan: You're right... but it's nice to be consistent with the official project skeleton

app/config/services.yml
```diff
@@ -1,83 +1,46 @@
...
+        public: false
 
-    app.twig.intl_extension:
+    # makes classes in src/AppBundle available to be used as services
+    # this creates a service per class whose id is the fully-qualified class name
+    AppBundle\:
+        resource: '../../src/AppBundle/*'
+        # you can exclude directories or files
+        # but if a service is unused, it's removed anyway
+        exclude: '../../src/AppBundle/{Entity,Repository}'
```
stof: you should also exclude Controller here IMO. It is a waste of resources to include them all here to overwrite them all just below

app/config/services.yml
```diff
@@ -1,83 +1,46 @@
...
+        arguments:
+            $sender: '%app.notifications.email_sender%'
+
+    AppBundle\EventListener\RedirectToPreferredLocaleSubscriber:
+        arguments:
+            $locales: '%app_locales%'
+            $defaultLocale: '%locale%'
+
+    Twig_Extensions_Extension_Intl:
         public: false
```
stof: `public: false` is useless due to the default

app/config/services.yml
```diff
@@ -1,83 +1,46 @@
...
         class:  Twig_Extensions_Extension_Intl
         tags:
```
stof: the tag is useless to, as you have `autoconfigure: true` in the defaults of the files

yceruto:
We could use `AbstractController` instead?

app/config/config.yml
```diff
@@ -42,7 +42,6 @@ framework:
         engines: ['twig']
     default_locale:  "%locale%"
     trusted_hosts:   ~
-    trusted_proxies: ~
     session:
         handler_id:  session.handler.native_file
         save_path:   "%kernel.root_dir%/../var/sessions/%kernel.environment%"
```
yceruto: `"%kernel.project_dir%/var/sessions/%kernel.environment%"` ?

app/config/config.yml
```diff
@@ -42,7 +42,6 @@ framework:
         engines: ['twig']
     default_locale:  "%locale%"
```
yceruto: Single quote in yamls?

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -86,7 +87,7 @@ public function newAction(Request $request)
         // However, we explicitly add it to improve code readability.
         // See https://symfony.com/doc/current/best_practices/forms.html#handling-form-submits
         if ($form->isSubmitted() && $form->isValid()) {
-            $post->setSlug($this->get('slugger')->slugify($post->getTitle()));
+            $post->setSlug($slugger->slugify($post->getTitle()));
 
             $entityManager = $this->getDoctrine()->getManager();
```
yceruto: Also we could inject `em` https://github.com/symfony/symfony-docs/pull/7933/files#diff-e530d0758aab398d19d2ed98bf9d6715R107

yceruto: Or why not extend from `AbstractController` and keep this as it?

weaverryan: FYI - the recommended practice for getting the EM is being discussed here: https://github.com/symfony/symfony-docs/pull/7933#discussion_r117676576

javiereguiluz:
I've updated commands to use DI too. Please let me know other places where we can use the new features. Thanks!

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -86,11 +88,10 @@ public function newAction(Request $request)
         // However, we explicitly add it to improve code readability.
         // See https://symfony.com/doc/current/best_practices/forms.html#handling-form-submits
         if ($form->isSubmitted() && $form->isValid()) {
-            $post->setSlug($this->get('slugger')->slugify($post->getTitle()));
+            $post->setSlug($slugger->slugify($post->getTitle()));
 
-            $entityManager = $this->getDoctrine()->getManager();
-            $entityManager->persist($post);
-            $entityManager->flush();
+            $em->persist($post);
```
yceruto: `$em` missing injection?

src/AppBundle/Command/ListUsersCommand.php
```diff
@@ -34,12 +34,19 @@
...
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  */
-class ListUsersCommand extends ContainerAwareCommand
+class ListUsersCommand extends Command
 {
-    /**
-     * @var ObjectManager
-     */
     private $entityManager;
+    private $mailer;
```
GuilhemN: missing `$emailSender` property

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -134,19 +135,16 @@ public function showAction(Post $post)
      * @Route("/{id}/edit", requirements={"id": "\d+"}, name="admin_post_edit")
      * @Method({"GET", "POST"})
      */
-    public function editAction(Post $post, Request $request)
+    public function editAction(Request $request, EntityManagerInterface $em, Post $post, Slugger $slugger)
```
GuilhemN: So we order arguments this way: `Request`>`$em`>entities>other dependencies?

javiereguiluz: I don't know. To me Request and the Doctrine thing look better at the beginning. But this is just an opinion.

GuilhemN: I see Doctrine as a regular dependency so I would put it at the end. 
Anyway we can decide this later in https://github.com/symfony/symfony-docs/pull/7933#discussion_r117676576, let's keep it as is for now :)

javiereguiluz:
Ready for the final review before merge. Thanks!

src/AppBundle/DataFixtures/ORM/PostFixtures.php
```diff
@@ -47,6 +47,7 @@ public function load(ObjectManager $manager)
 
             $post->setTitle($title);
             $post->setSummary($this->getRandomPostSummary());
+            // TODO: fix this 'slugger' call
```
GuilhemN: Planed for another pr ?

javiereguiluz: Good catch! Maybe I'm too naive, but I'm trying to do this and it doesn't work:

```php
class PostFixtures extends AbstractFixture implements DependentFixtureInterface, ContainerAwareInterface
{
    use ContainerAwareTrait;
    use FixturesTrait;

    private $slugger;

    public function __construct(Slugger $slugger)
    {
        $this->slugger = $slugger;
    }

    // ...
}
```

stof: Well, fixtures are instantiated by Doctrine, so you cannot do autowiring there

javiereguiluz: Yet another exception for where autowiring works. I don't like it.

stof: Autowiring works for services. Fixture classes have nothing to do with the DI container

weaverryan:
About fixtures, the fixtures bundle should probably be updated to allow for fixtures to be a tagged service. The fact that you can't use autowiring for them is due to this shortcoming, which is with the bundle imo

app/config/services.yml
```diff
@@ -1,83 +1,47 @@
...
+        arguments:
+            $sender: '%app.notifications.email_sender%'
+
+    AppBundle\EventListener\RedirectToPreferredLocaleSubscriber:
+        arguments:
+            $locales: '%app_locales%'
+            $defaultLocale: '%locale%'
+
+    Twig_Extensions_Extension_Intl:
+        class:  Twig_Extensions_Extension_Intl
```
weaverryan: If you upgrade to version 1.5.0 of Twig extensions, you can now use `Twig\Extensions\IntlExtension` (and so you won't need the `class` key

src/AppBundle/Controller/BlogController.php
```diff
@@ -65,7 +68,7 @@ public function indexAction($page, $_format)
      * value given in the route.
      * See https://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/converters.html
      */
-    public function postShowAction(Post $post)
+    public function postShowAction(Post $post, UserInterface $user = null)
```
weaverryan: Remove this too

src/AppBundle/EventListener/CommentNotificationSubscriber.php
```diff
@@ -59,6 +60,13 @@ public function __construct(\Swift_Mailer $mailer, UrlGeneratorInterface $urlGen
         $this->sender = $sender;
     }
 
+    public static function getSubscribedEvents()
+    {
+        return [
+            'comment.created' => 'onCommentCreated',
+        ];
```
weaverryan: Do you want to use `Events::COMMENT_CREATED`?

src/AppBundle/Form/Type/DateTimePickerType.php
```diff
@@ -30,9 +30,9 @@ class DateTimePickerType extends AbstractType
 {
     private $formatConverter;
 
-    public function __construct()
+    public function __construct(MomentFormatConverter $converter)
     {
-        $this->formatConverter = new MomentFormatConverter();
+        $this->formatConverter = $converter;
```
weaverryan: Hey, this is a cool benefit of the new system :). There is no longer any extra work to do things correctly :)

src/AppBundle/Controller/Admin/BlogController.php
```diff
@@ -52,10 +55,9 @@ class BlogController extends Controller
      * @Route("/", name="admin_post_index")
      * @Method("GET")
      */
-    public function indexAction()
+    public function indexAction(EntityManagerInterface $em, UserInterface $user = null)
     {
-        $entityManager = $this->getDoctrine()->getManager();
-        $posts = $entityManager->getRepository(Post::class)->findBy(['author' => $this->getUser()], ['publishedAt' => 'DESC']);
+        $posts = $em->getRepository(Post::class)->findBy(['author' => $user], ['publishedAt' => 'DESC']);
```
yceruto: Also I prefer `$this->getDoctrine()->getRepository()` shortcut if `$em` is not needed :)

app/config/services.yml
javiereguiluz: So nice! Thanks.

src/AppBundle/Controller/Admin/BlogController.php
javiereguiluz: I reverted `EntityManagerInterface` and `UserInterface` autowiring.

.travis.yml
```diff
@@ -36,5 +36,7 @@ script:
   - ./bin/console lint:yaml @CodeExplorerBundle
   # this checks that the Twig template files contain no syntax errors
   - ./bin/console lint:twig app/Resources @CodeExplorerBundle
+  # this checks that the XLIFF translations contain no syntax errors
+  - ./bin/console lint:xliff app/Resources
```
bocharsky-bw: What about more specific `app/Resources/translations`? We have translations only in that folder

app/config/services.yml
```diff
@@ -1,83 +1,47 @@
...
+        arguments:
+            $sender: '%app.notifications.email_sender%'
+
+    AppBundle\EventListener\RedirectToPreferredLocaleSubscriber:
+        arguments:
+            $locales: '%app_locales%'
+            $defaultLocale: '%locale%'
+
+
+    Twig\Extensions\IntlExtension: ~
```
bocharsky-bw: Is it necessary? Where is it used?

src/AppBundle/Command/ListUsersCommand.php
```diff
@@ -34,12 +34,20 @@
...
+    private $emailSender;
```
bocharsky-bw: I'd add `@param string` annotation since `$emailSender` does not have type hint and can't be resolved by IDE.

app/config/services.yml
javiereguiluz: If I remove this service definition, I see this error:

localizeddate-filter

src/AppBundle/Command/ListUsersCommand.php
javiereguiluz: You are right ... but at the same time, the type-hint is useful for objects or arrays of objects. In this case, even if the IDE knows this is a string ... there's nothing to complete or help us, right?

app/config/services.yml
bocharsky-bw: Ah, what about adding a small note to this service definition where say it's used by `localizeddate` filter in Twig templates?

composer.json
```diff
@@ -29,8 +29,8 @@
         "symfony/monolog-bundle"               : "^3.0",
         "symfony/polyfill-apcu"                : "^1.0",
         "symfony/swiftmailer-bundle"           : "^2.3",
-        "symfony/symfony"                      : "^3.2",
-        "twig/extensions"                      : "^1.3",
+        "symfony/symfony"                      : "3.3.0-RC1",
```
bocharsky-bw: Now we can use stable `3.3.0`

javiereguiluz:
After upgrading to 3.3.0 final, I see these deprecations:

symfony-3-3-deprecations

PR #564, opened 2017-05-24
Removed HHVM from Travis CI test matrix

javiereguiluz:
Following the decision of Symfony to drop HHVM support.


stof:
you should also add a `nightly` job to follow the Symfony decision 😄

javiereguiluz:
I'm not sure it's needed for this repo. In any case, it should be done in a different PR. Thanks!

Issue #472, opened 2017-02-14
Update commands to use the SymfonyStyle


javiereguiluz:
There is no need to show how to work with the low level `Question`, `Table`, etc. helpers. Let's just use `$io` everywhere.

javiereguiluz:
Closing ti as fixed by #566.

PR #566, opened 2017-05-27
Use Symfony console style

elkuku:
Ref.: #472

This will  use the `SymfonyStyle` style for console commands.

Unfortunately the advanced features used are no supported so we can't really live without the "low level" helpers...
* `Question` max attempts
* `Table` using `BufferedOutput` to "capture" the table

I have removed the "max attempts" feature from the `delete-user` command and was also tempted to remove this behavior in the `add-user` command on the fields that doesn't do much validation (user name and full name) - WDYT?


javiereguiluz:
@elkuku thanks for this contribution! I've merged this 515c00e6818a8c3c38ed9478ad8207fe782dfff2 and I'll create a new PR to propose some minor changes to this. Thanks!

PR #567, opened 2017-05-27
Validator service

elkuku:
This will move the validator methods in the console commands to a service.

Currently there is a bug caused by the fact that the `delete-user` command uses a different method to validate the username than the `add-user` command.

So you can actually **add** a user but you can't **remove** the user ...

<del>Later in 3.3 this can be nicely injected in the constructor</del> done :wink:

### DIFF
--- a/src/AppBundle/Utils/Validator.php
+++ b/src/AppBundle/Utils/Validator.php
@@ -17,7 +17,6 @@ class Validator
      * @param string $username
      *
      * @return string
-     * @throws \Exception
      */
     public function username($username)
     {
@@ -36,7 +35,6 @@ class Validator
      * @param string $plainPassword
      *
      * @return string
-     * @throws \Exception
      */
     public function password($plainPassword)
     {
@@ -55,7 +53,6 @@ class Validator
      * @param string $email
      *
      * @return string
-     * @throws \Exception
      */
     public function email($email)
     {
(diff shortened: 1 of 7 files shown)


src/AppBundle/Utils/Validator.php
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace AppBundle\Utils;
+
+class Validator
+{
+    public function username($username)
```
uikolas: Maybe the name could be: `usernameValidation`?

Koc: validateUsername

elkuku: Finding the "right" name has always been the hardest part in software design :wink:

voronkovich: @elkuku, You right, but a method name in most cases should begin with a verb. :)

elkuku:
This has been updated and corresponding unit tests have been added.

javiereguiluz:
Closing in favor of #615. @elkuku I'm sorry but I couldn't fix the conflicts of your PR while merging, so I needed to create a different PR. Thanks!

PR #569, opened 2017-05-30
Removed the deprecated getName() method in Twig extensions

javiereguiluz:
```
        /**
         * Returns the name of the extension.
         *
         * @return string The extension name
         *
         * @deprecated since 1.26 (to be removed in 2.0), not used anymore internally
         */
        function getName();
```


stof:
👍

PR #572, opened 2017-05-30
Use DSN-like strings to define credentials

javiereguiluz:
This is the new best practice for modern Symfony apps. See https://github.com/symfony/recipes/blob/master/symfony/swiftmailer-bundle/2.5/etc/packages/swiftmailer.yaml

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -10,7 +10,7 @@
     <php>
         <ini name="error_reporting" value="-1" />
         <server name="KERNEL_DIR" value="app/" />
-        <env name="DATABASE_URL" value="sqlite://%kernel.project_dir%/var/data/blog_test.sqlite"/>
+        <env name="DATABASE_URL" value="sqlite:///%kernel.project_dir%/var/data/blog_test.sqlite"/>
     </php>
 
     <testsuites>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -10,6 +10,7 @@
     <php>
         <ini name="error_reporting" value="-1" />
         <server name="KERNEL_DIR" value="app/" />
+        <env name="DATABASE_URL" value="sqlite://%kernel.project_dir%/var/data/blog_test.sqlite"/>
     </php>
 
     <testsuites>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -10,7 +10,7 @@
     <php>
         <ini name="error_reporting" value="-1" />
         <server name="KERNEL_DIR" value="app/" />
-        <env name="DATABASE_URL" value="sqlite:///%kernel.project_dir%/var/data/blog_test.sqlite"/>
+        <env name="DATABASE_URL" value="sqlite:///var/data/blog_test.sqlite"/>
     </php>
 
     <testsuites>
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -62,16 +62,6 @@ doctrine:
     dbal:
         # if you don't want to use SQLite, change the URL in parameters.yml or set the DATABASE_URL environment variable
         url: '%env(DATABASE_URL)%'
-
-        # instead of using a URL, you may also uncomment the following lines to configure your database
-        # driver:   pdo_mysql
-        # host:     '%database_host%'
-        # port:     '%database_port%'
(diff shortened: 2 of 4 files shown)


javiereguiluz:
I need someone to explain me why I needed to make this change --> f3d8340 to make this work. Thanks!

stof:
Well, your second commit is because what you defined was not a URL-based DSN for SQLite. It was only a path.
If you want to use it in `url`, use `sqlite://%kernel.project_dir%/...`

javiereguiluz:
@stof that was the problem! I'm an idiot 😊 

However, if I change that, I now see this:

```
Symfony\Component\DependencyInjection\Exception\EnvParameterException:
Incompatible use of dynamic environment variables "DATABASE_URL" found in parameters.

symfony-demo/vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php:182
symfony-demo/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:743
symfony-demo/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:569
symfony-demo/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:119
symfony-demo/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php:157
symfony-demo/tests/AppBundle/Command/AddUserCommandTest.php:104
symfony-demo/tests/AppBundle/Command/AddUserCommandTest.php:41
```

stof:
hmm, this message is not clear

voronkovich:
@javiereguiluz, dumper throws an exception because you don't use the `DATABASE_URL` variable in the `test` environment.  See https://github.com/symfony/symfony/blob/v3.3.0/src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php#L182

voronkovich:
@javiereguiluz, see @dmaicher's comment https://github.com/symfony/symfony-demo/pull/557#issuecomment-300897014

javiereguiluz:
@voronkovich thanks for your comments. That was indeed the error and the proposed fix solved the issue. Thanks!

phpunit.xml.dist
```diff
@@ -10,6 +10,7 @@
     <php>
         <ini name="error_reporting" value="-1" />
         <server name="KERNEL_DIR" value="app/" />
+        <env name="DATABASE_URL" value="sqlite://%kernel.project_dir%/var/data/blog_test.sqlite"/>
```
stof: I don't think you can use a parameter here, in an actual env var

dmaicher: Actually you are right but it still works :smile: 

Really weird what happens now when dumping the connection within a test to see the config:

```
  -_params: array:9 [
    "url" => "sqlite://%kernel.project_dir%/var/data/blog_test.sqlite"
    "host" => "%kernel.project_dir%"
    "port" => null
    "user" => "root"
    "password" => null
    "driver" => "pdo_sqlite"
    "driverOptions" => []
    "defaultTableOptions" => []
    "path" => "var/data/blog_test.sqlite"
  ]
```

This seems to be more correct:

```xml
        <env name="DATABASE_URL" value="sqlite:///var/data/blog_test.sqlite" />
```

==>

```
  -_params: array:9 [
    "url" => "sqlite:///var/data/blog_test.sqlite"
    "host" => "localhost"
    "port" => null
    "user" => "root"
    "password" => null
    "driver" => "pdo_sqlite"
    "driverOptions" => []
    "defaultTableOptions" => []
    "path" => "var/data/blog_test.sqlite"
  ]
```

dmaicher: @javiereguiluz FYI see above

javiereguiluz: @dmaicher you are right. I fixed that in 464e786 but I think Christophe is right and container params don't work here.

voronkovich: @javiereguiluz, compiled container doesn't interpolate parameters in environment variables.  See https://github.com/symfony/symfony/blob/v3.3.0/src/Symfony/Component/DependencyInjection/Container.php#L442

javiereguiluz:
Thank you all for your help making this mergeable!

PR #573, opened 2017-05-30
Override DATABASE_URL by setting default parameter's value

voronkovich:
This approach is more flexible than we used in #572

### DIFF
--- a/app/config/config_test.yml
+++ b/app/config/config_test.yml
@@ -1,6 +1,13 @@
 imports:
     - { resource: config_dev.yml }
 
+parameters:
+    # It's recommended to use a separate database for tests. This allows to have a
+    # fixed and known set of data fixtures, it simplifies the code of tests and it
+    # makes them more robust.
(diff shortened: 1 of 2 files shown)


dmaicher:
Indeed much nicer :+1:

Rewieer:
Just wondering... now to use a testing database we must define this in an environment variable set in a `phpunit.xml.dist` which is [suggested](https://github.com/symfony/recipes/issues/41) to be committed with the project. But these credentials are unique to each machine, And I certainly don't want to commit my credentials. 

I know it's an old commit but being concerned with this issue I want to point that using a good .env file for credentials seems to be a better fit than using a commited phpunit file.

uuf6429:
@rewieer the point of the `phpunit.xml.dist` is to **not commit** your specific changes.

You can have your own `phpunit.xml` that overrides changes. Eg; you can remove the entry from your customer xml file and user the environment variable provided by your machine.

PR #574, opened 2017-05-30
Fix console error listener when there is no command yet

ogizanagi:
Fixes an issue when there isn't any command yet and an error was triggered. For instance, when trying to use a non-existing command.

> PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function getName() on null in src/AppBundle/EventListener/CheckRequirementsSubscriber.php:70


javiereguiluz:
Very nice catch! Thanks @ogizanagi

PR #575, opened 2017-05-31
Showcase short definition syntax

ogizanagi:


app/config/services.yml
```diff
@@ -30,18 +30,17 @@ services:
...
     AppBundle\EventListener\RedirectToPreferredLocaleSubscriber:
-        arguments:
-            $locales: '%app_locales%'
-            $defaultLocale: '%locale%'
+        $locales: '%app_locales%'
+        $defaultLocale: '%locale%'
 
     # needed for the localizeddate Twig filter
-    Twig\Extensions\IntlExtension: ~
+    Twig\Extensions\IntlExtension: ~ # same as []
```
voronkovich: @ogizanagi, I think it should be `same as null`

app/config/services.yml
```diff
@@ -30,18 +30,17 @@ services:
...
     AppBundle\Twig\AppExtension:
-        arguments:
-            $locales: '%app_locales%'
+        $locales: '%app_locales%'
 
-    AppBundle\EventListener\CommentNotificationSubscriber:
-        arguments:
-            $sender: '%app.notifications.email_sender%'
+    # The short syntax can also be inlined:
+    AppBundle\EventListener\CommentNotificationSubscriber: { $sender: '%app.notifications.email_sender%' }
```
bocharsky-bw: I'd prefer to be consistent and use the same syntax in the whole app, i.e.:
```
    AppBundle\EventListener\CommentNotificationSubscriber:
        $sender: '%app.notifications.email_sender%'
```

You explain the Yaml syntax here, which is not related to the Symfony Demo I think.

app/config/services.yml
ogizanagi: You're right. `~` means `null` using yaml. 
But actually, configuring `null` for a service definition does not mean anything, and [is converted to an empty array behind the scene](https://github.com/symfony/symfony/pull/21214).

I felt like adding this comment would help the user to understand the service definition is either the full mapping of definition's attributes (`{ attributes: [...], tags: [...], public: false, ... }`) or directly the list (or mapping for named arguments) of constructor parameters.

But maybe that's not necessary.

app/config/services.yml
ogizanagi: Sure, let's do this. :)

javiereguiluz:
@ogizanagi thanks for this! I like it ... and I didn't remember this shortcut syntax, so it's great to re-learn it!

PR #576, opened 2017-05-31
Use 3.3 DI features in CodeExplorerBundle

ogizanagi:
Since 3.3, [listeners do not have to be public anymore](https://github.com/symfony/symfony/pull/20953/files#diff-8604f72b4ae78c6c364c1af8c1dee17cL63).
I also believe declaring services private by default by adding `_defaults: { public: false }` on top will become a best practice, so I think we should use it in the CodeExplorerBundle.

Actually, we could also use classnamed services, autoconfigure and autowiring, but at least for the last one, I think we should keep a place in this application where services are declared entirely manually (autoconfigure is debatable this way too, that's true).

WDYT?

### DIFF
--- a/src/CodeExplorerBundle/EventListener/ControllerListener.php
+++ b/src/CodeExplorerBundle/EventListener/ControllerListener.php
@@ -12,22 +12,18 @@
 namespace CodeExplorerBundle\EventListener;
 
 use CodeExplorerBundle\Twig\SourceCodeExtension;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
 
 /**
  * Defines the method that 'listens' to the 'kernel.controller' event, which is
  * triggered whenever a controller is executed in the application.
  *
- * See https://symfony.com/doc/current/book/internals.html#kernel-controller-event
- *
- * Tip: listeners are common in Symfony applications, but this particular listener
- * is too advanced and too specific for the demo application needs. For more common
- * examples see https://symfony.com/doc/current/cookbook/service_container/event_listener.html
- *
  * @author Ryan Weaver <weaverryan@gmail.com>
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  */
-class ControllerListener
+class ControllerListener implements EventSubscriberInterface
 {
     private $twigExtension;
 
@@ -44,5 +40,15 @@ class ControllerListener
         if ($event->isMasterRequest()) {
             $this->twigExtension->setController($event->getController());
         }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getSubscribedEvents()
+    {
+        return [
+            KernelEvents::CONTROLLER => 'registerCurrentController',
+        ];
     }
 }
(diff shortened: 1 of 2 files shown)


src/CodeExplorerBundle/Resources/config/services.yml
```diff
@@ -1,9 +1,9 @@
 services:
+    _defaults: { public: false }
+
     code_explorer.twig.source_code_extension:
-        public:    false
-        class:     CodeExplorerBundle\Twig\SourceCodeExtension
-        tags:
-            - { name: twig.extension }
+        class: CodeExplorerBundle\Twig\SourceCodeExtension
+        tags: [twig.extension]
```
stof: use autoconfiguration to simplify the file too. almost everything can be removed then

ogizanagi: Sure, if there is a consensus on this, I'd be pleased to add autoconfigure and classnamed services here, as explained in the description (I just want to gather opinions here first) :)

stof: Well, AppBundle was migrated to use all new features. I think CodeExplorerBundle should be updated too

javiereguiluz: At first I thought as Maxime: let's show the new Symfony in AppBundle and the old Symfony in CodeExplorerBundle. But now I'm starting to think as Christophe. If this is the new and recommended Symfony ... why show the old Symfony too?

Besides, before November we'll delete the CodeExplorerBundle (and Appbundle too!) because we'll use Symfony 4 and there are no bundles in that version!

ogizanagi:
Following https://github.com/symfony/symfony-demo/pull/576#pullrequestreview-41300264, I've updated the PR to use most of the new 3.3 DI features. But I still avoided to use PSR-4 autoloading & autowiring as we did for the AppBundle.

> At first I thought as Maxime: let's show the new Symfony in AppBundle and the old Symfony in CodeExplorerBundle. But now I'm starting to think as Christophe. If this is the new and recommended Symfony ... why show the old Symfony too?

Because we could consider this bundle as a more advanced entrypoint for intermediate developers, while the AppBundle is the main show case for newcomers. I still think it's important to show how to wire things manually and demonstrate the Symfony DI & its configuration capabilities. Developers can learn a lot by going deeper on how things work behind the scene (actually, it's not even that far, its only basic configuration we wrote since years. Still, it feels important to me).

But if my concerns aren't shared, I can still update the PR to use PSR-4 autoloading & autowiring on the whole `src` directory instead. Let me know 😄

src/CodeExplorerBundle/EventListener/ControllerListener.php
```diff
@@ -12,22 +12,24 @@
...
  * triggered whenever a controller is executed in the application.
  *
  * See https://symfony.com/doc/current/book/internals.html#kernel-controller-event
  *
- * Tip: listeners are common in Symfony applications, but this particular listener
- * is too advanced and too specific for the demo application needs. For more common
- * examples see https://symfony.com/doc/current/cookbook/service_container/event_listener.html
+ * Tip: listeners and subscribers are common in Symfony applications, but this particular
+ * one is too advanced and too specific for the demo application needs.
+ * For more common examples see https://symfony.com/doc/current/event_dispatcher.html
```
stof: I would remove this extended doc. We already have it on the subscribers in the AppBundle. People digging into the internals of CodeExplorerBundle should already know what a subscriber is. And it will avoid having to maintain this message multiple time.

ogizanagi: Removed

javiereguiluz:
Thanks @ogizanagi.

PR #577, opened 2017-06-01
Update logs configuration

ogizanagi:
to match the standard edition configuration, which is way more convenient.
In dev, it'll now log every debug information but events, console errors are now logged natively and we don't need them to be output on stderr (we already have the stacktrace when running the console), and the `server:log` command allows to see logs in real-time.

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -14,12 +14,20 @@ web_profiler:
 monolog:
     handlers:
         main:
-            type:   stream
-            path:   "%env(SYMFONY_LOG)%"
-            level:  info
+            type: stream
+            path: '%env(LOG_URL)%'
+            level: debug
+            channels: ['!event']
         console:
-            type:   console
-            bubble: false
+            type: console
+            process_psr_3_messages: false
+            channels: ['!event', '!doctrine', '!console']
+        # To follow logs in real time, execute the following command:
+        # `bin/console server:log -vv`
+        server_log:
+            type: server_log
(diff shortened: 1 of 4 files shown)


app/config/config_dev.yml
```diff
@@ -14,12 +14,20 @@ web_profiler:
 monolog:
     handlers:
         main:
-            type:   stream
-            path:   "%env(SYMFONY_LOG)%"
-            level:  info
+            type: stream
+            path: "%env(SYMFONY_LOG)%"
```
yceruto: As convention we're using single quotes ;)

voronkovich: @ogizanagi, what about to use a name like `LOG_URL` for consistency with others variables names (`DATABASE_URL`, `MAILER_URL`)?

yceruto: @voronkovich In any case it would be `APP_LOG_PATH` (as it's the path to the log file), but I'm no sure why is it configurable here, so by default it's `'%kernel.logs_dir%/%kernel.environment%.log'` :/

ogizanagi: `LOG_URL` makes sense to me. It's not only a path, it's a url used by the StreamHandler: https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/StreamHandler.php#L50

javiereguiluz:
@ogizanagi thanks for these changes. They are great and, as you said, they align nicely with the default Symfony practices.

PR #578, opened 2017-06-01
to match symfony-standard autoload path

mstrom:


javiereguiluz:
I love getting rid of files like the custom autoloader. And as you said, this aligns with the changes of the Symfony Standard edition. Thanks!

PR #579, opened 2017-06-03
Update sl translation

petk:
This patch updates Slovenian (sl) translation with missing translations. Thank you.


javiereguiluz:
Thank you Peter.

PR #581, opened 2017-06-05
Alias the slugger to fix fixtures loading

ogizanagi:
Fixes current exception:

```php
[Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]         
  You have requested a non-existent service "slugger". Did you mean this: "logger"?  
```

`ContainerAwareInterface` still is the way to access services inside fixture loaders. Maybe something could be done on the doctrine-fixtures-bundle side, but for now we need to have a public alias for this service.
We could also exclude the `DataFixtures\ORM` folder from PSR-4 loading, but it's probably not worth it and may "overwhelm" a little newcomers:

```diff
-        exclude: '../../src/AppBundle/{Controller,Entity,Repository}'
+        exclude: '../../src/AppBundle/{Controller,Entity,Repository,DataFixtures/ORM}'
```

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -44,3 +44,9 @@ services:
 
     # needed for the localizeddate Twig filter
     Twig\Extensions\IntlExtension: ~
+
+    # the slugger service needs a public alias for getting it from
+    # the container when loading doctrine fixtures
+    slugger:
+        alias: AppBundle\Utils\Slugger
+        public: true


yceruto:
Somehow this is working here https://github.com/symfony/symfony-demo/pull/580 without doing it public :/

ogizanagi:
@yceruto : Sure it's working (I did that first, but then remembered we default to private services...so, was weird). As shown in https://github.com/symfony/symfony-demo/pull/580#issuecomment-306258476, this is because sf 3.3 still support getting private services, but we shoudl avoid relying on this as it's now deprecated and will be removed in 4.0.

yceruto:
> [...]  this is because sf 3.3 still support getting private services, but we should avoid relying on this as it's now deprecated and will be removed in 4.0.

Yep! I forgot that :weary:

app/config/services.yml
```diff
@@ -44,3 +44,9 @@ services:
 
     # needed for the localizeddate Twig filter
     Twig\Extensions\IntlExtension: ~
+
+    # the slugger service needs a public alias for getting it from
+    # the container when loading doctrine fixtures
+    slugger:
```
bocharsky-bw: Why not the new syntax?
```yml
AppBundle\Utils\Slugger:
     public: true
```
and use `Slugger::class` in `PostFixtures`.

ogizanagi: Whatever solution is good to me. I only choose this as it requires less diff, and the `AppBundle\Utils\Slugger` service does not really need to be public. It's only the doctrine fixtures that use the container as a service locator, so a proper alias looks better to me.

javiereguiluz:
I must say that I don't fully like this solution ... but it looks like it's the best way Symfony allows us to solve this. I'd like to be able to use services inside fixtures as in any other place.

PR #584, opened 2017-06-11
Add missing pt_BR validator translation

davialexandre:


javiereguiluz:
Thanks @davialexandre.

PR #586, opened 2017-06-12
Use Webpack Encore to manage assets

javiereguiluz:
Most of the features are working ... but I'm having troubles with the autocompletion of the lables. It looks horrible and it doesn't autocomplete anything:

tags-autocomplete

I know that the `admin.js` committed in this PR is wrong, but I just wanted to show the last I tried. I've spent most of this morning trying to make this work 😢 

Maybe @yceruto, who added this feature originally, or @weaverryan, who created Webpack Encore, can help us here. Thanks!


javiereguiluz:
When I try to compile the production assets, I get this error:

```
$ ./node_modules/.bin/encore production
Running webpack ...

 ERROR  Failed to compile with 8 errors                                                                                                                                                                                      7:59:36 PM

These relative modules were not found:

* ../fonts/lato-normal/lato-normal.woff2 in ./~/css-loader!./~/resolve-url-loader!./~/sass-loader/lib/loader.js?{"sourceMap":true}!./app/Resources/assets/scss/app.scss
* ../fonts/lato-normal/lato-normal.woff in ./~/css-loader!./~/resolve-url-loader!./~/sass-loader/lib/loader.js?{"sourceMap":true}!./app/Resources/assets/scss/app.scss
* ../fonts/lato-normal-italic/lato-normal-italic.woff2 in ./~/css-loader!./~/resolve-url-loader!./~/sass-loader/lib/loader.js?{"sourceMap":true}!./app/Resources/assets/scss/app.scss
* ../fonts/lato-normal-italic/lato-normal-italic.woff in ./~/css-loader!./~/resolve-url-loader!./~/sass-loader/lib/loader.js?{"sourceMap":true}!./app/Resources/assets/scss/app.scss
* ../fonts/lato-bold/lato-bold.woff2 in ./~/css-loader!./~/resolve-url-loader!./~/sass-loader/lib/loader.js?{"sourceMap":true}!./app/Resources/assets/scss/app.scss
* ../fonts/lato-bold/lato-bold.woff in ./~/css-loader!./~/resolve-url-loader!./~/sass-loader/lib/loader.js?{"sourceMap":true}!./app/Resources/assets/scss/app.scss
* ../fonts/lato-bold-italic/lato-bold-italic.woff2 in ./~/css-loader!./~/resolve-url-loader!./~/sass-loader/lib/loader.js?{"sourceMap":true}!./app/Resources/assets/scss/app.scss
* ../fonts/lato-bold-italic/lato-bold-italic.woff in ./~/css-loader!./~/resolve-url-loader!./~/sass-loader/lib/loader.js?{"sourceMap":true}!./app/Resources/assets/scss/app.scss
```

weaverryan:
See https://github.com/symfony/webpack-encore/pull/4 for production error issue :)

weaverryan:
About the tags thing, before, we had this file: https://github.com/symfony/symfony-demo/blob/master/web/css/bootstrap-tagsinput.css

But this file was deleted, and is not being loaded in the new setup. You *do* have this in `main.scss`:

```sass
@import "~bootstrap-tagsinput/src/bootstrap-tagsinput.css";
```

But that doesn't have the same contents as the old `bootstrap-tagsinput.css`. If I bring back the old contents, it "looks" correctly.

There is some other problem too, as even with this, I don't see the old tags. There's no JS error, but for some reason the autocomplete box thinks that there is never any results (I can see the class `tt-empty` in the DOM)

weaverryan:
Production issue should be fixed in 0.7.2

yceruto:
**CSS**
The actual `web/css/bootstrap-tagsinput.css` (this one includes typeahead style) is a custom stylesheet  adapted to current demo style (e.g. FontAwesome), so I suggest you use it as is, for now. Later we would include the originals overwriting the relevant classes.

**JS**
Try this:
```diff
  var source = new Bloodhound({
      local: $input.data('tags'),
      queryTokenizer: Bloodhound.tokenizers.whitespace,
      datumTokenizer: Bloodhound.tokenizers.whitespace
  });
+ source.initialize();
  $input.tagsinput({
      trimValue: true,
      focusClass: 'focus',
      typeaheadjs: {
          name: 'tags',
-         source: source
+         source: source.ttAdapter()
      }
  });
```

app/Resources/assets/js/admin.js
```diff
@@ -0,0 +1,69 @@
+var $ = require('jquery');
```
weaverryan: What about putting the `assets` directory at the root of the project? It's a bit more "flex" like

app/Resources/assets/js/admin.js
```diff
@@ -0,0 +1,69 @@
...
+window.$ = $;
+window.jQuery = $;
```
weaverryan: You shouldn't need either of these unless we still have some JavaScript code in a template somewhere that is relying on these to be globally available. The only place I could see was `login.html.twig`. We could move that little bit of JavaScript here, or into a new `login` entry as an example of a page-specific JS file.

app/Resources/assets/js/admin.js
```diff
@@ -0,0 +1,69 @@
...
+
+require('bootstrap-sass');
```
weaverryan: How about a comment above this:

```javascript
// loads the Bootstrap jQuery plugins
```

app/Resources/assets/js/admin.js
```diff
@@ -0,0 +1,69 @@
...
+require('moment');
```
weaverryan: This shouldn't be needed. I think only the `eonasdan-bootstrap-datetimepicker` library uses it, and it should (and appears to) require `moment` internally on its own https://github.com/Eonasdan/bootstrap-datetimepicker/blob/25c11d79e614bc6463a87c3dd9cbf8280422e006/src/js/bootstrap-datetimepicker.js#L42

app/Resources/assets/js/admin.js
```diff
@@ -0,0 +1,69 @@
...
+require('eonasdan-bootstrap-datetimepicker');
+
+require('imports-loader?define=>false!typeahead.js/dist/typeahead.jquery.min.js');
+const Bloodhound = require('imports-loader?define=>false!typeahead.js/dist/bloodhound.js');
+window.Bloodhound = Bloodhound;
```
weaverryan: Again, this shouldn't be needed (in theory), but it's possible there's some legacy library that isn't requiring this correctly. We'll need to make sure each line is needed, and document which lines *are* needed and why

app/Resources/assets/js/app.js
```diff
@@ -0,0 +1,13 @@
+var $ = require('jquery');
+window.$ = $;
+window.jQuery = $;
```
weaverryan: Should be able to remove lines 2 and 3

app/Resources/assets/js/app.js
```diff
@@ -0,0 +1,13 @@
...
+
+require('bootstrap-sass/assets/javascripts/bootstrap/dropdown.js');
+require('bootstrap-sass/assets/javascripts/bootstrap/modal.js');
+require('bootstrap-sass/assets/javascripts/bootstrap/transition.js');
+
+var hljs = require('highlight.js');
+hljs.registerLanguage('php', require('highlight.js/lib/languages/php'));
+hljs.registerLanguage('twig', require('highlight.js/lib/languages/twig'));
+window.hljs = hljs;
```
weaverryan: This shouldn't be needed

app/Resources/assets/js/app.js
```diff
@@ -0,0 +1,13 @@
...
+hljs.initHighlightingOnLoad();
```
weaverryan: does this need to be in a document.ready?

app/Resources/assets/scss/admin.scss
```diff
@@ -0,0 +1,26 @@
+@import "~bootswatch/flatly/variables";
+@import "~eonasdan-bootstrap-datetimepicker/src/sass/bootstrap-datetimepicker-build.scss";
+@import "~bootstrap-tagsinput/src/bootstrap-tagsinput.css";
```
weaverryan: When a CSS file is a dependency of some JS library, I usually like to require the CSS file right in our `.js` file, right next to where I require the related JS file. Later, if we stopped needing the bootstrap-datetimepicker JS file, we would also remove the CSS require right next to it.

package.json
```diff
@@ -0,0 +1,21 @@
...
+    "bootswatch": "^3.3.7",
+    "eonasdan-bootstrap-datetimepicker": "^4.17.47",
+    "font-awesome": "^4.7.0",
+    "highlight.js": "^9.12.0",
+    "imports-loader": "^0.7.1",
+    "jquery": "^3.2.1",
+    "lato-font": "^3.0.0",
+    "moment": "^2.18.1",
+    "typeahead.js": "^0.11.1"
+  }
```
weaverryan: These should all go into `devDependencies`. It doesn't really matter, but the idea is that you really only need all of this stuff while your "developing". You won't need these files on production, because you will have already built your final files and are shipping those statically.

webpack.config.js
```diff
@@ -0,0 +1,17 @@
+var Encore = require('@symfony/webpack-encore');
+
+Encore
+    .setOutputPath('web/build/')
+    .setPublicPath('/build')
+    .setManifestKeyPrefix('build')
```
weaverryan: We shouldn't need this line - but if you do, let me know :)

app/Resources/assets/js/admin.js
javiereguiluz: Yes ... but let's do that when we switch the entire app to Flex.

app/Resources/assets/js/admin.js
javiereguiluz: If I remove those, I see this error:

```
Uncaught ReferenceError: Bloodhound is not defined
    at HTMLDocument.<anonymous> (admin.js:35043)
    at mightThrow (admin.js:11908)
    at process (admin.js:11976)
```

javiereguiluz: I see a lot of errors related to this library. Maybe we can replace it with some simpler alternative library? After all, our needs are super basic.

app/Resources/assets/js/app.js
javiereguiluz: If I remove them, I see this error:

```
Uncaught TypeError: $input.tagsinput is not a function
    at HTMLDocument.<anonymous> (admin.js:35050)
    at mightThrow (admin.js:11908)
    at process (admin.js:11976)
```

package.json
stof: @weaverryan it depends how you deploy actually. When deploying on heroku, the build happens on the heroku build server, and it does not include dev deps

app/Resources/assets/js/admin.js
javiereguiluz: If I remove it, I see this error in any backend page:

```
admin.js:38009 Uncaught TypeError: Cannot read property 'fn' of undefined
    at admin.js:38009
    at Object.<anonymous> (admin.js:38134)
    at __webpack_require__ (admin.js:20)
    at Object.<anonymous> (admin.js:35021)
    at __webpack_require__ (admin.js:20)
    at Object.<anonymous> (admin.js:42417)
    at __webpack_require__ (admin.js:20)
    at module.exports.ctor.super_ (admin.js:66)
    at admin.js:69
```

app/Resources/assets/scss/app.scss
```diff
@@ -39,288 +45,264 @@ i {
 /* Page elements
    ------------------------------------------------------------------------- */
 body {
-    display: -webkit-flex;
-    display: flex;
-    -webkit-flex-direction: column;
-    flex-direction: column;
-    min-height: 100vh
+  display: -webkit-flex;
```
stof: could we use autoprefixer instead ?

app/Resources/assets/js/app.js
stof: then it means that the typeahead plugin does not support getting jQuery through CommonJS

app/Resources/assets/scss/app.scss
javiereguiluz: I propose to remove those instead. They are unneeded now.

stof: yeah, we could consider that we target modern browsers only as we target developers. But I would still consider it useful to demo how autoprefixer can be used (many sites will actually need it)

app/Resources/assets/js/admin.js
```diff
@@ -0,0 +1,69 @@
+var $ = require('jquery');
+window.$ = $;
+window.jQuery = $;
+
+require('bootstrap-sass');
+require('moment');
+require('eonasdan-bootstrap-datetimepicker');
+
+require('imports-loader?define=>false!typeahead.js/dist/typeahead.jquery.min.js');
```
stof: why using the imports-loader ? The dist files are using UMD, so they should work fine with CommonJS requirements.

I think all your issues about defining global variables are actually related to the weird loading you do here.

javiereguiluz: I have no idea ... I spent a lot of time trying to make typeahead/bloodhound work. Nothing worked. This was one of the tests that I did after reading some of the issues reported about this library (there are like a lot of issues reported!)

app/Resources/assets/js/admin.js
alOneh: add something like `.autoProvideVariables({ 'window.Bloodhound': "bloodhound")` to your webpack.config.js

stof: We might decide to change: https://github.com/twitter/typeahead.js/issues/1618

javiereguiluz:
I'm going to merge this "as is" although it's broken. That will allow others take over this and fix my mistakes. I removed all the `window.$ = $; window.jQuery = $` and now most features are broken (probably my fault!).

List of things to fix:

* Typeahead library is dead ... use a fork instead (Algolia probably is the most reassuring one)
* The bootstrap datepicker doesn't work (it worked before removing the jQuery thing)
* The highlighting doesn't work (it worked before ... but it generated a 600KB app.js because it includes ALL the highlighting languages, instead of just the PHP and Twig we need ... I tried to reduce the file size ... but now it doesn't work)
* The modal window of "Show code" works ... but the modal window of "Delete post" doesn't work. It worked yesterday (so maybe again is the jQuery thing).

---

@alOneh knows all this well and he's very skilled with Webpack (he even updated the Symfony Demo to use pure Webpack, without Encore. See: https://github.com/alOneh/sf-live-2017-symfony-webpack/blob/master/webpack.config.js) So maybe this can be fixed between @weaverryan and @alOneh

stof:
@javiereguiluz the issue with moment locales and syntax highlighting is probably the same:
webpack sees a dynamic require call loading the translation/grammar dynamically for server-side usage of the library, and so it figures out it needs to bundle all matching files to avoid breaking the client-side usage.

PR #588, opened 2017-06-14
Clean outdated committed bundle assets

stof:
FrameworkBundle and SensioDistributionBundle don't have installable assets anymore, but the deletion was not committed when updated them.


javiereguiluz:
Thanks @stof.

PR #589, opened 2017-06-14
Commit missing built asset files

stof:
Assets are meant to be committed, to make the demo work out of the box without having to run Encore. But this works only if all files gets committed. Committing only 2 of 15 files is broken.


stof:
Btw, there is not contributing doc about how to build assets AFAIK (we need to tell people to run `./node_modules/.bin/encore production` before committing to ship with the working version)

stof:
@javiereguiluz I will rebase this PR and rebuild files, as #587 has been merged.

stof:
@javiereguiluz done

javiereguiluz:
@weaverryan please merge this PR because for some unknow reason I can't do that:

```
<error>Unable to merge branch pull/589</error>.
Unable to merge branch pull/589
Resolve the conflict by hand and start the command again
```

I can't see any conflict after that message, so I can't solve nothing ... but if I don't solve something, it doesn't let me merge this  😡  😡  😡

bocharsky-bw:
@javiereguiluz Sounds like a bug in your internal tool ;) Maybe you need to make a meanless commit after this error message and try to merge again?

stof:
@javiereguiluz I think you may have the non-committed build files locally, which prevent the merge because the merge would have to overwrite untracked files.
Run a `git clean` on your project to delete untracked asset files and start again

javiereguiluz:
I've tried `git clean -f` but it doesn't work. I'm going to report the error in the repo of the tool we use to merge things. Thanks.

stof:
@javiereguiluz don't use the GH resume feature either. It was not made for the case where the merge is rejected entirely (which fails earlier in git)

javiereguiluz:
Closing it in favor of #593. @stof I'm sorry but I couldn't merge your PR in any way.

PR #594, opened 2017-06-19
Use a safer endpoint to check for security vulnerabilities

### LINKED ISSUE #543: Use a secure endpoint for the security checker command
In #536 we temporarily set the security checked endpoint to a non-https URL. Fix this when everything works again on the other side.

javiereguiluz:
This fixes #543.


javiereguiluz:
Not sure why this keeps failing. On my local computer the https:// link works again.

PR #597, opened 2017-06-20
Skip tests that require stty when it's not available

### LINKED ISSUE #530: Tests fail for AddUserCommand if shell is not a TTY
This can happen in certain CI environments etc.

To reproduce: `sh -c 'vendor/bin/phpunit tests/AppBundle/Command/' < /dev/null 2>&1`:

```
PHPUnit 4.8.35 by Sebastian Bergmann and contributors.

..stty: stdin isn't a terminal
stty: stdin isn't a terminal
stty: stdin isn't a terminal
stty: stdin isn't a terminal
stty: stdin isn't a terminal
stty: stdin isn't a terminal
stty: stdin isn't a terminal
stty: stdin isn't a terminal
stty: stdin isn't a terminal
stty: stdin isn't a terminal
E
…
```

The culprit is the "hidden" mode for the password field. I am not sure whether this is actually a bug in `Symfony\Component\Console\Helper\QuestionHelper::getHiddenResponse()`, where the error is triggered by all those `shell_exec` calls, or whether the command (or its test) should conditionally not set that hidden mode.

Pinging @nicolas-grekas for insight :) Also pinging @dmaicher who introduced the tests in #480.

javiereguiluz:
This fixes #530.

### DIFF
--- a/tests/AppBundle/Command/AddUserCommandTest.php
+++ b/tests/AppBundle/Command/AddUserCommandTest.php
@@ -26,6 +26,16 @@ class AddUserCommandTest extends KernelTestCase
         'full-name' => 'Chuck Norris',
     ];
 
+    protected function setUp()
+    {
+        // check if stty is supported, because it's needed for questions with hidden answers
+        exec('stty 2>&1', $output, $exitcode);
+
+        if (0 !== $exitcode) {
+            $this->markTestSkipped('`stty` is required to test commands.');
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


stof:
you should also skip it on windows all the time (your code may do it already, not sure), as the windows implementation is not supported by your test either AFAIK

dzuelke:
Thanks @javiereguiluz !

PR #599, opened 2017-06-22
services: added name to the quote

TomasVotruba:
So it's clear it's name of the filter and not English.

I was about to send PR with space "localized date", but then I've read the "filter". So I hope this is the correct option :)


javiereguiluz:
@TomasVotruba thanks for this. Adding quotes makes this impossible to misunderstand. Thanks!

TomasVotruba:
Thank

PR #602, opened 2017-07-02
Moving assets into a root directory for simplicity

weaverryan:
I'd like to see `assets/` put at the root directory, instead of inside `app/Resources` (which is really specific to the framework). It's a bit simpler, and I think in Flex, it will need to be this way.

Btw, the changes to the `web/build/*` files is due to a small, internal change in Encore 0.9.0 - e.g. https://github.com/symfony/symfony-demo/compare/master...weaverryan:move-assets?expand=1#diff-0d9100b0501984c386980f99e9fdb4b2

Cheers!


20uf:
Good practices, I approve.

javiereguiluz:
The new structure looks great! Thanks Ryan.

PR #604, opened 2017-07-02
Moving translations and templates folders to root folder

mstrom:
Symfony 4 will have templates folder under root folder
So I'm proposing to move templates to root folder for simplicity and translations as well


app/config/config.yml
```diff
@@ -30,7 +30,7 @@ framework:
     # and with different cache configurations for each fragment
     # https://symfony.com/doc/current/book/http_cache.html#edge-side-includes
     esi:             { enabled: true }
-    translator:      { fallback: '%locale%' }
+    translator:      { fallback: '%locale%', paths: ["%kernel.project_dir%/translations"] }
```
yceruto: Single quotes

app/config/config.yml
```diff
@@ -64,6 +64,9 @@ twig:
     form_themes:
         - 'form/layout.html.twig'
         - 'form/fields.html.twig'
+    paths:
+        "%kernel.project_dir%/templates/TwigBundle": Twig
+        "%kernel.project_dir%/templates": ~
```
yceruto: Same here  ;)

mstrom: fixed

app/config/config.yml
mstrom: fixed

javiereguiluz:
@mstrom thanks A LOT for this contribution. Sadly, I must close for now as "won't fix" for the reasons explained by @yceruto. This is the dir structure used by Flex/Symfony 4 ... but we are not using any of those yet (we can't until November 2017).

A similar, but different, proposal is #602 which moves assets dir. This is because the decision is based on Webpack Encore, which is a library independent from Flex/Sf4 and which we already use in this app.

Thanks for understanding it!

PR #614, opened 2017-07-27
Update the command used to check for security vulnerabilities

### LINKED ISSUE #543: Use a secure endpoint for the security checker command

javiereguiluz:
Third (and last) attempt to fix #543.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,4 +32,4 @@ script:
   # this checks that the XLIFF translations contain no syntax errors
   - ./bin/console lint:xliff app/Resources
   # this checks that the application doesn't use dependencies with known security vulnerabilities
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Closing as "it doesn't work". I've notified internally about this error and I'll fix it when I get an answer explaining how to solve it. Thanks!

stof:
@javiereguiluz could you retry this now that Travis updated their infra ? I think the issue might be related to the usage of a Let's Encrypt certificate on https://security.sensiolabs.org/. The old Precise infra was probably missing the certificate.

PR #619, opened 2017-07-29
Translate a few new search translation messages to Ukrainian

bocharsky-bw:


javiereguiluz:
Thanks @bocharsky-bw.

PR #623, opened 2017-08-30
Don't use getenv(), use $_SERVER instead

yceruto:
See https://github.com/symfony/recipes/pull/156

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -17,11 +17,11 @@ use Symfony\Component\HttpFoundation\Request;
 require __DIR__.'/../vendor/autoload.php';
 
 // The check is to ensure we don't use .env in production
-if (!getenv('APP_ENV')) {
+if (!isset($_SERVER['APP_ENV'])) {
     (new Dotenv())->load(__DIR__.'/../.env');
 }
 
-if (getenv('APP_DEBUG')) {
+if ($_SERVER['APP_DEBUG'] ?? true) {
     // WARNING: You should setup permissions the proper way!
     // REMOVE the following PHP line and read
     // https://symfony.com/doc/current/book/installation.html#checking-symfony-application-configuration-and-setup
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Thanks @yceruto.

public/index.php
```diff
@@ -17,11 +17,11 @@
 require __DIR__.'/../vendor/autoload.php';
 
 // The check is to ensure we don't use .env in production
-if (!getenv('APP_ENV')) {
+if (!isset($_SERVER['APP_ENV'])) {
     (new Dotenv())->load(__DIR__.'/../.env');
 }
 
-if (getenv('APP_DEBUG')) {
+if ($_SERVER['APP_DEBUG'] ?? true) {
```
stof: this looks broken, as setting it to `'0'` will still be a truthy string.

public/index.php
```diff
@@ -32,7 +32,7 @@
 
 // Request::setTrustedProxies(['0.0.0.0/0'], Request::HEADER_FORWARDED);
 
-$kernel = new Kernel(getenv('APP_ENV'), getenv('APP_DEBUG'));
+$kernel = new Kernel($_SERVER['APP_ENV'] ?? 'dev', $_SERVER['APP_DEBUG'] ?? true);
```
stof: this is weird too. The second argument should be a boolean. The string `'0'` (as supported by the console) won't be false.

public/index.php
yceruto: and `'0' == false` is true

public/index.php
yceruto: and `(bool) '0'` is `false` 

https://github.com/symfony/symfony/blob/edb8c33f35b981b72adbe23798753703edfe3819/src/Symfony/Component/HttpKernel/Kernel.php#L83

PR #624, opened 2017-08-30
Fix some minor configuration issues

yceruto:
* Use `public` instead of `web` as assets output path
* Exclude public/build from CS Fixer
* Fix comment indent


javiereguiluz:
@yceruto I was sure about having fixed all this (and I thin kit was you who reported them in the first place) ... but it's obvious that I didn't. Thanks for fixing all this!

PR #625, opened 2017-08-30
Update php version requirement in readme

ScullWM:
Hi

Seems that symfony-demo in no longer available for php 5.5.9.
Composer.json define the minimum requirement to 7.1.3.

Readme update

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ to develop Symfony applications following the recommended best practices.
 Requirements
 ------------
 
-  * PHP 5.5.9 or higher;
+  * PHP 7.1.3 or higher;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks for this! I totally missed it. However, I prefer to merge this when we tag the new version, which will require Flex. At the moment, when using the Symfony Installer to install this demo app, you get the non-Flex app. So, the PHP 7.1 requirement is only when cloning the project and using the `master` branch. Anyway, thanks for this because we'll need it soon.

javiereguiluz:
Thank you @ScullWM.

PR #626, opened 2017-08-30
Default to non-debug in index.php and console

yceruto:
See https://github.com/symfony/recipes/pull/163


javiereguiluz:
Thank you Yonel.

PR #627, opened 2017-08-30
Remove forced configuration for tests environment

yceruto:
We can configure the test database file in `phpunit.xml.dist` and keep it simple.


phpunit.xml.dist
```diff
@@ -18,6 +18,7 @@
                Environment variable not found: "APP_SECRET".
         -->
```
yceruto: This TODO comment feel like a work around when defining environment variables here should be the early way to go. Can be removed or modified?

javiereguiluz: Yes! Let's do that in a separate PR. Thanks

PR #628, opened 2017-08-30
Workaround to use DATABASE_URL with %kernel.project_dir% parameter

yceruto:
Setting `DATABASE_URL=sqlite:///var/data/blog.sqlite` as default, causes this known error through `index.php`:
```
An exception occurred in driver: SQLSTATE[HY000] [14] unable to open database file
```
So the right relative path in this case should be `sqlite:///../var/data/blog.sqlite` BUT then it wouldn't work for `bin/console` entries. Then, relative paths shouldn't be the right way but absolute paths.

The right thing is that soon (in 3.4 & 4.0) we can use environment variables with configuration parameters https://github.com/symfony/symfony/pull/23901, so the final configuration would be the one proposed. For now a workaround to make both scenarios work `index.php` and `bin/console`.


javiereguiluz:
Thanks for fixing this issue and for the perfect explanation!

PR #631, opened 2017-08-30
Setting tests directory consistent with src directory structure

yceruto:
> By convention, the `tests/AppBundle` directory should replicate the directory of your bundle for unit tests. So, if you're testing a class in the `src/AppBundle/Util/` directory, put the test in the `tests/AppBundle/Util/` directory. (http://symfony.com/doc/current/testing.html)

The future version of it for 3.4/4.0 would be:
> By convention, the `tests/` directory should replicate your `src` directory for unit tests. So, if you're testing a class in the `src/Util/` directory, put the test in the `tests/Util/` directory.

Nothing to update in `composer.json` as PSR-4 autoload-dev section already known about this.
```json
    "autoload-dev": {
        "psr-4": {
            "App\\Tests\\": "tests/"
        }
    },
```


javiereguiluz:
I totally missed this! Thanks for fixing it.

PR #632, opened 2017-08-31
Fix jQuery instantSearch plugin and search result

yceruto:
search

**instantSearch**
 * Refactored to some known JavaScript pattern (implemented Prototype Design Pattern)
 * Allow override options through input attributes `data-*` and added support to text translation ("No results found")
 * Add article template prototype and render method (improved visibility)
 * Sent configured limit of items to search controller (this allows to configure more than 10 items as results from plugin)
 * Simplify properties and methods (improved visibility)

**Search result**
 * Removed old effect of hide/show results on focus out/in of the input
 * Show publishedDate and author for each article found
 * Sync style with blog index page and some margin adjustments

**Minor unrelated changes**
 * Fix @return type-hints for `PostRepository::findBySearchQuery()` method (better autocomplete)
 * Show logout link as last link always


javiereguiluz:
What can I say? This is fantastic and amazing. I tested it and I love all the changes. Thanks!

PR #634, opened 2017-08-31
Mention to create .env file before running this application

bocharsky-bw:
After `git pull` and `composer install` I was not able to run this application due to the lack of `.env` file.

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -1,4 +1,4 @@
-# This file is a "template" of which env vars needs to be defined in your configuration or in an .env file
+# This file is a "template" of which env vars needs to be defined in your configuration or in a .env file
 # Set variables here that may be different on each deployment target of the app, e.g. development, staging, production.
 # https://symfony.com/doc/current/best_practices/configuration.html#infrastructure-related-configuration
 
--- a/README.md
+++ b/README.md
@@ -48,6 +48,13 @@ most recent version executing the `symfony self-update` command.
 Usage
 -----
 
+First of all, create a `.env` file from a `.env.dist` template which contains default
+values of infrastructure-related configuration of this application:
+
+```bash
+cp .env.dist .env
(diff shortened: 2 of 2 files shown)


javiereguiluz:
@bocharsky-bw thanks for your report ... but I'm afraid we can't accept this PR. The promise we made since day one was that the Symfony Demo app could be run just after installing it without the need to execute any command, add/modify any config or do anything at all. So we must consider this as a bug.

Can anybody think of a better solution that committing the `.env` file to the repo? Thanks!

bocharsky-bw:
I think the best way is to load `.env.dist` file by default in `index.php` if no `.env` exists. Or, we can consider adding a Composer script to actual copy `.env.dist` -> `.env` on composer install/update if not exists

yceruto:
I think it's an issue that could be fixed by symfony/flex in `EnvConfigurator`:

https://github.com/symfony/flex/blob/17e56322f33990daa08cf5761c80188b3a2b41b1/src/Configurator/EnvConfigurator.php#L21-L48

I mean, in a fresh install it's copied from `.env.dist` automatically BUT when we're cloning a project that already has `.env.dist` file updated (no changes regarding recipe) then `.env` file is not created. So in the next `composer require <package-with-recipe>` if they are changes in `.env.dist` then `.env` file is created.

yceruto:
and same here if the original lock hash and composer lock hash doesn't has been changed:

https://github.com/symfony/flex/blob/17e56322f33990daa08cf5761c80188b3a2b41b1/src/Flex.php#L140-L147

fabpot:
Good analysis @yceruto. Fixed in Flex 1.0.17. So, when demo is upgraded to that version, it will fix this issue.

javiereguiluz:
Closing in favor of #638.

PR #636, opened 2017-08-31
Change tests namespaces: 'App\Tests' -> 'Tests\App'

bocharsky-bw:
I'm pretty sure this way is more intuitive, and, actually, this way is used in Symfony Best Practices: http://symfony.com/doc/current/best_practices/tests.html

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "App\\Tests\\": "tests/"
+            "Tests\\App\\": "tests/"
         }
     },
     "scripts": {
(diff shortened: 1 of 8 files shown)


javiereguiluz:
I'm afraid using this namespace is the new best-practice of Symfony. The skeleton used to create apps defines it: https://github.com/symfony/skeleton/blob/master/composer.json#L28.

There is an open discussion about this: https://github.com/symfony/skeleton/pull/15 and most people seem to be against this change. I'm going to close this as "won't fix" for now ... but if Symfony ultimately changes this, we'll reopen and merge your PR.

Thanks!

PR #637, opened 2017-09-01
Changing name convention of "logs" directory to "log"

yceruto:
https://github.com/symfony/recipes/commit/2117e0eb32b0143bd396c8292d066b841192c44f


yceruto:
Thanks @ogizanagi by the warning, I had not noticed that :+1:, 
Ready.

PR #639, opened 2017-09-04
Remove useless PhpDoc comments and minor improvements

yceruto:
* Remove useless PhpDocs comments.
* Improving Twig `locales()` function.


javiereguiluz:
@20uf I think I understand your downvote ... but don't be afraid because this won't break PHPStorm autocompletion. Most of these annotations are indeed useless because PHPStorm already can infer their types (through the constructor or other assignments). There's at least one `string` annotation that we'll loose, but `string` typehint is not very useful anyway.

Moreover, we now use PHP 7.1, so we can typehint everything and remove all PHPdoc.

javiereguiluz:
Thank you @yceruto.

20uf:
The typehint does not completely replace the annotations and the strong typing is not fully respected, for example the type of return is not specified.

PR #640, opened 2017-09-04
Remove container dependency in PostFixtures

yceruto:
and remove useless public service alias.


javiereguiluz:
I like this simplification. Thanks!

PR #641, opened 2017-09-04
Using the new PHPUnit namespace

yceruto:
Moving on to the most recent updates.


javiereguiluz:
A little nice change to make our code use modern features only. Thanks!

PR #643, opened 2017-09-05
Update env variables for the heroku deployment

stof:
I have not updated the SYMFONY_LOG one, as the monolog config does not use an env variable anymore for the log output, so it requires more work.


javiereguiluz:
Thank you Christophe.

PR #644, opened 2017-09-06
Using the new Twig namespace

yceruto:
Moving on to the latest changes...


javiereguiluz:
Looks great! @yceruto thanks for this.

The only thing I don't like is this change:

```diff
-    public function showSourceCode(\Twig_Environment $twig, $template)
+    public function showSourceCode(Environment $twig, $template)
```

`Environment` looks confusing and out of place, because it has nothing to do with `Twig`, which is what we want here.

PR #645, opened 2017-09-06
Add scalar typehints/return types

yceruto:
Fixes https://github.com/symfony/symfony-demo/pull/639#issuecomment-327434749

### DIFF
--- a/src/Entity/Post.php
+++ b/src/Entity/Post.php
@@ -12,6 +12,7 @@
 namespace App\Entity;
 
 use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
 use Symfony\Component\Validator\Constraints as Assert;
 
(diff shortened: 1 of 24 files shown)


javiereguiluz:
Can we have a rebase here, please? Thanks a lot!

yceruto:
Done.

src/Entity/Post.php
```diff
@@ -126,121 +127,103 @@ public function __construct()
...
     }
 
-    public function getId()
+    public function getId(): int
     {
         return $this->id;
     }
 
-    public function getTitle()
+    public function getTitle(): ?string
```
javiereguiluz: Not for this PR, but we should make the title (and slug) mandatory so this return value can never be null, don't you think?

src/Form/DataTransformer/TagArrayToStringTransformer.php
```diff
@@ -36,7 +36,7 @@ public function __construct(ObjectManager $manager)
     /**
      * {@inheritdoc}
      */
-    public function transform($array)
+    public function transform($array): string
```
javiereguiluz: `array` typehint is missing. What if we rename the argument too from `$array` to `$tags` ?

src/Entity/Post.php
yceruto: I thought the same at beginning, but they are already mandatory in DB & FormValidation. so when I tested the real app (`/es/admin/post/new`):

> Type error: Return value of App\Entity\Post::getTitle() must be of the type string, null returned

The reason is on form initialization, it needs `mapDataToForms()` and to read each defined form field from underlying object (default values), so `PropertyAccessor->readProperty(array(object(Post)), 'title')` calls to `Post->getTitle()` and boom! `null` value.

The solution could be set a fake or default title/slug for new `Post`s, now that `Slugger::slugify()` is static, it can be easier. WDYT? then we can change it to strict `string`.

src/Form/DataTransformer/TagArrayToStringTransformer.php
yceruto: > `array` typehint is missing.

Yep, but I can't do that, because declaration must be compatible with `DataTransformerInterface->transform(value)`.

> What if we rename the argument too from $array to $tags ?

:+1: Done.

src/Entity/Post.php
javiereguiluz: I'm not sure it's worth it. Now that you explained that, I remember that I faced this issue too a while ago. I needed to put `?string` in an entity because of the Form behavior that you mentioned. So let's keep it. Thanks.

javiereguiluz:
It's nice to have PHP 7.1 type hints and return types everywhere. Thanks a lot for contributing this feature!

yceruto:
Glad to help :tada:!!

PR #647, opened 2017-09-07
Remove useless constants in commands

yceruto:
After https://github.com/symfony/symfony-demo/commit/5180bdcce3bac08e59ba9b5d34fbd7c2b007f7e3


javiereguiluz:
Good catch, thanks @yceruto.

Issue #649, opened 2017-09-15
Move BrowserKit to dev requirements?


garak:
I would propose a PR myself, but I want to be sure: is there any reason to keep "symfony/browser-kit" in "require" section of composer.json (instead of "require-dev")?
It looks like this application is never using anything in the `Symfony\Component\BrowserKit` namespace, so the only use I can see is for tests.

yceruto:
:+1:

javiereguiluz:
Nice catch. This definitely looks like a mistake. Thanks!

javiereguiluz:
Fixed by #650

PR #650, opened 2017-09-19
Move browser-kit and css-selector to require-dev

garak:
See #649


javiereguiluz:
Thanks @garak.

stof:
you need to run `composer update --lock` to avoid warnings about outdated lock file

PR #653, opened 2017-09-20
Add validation of the composer files on Travis

stof:
This performs validation of the composer metadata to ensure they are valid:

- composer.json does not contain any error
- composer.json does not trigger any warning (thanks to the strict mode)
- composer.lock is uptodate

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,3 +29,5 @@ script:
   - ./bin/console lint:xliff translations
   # this checks that the application doesn't use dependencies with known security vulnerabilities
   - ./bin/console security:check --end-point=http://security.sensiolabs.org/check_lock
+  # This checks that the composer.json and composer.lock files are valid
+  - composer validate --strict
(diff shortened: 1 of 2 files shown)


stof:
this will avoid mistakes like what happened in #650 where the lock file was not updated with `composer update --lock`) (moving deps to dev deps or the opposite does not require changing the version of deps, but it impacts the lock file and composer was displaying a warning about outdated lock file when running `composer install`)

stof:
Note that I added this on Travis only.
Running this on both Travis and Appveyor is useless (the validator does not execute any of our code, and the JSON content will be the same on Windows and Linux).

And Travis collapses the installation instruction so it is easier to reach the test output

javiereguiluz:
@stof thanks for adding this useful check and for explaining it perfectly!

stof:
Note that there is a few variation of the command to add on CI depending on the project:

- if the project is not meant to be published on Packagist (or in a private registry), you can add `--no-check-publish` to avoid requiring a name and description if you don't have them. Libraries registered on Packagist should never use this flag.
- if the repo does not commit the lock file (which should only happen for libraries), you may want to add `--no-check-lock` to skip useless validation (if the composer.lock was generated by the CI job itself when running `composer update`, the lock file will never be outdated, and if it is due to a composer bug, you cannot fix the lock file anyway)

Note that Packagist validates the composer.json before accepting updates, so an invalid one might prevent the new commit to become visible to users.

PR #655, opened 2017-09-20
Cache the phpunit install between builds on CI

stof:
This leverages the CI caches to avoid redownloading PHPUnit on each build with the bridge.
This implies changing the folder in which it gets downloaded to allow caching without side-effects on the composer install (by default, the install happens in `vendor/symfony/phpunit-bridge/bin/.phpunit` but composer cleans existing folders when installing packages)

This also separates the PHPUnit install from the run instead of relying on the lazy-install, to make the output more understandable (especially on Travis where the install part will get collapsed).
Seeing a composer output when running tests can be confusing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,11 @@ sudo: false
 cache:
   directories:
     - $HOME/.composer/cache/files
+    - $HOME/symfony-bridge/.phpunit
+
+env:
+  global:
+    - SYMFONY_PHPUNIT_DIR="$HOME/symfony-bridge/.phpunit"
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Our test builds were reasonably fast ... but performance gains are always welcomed! Thank you.

stof:
thus, it gives an example for other projects using the Symfony bridge 😄

stof:
Note that the caching was not that bad already, as the bridge would still benefit from the composer cache when calling composer.

The impact of such caching was bigger on the symfony/symfony repo, where the composer cache is not persisted between builds (the magic we do to expose the local version of sub-components when running composer would hurt us if the cache was shared with next builds)

PR #658, opened 2017-09-22
Changed "null" by "~" in yaml files

yceruto:
According to:

* https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/config/packages/test/framework.yaml#L2

* https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/config/packages/framework.yaml#L6

* https://github.com/symfony/recipes/blob/master/symfony/routing/3.3/config/packages/routing.yaml#L3

### DIFF
--- a/config/packages/framework.yaml
+++ b/config/packages/framework.yaml
@@ -36,6 +36,6 @@ framework:
         handler_id: session.handler.native_file
         save_path: '%kernel.project_dir%/var/sessions/%kernel.environment%'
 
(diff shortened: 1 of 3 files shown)


stof:
Well, `null` and `~` are exactly the same (both are YAML representations of `null`

yceruto:
Yeah, it's more about consistency with the default recipes.

Issue #620, opened 2017-07-31
Update Entities to use Doctrine 2.6 types


HeahDude:
Thanks to https://github.com/symfony/symfony/pull/22689 `Comment` and `Post` entities should now use the new `date time_immutable` type for their `$publishedAt` property, so getters/setters should accept `DateTimeInterface`.

If you agree, this should be an easy pick.

javiereguiluz:
It looks like a good idea! But let's not submit any pull request until #617 is merged. Thanks!

voronkovich:
Also we should replace the `json_array` type by `json` because the `json_array` is deprecated since Doctrine 2.6.

javiereguiluz:
This proposal is now unlocked and ready for someone who wants to contribute to this project with this easy pick. Thanks!

gabel:
I guess `datetime_immutable` can not be applied as the `\App\Controller\Admin\BlogController::newAction` uses handle($request) which transforms the input into a `\DateTime` object and that would break the ORM mapping. Looks like this is related to https://github.com/symfony/symfony/pull/19889.

If you need more details I can provide a PR with example implementation.

gabel:
Talk is cheep.. PR is here https://github.com/symfony/symfony-demo/pull/659 but test failing due to the mentioned issue.

javiereguiluz:
Let's close this as "fixed". The JSON types are now used and the immutable date can't be used because of Symfony Forms. The related issue (https://github.com/symfony/symfony/issues/9508) has been opened for more than 4 years, so this is more difficult to implement than it looks. Thanks!

PR #659, opened 2017-09-26
Update Entities to use Doctrine 2.6 types

gabel:
Tests will fail due to DateTime conversion of `handleRequest()` (src/Controller/Admin/BlogController.php:84)

### DIFF
--- a/src/Entity/Post.php
+++ b/src/Entity/Post.php
@@ -83,9 +83,9 @@ class Post
     private $content;
 
     /**
-     * @var \DateTime
+     * @var \DateTimeImmutable
      *
-     * @ORM\Column(type="datetime")
+     * @ORM\Column(type="datetime_immutable")
      * @Assert\DateTime
      */
     private $publishedAt;
@@ -122,7 +122,7 @@ class Post
 
     public function __construct()
     {
-        $this->publishedAt = new \DateTime();
+        $this->publishedAt = new \DateTimeImmutable();
         $this->comments = new ArrayCollection();
         $this->tags = new ArrayCollection();
     }
@@ -162,12 +162,12 @@ class Post
         $this->content = $content;
     }
 
-    public function getPublishedAt(): \DateTime
+    public function getPublishedAt(): \DateTimeInterface
     {
         return $this->publishedAt;
     }
 
-    public function setPublishedAt(\DateTime $publishedAt): void
+    public function setPublishedAt(\DateTimeInterface $publishedAt): void
     {
         $this->publishedAt = $publishedAt;
     }
--- a/src/DataFixtures/ORM/PostFixtures.php
+++ b/src/DataFixtures/ORM/PostFixtures.php
@@ -48,7 +48,7 @@ class PostFixtures extends AbstractFixture implements DependentFixtureInterface
             $post->setSummary($this->getRandomPostSummary());
             $post->setSlug(Slugger::slugify($post->getTitle()));
             $post->setContent($this->getPostContent());
-            $post->setPublishedAt(new \DateTime('now - '.$i.'days'));
+            $post->setPublishedAt(new \DateTimeImmutable('now - '.$i.'days'));
 
             // Ensure that the first post is written by Jane Doe to simplify tests
             // "References" are the way to share objects between fixtures defined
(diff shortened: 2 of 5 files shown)


javiereguiluz:
What can we do here to move forward?

* The application is broken after this change. Can we easily fix it?
* What's the practical advantage of using immutable date types in this application?

Thanks!

jkufner:
> What's the practical advantage of using immutable date types in this application?

When you pass the same datetime object to two components and one of them adds a day without cloning, the immutable type won't affect the other component, but the mutable will change in both places.

javiereguiluz:
@jkufner yes, that's the theory part. But now, focusing exclusively on practice, which bug does this change fix or what real improvement does it bring to this specific application? Thanks!

jkufner:
@javiereguiluz When there is no bug, it does not matter which of the types you use, but you can never be sure there is no bug. Immutable types make code more robust and bugs easier to locate.

javiereguiluz:
@jkufner yes, in general that's true. But in this specific application, which is the benefit of this change? I'm not saying there's none, I just don't see it and that's why I'm asking.

jkufner:
@javiereguiluz It is a good practice and a precaution for the future. Something that an example app should highlight. The code itself does the same.

gabel:
I assume we cannot move forward here until the underlying libs have fixed the immutable usage. If @javiereguiluz sees no benefit for now this PR and #620 may be closed for now.

javiereguiluz:
Let's ask other regular contributors of this repository to see if they agree with this change or prefer to close it (and add it back later when things improve): @yceruto @voronkovich @bocharsky-bw. Thanks!

yceruto:
> Expected argument of type "DateTimeImmutable", "DateTime" given

> Could not convert PHP value of type 'DateTime' to type 'datetime_immutable'. Expected one of the following types: null, DateTimeImmutable

Sadly the Form Component is not ready completely to return a `DateTimeImmutable` instance upon reverse transformation:

https://github.com/symfony/symfony/blob/0082981e4135cf6c04545156db8fcda2c9459f9a/src/Symfony/Component/Form/Extension/Core/DataTransformer/DateTimeToStringTransformer.php#L123

So we might have to do something to fix it, probably into `DateTimePickerType` and then I'm not sure if these changes are worth it.

src/Entity/Post.php
```diff
@@ -162,12 +162,12 @@ public function setContent(string $content): void
...
     }
 
-    public function getPublishedAt(): \DateTime
+    public function getPublishedAt(): \DateTimeInterface
     {
         return $this->publishedAt;
     }
 
-    public function setPublishedAt(\DateTime $publishedAt): void
+    public function setPublishedAt(\DateTimeInterface $publishedAt): void
```
yceruto: Should be `\DateTimeImmutable` to avoid `\DateTime` instances, same for `Comment::setPublishedAt()`.

voronkovich:
I agree with @jkufner that the usage of the `DataTimeImmutable` makes an application more robust. So, I think we should wait until the `Form` component has been fixed.

src/Entity/Post.php
yceruto: Otherwise, we can convert here to immutable date time if so:
```php
if ($publishedAt instanceof \DateTime) {
    $publishedAt = \DateTimeImmutable::createFromMutable($publishedAt);
}
```

src/Entity/Post.php
```diff
@@ -162,12 +162,12 @@ public function setContent(string $content): void
         $this->content = $content;
     }
 
-    public function getPublishedAt(): \DateTime
+    public function getPublishedAt(): \DateTimeInterface
```
yceruto: Should be `\DateTimeImmutable`.

javiereguiluz:
If the Form component doesn't support immutable dates, that's a huge blocker for me. Is there any open pull request to improve the Form component about this? Will it be merged soon? If not, I'm afraid we should close this as "can't fix" until the Form component improves.

src/Entity/Post.php
voronkovich: @yceruto, What about to create a data transformer for those properties?

yceruto: Yes, it could be so, but still I'm not sure if it's worth it.

gabel:
As already mentioned in #620 - I guess that somebody has to pick up this unfinished PR https://github.com/symfony/symfony/pull/19889 before it makes sense to continue here.

voronkovich:
@javiereguiluz, @gabel we could use a `datetime_immutable` type only for the `Comment::publishedAt` property.

javiereguiluz:
Let's close this until Symfony Forms fully support immutable dates (which sadly could take a lot of time!) Meanwhile, I've extracted in #669 the part of this pull request unrelated to immutable dates. Thank you all for the discussion.

PR #663, opened 2017-10-01
Fixing add user password validation

chr-hertel:
Fixing password validation in `AddUserCommand`.
added type hint to `$io`-var enabling IDE completion to prevent things like that in future changes


javiereguiluz:
Thanks @chr-hertel.

PR #673, opened 2017-10-14
Upgrade from 3.3 to 3.4 & 4.0

yceruto:
Remaining deprecation notices:

SwiftmailerBundle
---
- [x] <del>User Deprecated: The **"swiftmailer.mailer.default.plugin.messagelogger"** service is private, checking for its existence is deprecated since Symfony 3.2 and will fail in 4.0.</del>

- [x] <del>User Deprecated: The **"swiftmailer.mailer.default.plugin.messagelogger"** service is private, getting it from the container is deprecated since Symfony 3.2 and will fail in 4.0. You should either make the service public, or stop using the container directly and use dependency injection instead. </del>

- [x] <del>User Deprecated: The **"mailer"** service is private, checking for its existence is deprecated since Symfony 3.2 and will fail in 4.0.</del>

- [x] <del>The **"swiftmailer.mailer.default.transport.real"** service is private, getting it from the container is deprecated since Symfony 3.2 and will fail in 4.0. You should either make the service public, or stop using the container directly and use dependency injection instead.</del>

WhiteOctoberPagerfantaBundle
---
- [x] <del>User Deprecated: The **"white_october_pagerfanta.view_factory"** service is private, getting it from the container is deprecated since Symfony 3.2 and will fail in 4.0. You should either make the service public, or stop using the container directly and use dependency injection instead.</del>

Symfony
---
- [x] <del>The **Symfony\Component\HttpFoundation\Session\Storage\Handler\NativeSessionHandler** class is deprecated since version 3.4 and will be removed in 4.0. Use the **\SessionHandler** class instead.</del>

- [x] <del>User Deprecated: The **Symfony\Component\HttpFoundation\Session\Storage\Proxy\SessionHandlerProxy** class is deprecated since version 3.4 and will be removed in 4.0. Use your session handler implementation directly.</del>

- [x] <del>User Deprecated: The **Symfony\Component\HttpFoundation\Session\Storage\Proxy\AbstractProxy** class is deprecated since version 3.4 and will be removed in 4.0. Use your session handler implementation directly.</del>

- [x] <del>User Deprecated: The **"Doctrine\DBAL\Configuration"** class is considered internal When adding a new configuration option just write a getter/setter pair and add the option to the _attributes array with a proper default value. It may change without further notice. You should not use it from **"Doctrine\ORM\Configuration"**.</del>

- [x] <del>Setting **logout_on_user_change** to false is deprecated as of 3.4 and will always be true in 4.0. Set **logout_on_user_change** to true in your firewall configuration. (isn't a blocker for 4.0)</del>

- [x] <del>Relying on service auto-registration for type **"App\Entity\Post"** is deprecated since version 3.4 and won't be supported in 4.0. Create a service named **"App\Entity\Post"** instead. (isn't a blocker for 4.0)</del>

- [x] <del>User Deprecated: The **"Symfony\Bridge\Doctrine\ManagerRegistry::setContainer()"** method is deprecated since version 3.4 and will be removed in 4.0. Inject a PSR-11 container using the constructor instead.</del>


yceruto:
> Relying on service auto-registration for type "App\Entity\Post" is deprecated since version 3.4 and won't be supported in 4.0. Create a service named "App\Entity\Post" instead.

Even though it doesn't a blocker to migrate to 4.0, look so weird auto-registrate classes that already are excluded by configuration.

javiereguiluz:
The SwiftmailerBundle deprecations have been reported in https://github.com/symfony/swiftmailer-bundle/issues/191.

javiereguiluz:
The WhiteOctoberPagerfantaBundle deprecation is being fixed in https://github.com/whiteoctober/WhiteOctoberPagerfantaBundle/pull/173

javiereguiluz:
The strange deprecation related to `App\Entity\Post` has been reported here: https://github.com/symfony/symfony/issues/24569

composer.json
```diff
@@ -6,39 +6,40 @@
...
-        "symfony/orm-pack": "^1.0",
-        "symfony/polyfill-apcu": "^1.4",
-        "symfony/security-bundle": "^3.3",
-        "symfony/swiftmailer-bundle": "^3.0",
-        "symfony/translation": "^3.3",
-        "symfony/validator": "^3.3",
-        "symfony/yaml": "^3.3",
-        "twig/extensions": "^1.5",
+        "doctrine/doctrine-fixtures-bundle": "2.4.0",
+        "erusev/parsedown": "1.6.3",
```
stof: Why exact requirements everywhere ? This is clearly no a best practice (forbidding to update patch releases with a simple `composer update` makes maintenance harder)

yceruto: NVM, It's a temporal change to avoid update these dependencies right now, later these constraints will be reverted.

javiereguiluz:
In this PR we can also move `src/Resources/TwigBundle/views/*` to `templates/bundles/TwigBundle/`

javiereguiluz:
Makefile was removed from Flex (see https://github.com/symfony/flex/pull/184) so we need to remove it here too. Thanks!

composer.json
stof: But 3/4 of the dependencies are not in the composer.json (as they are transitive deps), and these are still updated to dev versions. So I don't think these changes make sense (and they make the review harder, as well as the future maintenance)

yceruto: > these are still updated to dev versions.

That's why I added `"symfony/lts": "dev-master"` to avoid transitive deps update to 4-dev, all of them are forced to 3-dev state.

> So I don't think these changes make sense (and they make the review harder, as well as the future maintenance)

I guess you're refering to changes like this `"erusev/parsedown": "1.6.3"`? if so, there are temporal changes to avoid update these deps right now. This `"minimum-stability": "dev"` option forces all deps to dev unless there is a constrains that prevents it (that's what I did with Symfony deps ``"symfony/lts": "dev-master"`). 

I did these changes to rely on Symfony 3/4 dev deps only. I think isn't a good idea rely on all dev deps during the migration to 4.0, since we would be testing against dev deps that later they might not work.

When the upgrade finishes, we can go back safely to normal constraints `1.6.3` > `^1.6`, remove `"minimum-stability": "dev"` option and fix the Symfony release deps.

If it isn't reasonable, please, what would be the alternative during this upgrade?

yceruto:
The session deprecations are gone after update vendors.

src/Command/AddUserCommand.php
```diff
@@ -44,6 +44,10 @@
  */
 class AddUserCommand extends Command
 {
+    // to make your command lazily loaded, either define its $defaultName static property,
```
stof: `either` does not make sense if you don't have a `or` later

src/Command/AddUserCommand.php
```diff
@@ -68,8 +72,7 @@ public function __construct(EntityManagerInterface $em, UserPasswordEncoderInter
     protected function configure()
     {
         $this
-            // a good practice is to use the 'app:' prefix to group all your custom application commands
-            ->setName('app:add-user')
+            // you don't need to call ->setName() for configuring the command when it is lazy.
```
stof: this comment is useless (people won't see setName at all)

src/Command/AddUserCommand.php
yceruto: Indeed, thanks!

src/Command/AddUserCommand.php
yceruto: removed.

src/Command/AddUserCommand.php
```diff
@@ -44,6 +44,10 @@
  */
 class AddUserCommand extends Command
 {
+    // to make your command lazily loaded, configures the $defaultName static property,
```
javiereguiluz: Typo: `configures the $defaultName...` -> `configure the $defaultName...`

Suggestion: `only when the app:add-user command is actually called` -> `only when the command is actually called`

yceruto: Thanks Javier, done.

yceruto:
* After https://github.com/symfony/swiftmailer-bundle/pull/200 "mailer" deprecation is gone. 
* Also I updated the `composer.json` to move all non-symfony deps to the latest stable version. (removed static constraints)

yceruto:
* After https://github.com/symfony/swiftmailer-bundle/pull/203 remaining deprecations from `swiftmailer-bundle` are gone.

yceruto:
[symfony/lts has created the tag v3](https://github.com/symfony/lts/releases/tag/v3) so, updating its constraint here.

yceruto:
Doctrine warnings are gone after update `doctrine/doctrine-bundle` to 1.6.

javiereguiluz:
@yceruto I'm testing the latest changes of this PR. Your work is great. Thank you!  I only have a question: when running tests I just see 1 deprecation, but when running the application in the browser I see many more. Do you know why?

deprecations

stof:
@javiereguiluz I think some of them are coming from the DebugClassLoader, and so depend on whether the class is loaded or no.

stof:
and the profiler also redisplays the deprecations from the compile-time, even though the cached container was used. The PHPUnit Bridge does not. It only reports deprecations actually happening during the tests.

javiereguiluz:
Do you consider this behavior something to be improved? I'd like to run tests and get ALL deprecations, instead of having to browse the application manually. Thanks!

stof:
@javiereguiluz reporting compile-time deprecations in the bridge would be quite hard. There is nothing triggering a `E_USER_DEPRECATED` here for them. The data collector is reading the cached file in which they were saved.
The PHPUnit bridge cannot do it, as it cannot be sure that the test being executed actually used the container at all (most unit tests don't use it, and we don't want them to report deprecations from a previous functional test, as you would then have to mark the whole testsuite as legacy).
If you want to have compile-time deprecations in your testsuite, write a test forcing a cache warmup so that the cache is not used.

yceruto:
@dmaicher we might need a new release to avoid use `dev-master` constraint for Symfony 4 ;) thanks!

yceruto:
We can't move forward until `white-october/pagerfanta-bundle` adds `FrameworkBundle` 4.0 compat to `composer.json`.

dmaicher:
@yceruto I created a new release that allows Symfony 4

javiereguiluz:
I've tested the latest changes and I still see these four deprecations. How can we solve them? Thanks!

* The "Doctrine\DBAL\Configuration" class is considered internal When adding a new configuration option just write a getter/setter pair and add the option to the _attributes array with a proper default value. It may change without further notice. You should not use it from "Doctrine\ORM\Configuration".
* Setting logout_on_user_change to false is deprecated as of 3.4 and will always be true in 4.0. Set logout_on_user_change to true in your firewall configuration.
* The "Doctrine\DBAL\Configuration" class is considered internal When adding a new configuration option just write a getter/setter pair and add the option to the _attributes array with a proper default value. It may change without further notice. You should not use it from "Doctrine\ORM\Configuration".
* Relying on service auto-registration for type "App\Entity\Post" is deprecated since version 3.4 and won't be supported in 4.0. Create a service named "App\Entity\Post" instead.

javiereguiluz:
In my local fork I've removed all deprecations as follows:

1) Change `"minimum-stability": "beta"` to `"minimum-stability": "dev"` so you can get the latest Symfony code and get rid of:

    * The "Doctrine\DBAL\Configuration" class is considered internal When adding a new configuration option just write a getter/setter pair and add the option to the _attributes array with a proper default value. It may change without further notice. You should not use it from "Doctrine\ORM\Configuration".
    * The "Doctrine\DBAL\Configuration" class is considered internal When adding a new configuration option just write a getter/setter pair and add the option to the _attributes array with a proper default value. It may change without further notice. You should not use it from "Doctrine\ORM\Configuration".

2) Edit `config/packages/routing.yaml` and add `logout_on_user_change: true` to the `main` firewall to get rid of:

    * Setting logout_on_user_change to false is deprecated as of 3.4 and will always be true in 4.0. Set logout_on_user_change to true in your firewall configuration.

3) Update the `src/Kernel.php` file according to https://github.com/symfony/recipes/pull/221/files to get rid of:

    * Relying on service auto-registration for type "App\Entity\Post" is deprecated since version 3.4 and won't be supported in 4.0. Create a service named "App\Entity\Post" instead.

javiereguiluz:
@yceruto minor comment. We should also update in this PR the following "bundle notation" to use the new FQCN-based notation: https://github.com/symfony/demo/blob/c68bcdd6373a92800d424a55809d3a386aeee00b/templates/base.html.twig#L101

yceruto:
@javiereguiluz perhaps we can ignore 2 & 3, as they'll disappear in 4.0

yceruto:
I don't know why, but after update deps `doctrine/doctrine-migrations-bundle` has been installed transitively, which it's weird :/

ogizanagi:
@yceruto : It's now part of the [orm-pack](https://github.com/symfony/orm-pack/blob/master/composer.json)

yceruto:
Well, it is part of `symfony/orm-pack` package

ogizanagi:
About

> Relying on service auto-registration for type "App\Entity\Post" is deprecated since version 3.4 and won't be supported in 4.0. Create a service named "App\Entity\Post" instead. (isn't a blocker for 4.0)

You should set the [`container.autowiring.strict_mode`](https://github.com/symfony/symfony/pull/24671) to true :)

It's done directly in the Kernel in the [fwb recipe](https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/src/Kernel.php#L39)

yceruto:
@ogizanagi yeah, I know :) but [the parameter is useless in 4.0](https://github.com/symfony/symfony/blob/v4.0.0-BETA2/src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php) and this PR ends there. 

Same for the another one, setting `logout_on_user_change` to `true` removes the warning, but in 4.0 it's `true` by default, so we don't need to do something either.

The only blocker to move to 4.0 is https://github.com/symfony/demo/pull/673#issuecomment-338646414 right now :/

yceruto:
We are already in Symfony 4.0!

Just waiting for a new release of `white-october/pagerfanta-bundle` to see the test green.

yceruto:
Tests are green now, this PR is ready!

javiereguiluz:
@yceruto to make your life easier, I think we should merge this as is ... and keep updating to other betas, RCs and the stable version in separate PRs. That would also help others test these changes more easily. If nobody disagrees with this idea, I'll merge it tomorrow morning. Thanks!

yceruto:
@javiereguiluz totally agree!

javiereguiluz:
Merged! We're in Symfony 4!!! Thanks @yceruto!

config/services.yaml
```diff
@@ -16,37 +16,27 @@ services:
...
-    App\Command\ListUsersCommand:
-        $emailSender: '%app.notifications.email_sender%'
-
     # when the service definition only contains arguments, you can omit the
     # 'arguments' key and define the arguments just below the service class
-    App\Twig\AppExtension:
-        $locales: '%app_locales%'
-
     App\EventSubscriber\CommentNotificationSubscriber:
         $sender: '%app.notifications.email_sender%'
```
TomasVotruba: Maybe rename property and remove this definition too?

PR #674, opened 2017-10-16
update dama/doctrine-test-bundle to ^4.0

dmaicher:
Updates to latest version of `dama/doctrine-test-bundle` which uses the same listener class for PHPUnit 5 and 6+.


javiereguiluz:
Thank you @dmaicher.

PR #676, opened 2017-10-17
Simplified the README instructions

javiereguiluz:
This is needed because we're moving away from the Symfony Installer and we'll use Composer's create-project from now on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@ Symfony Demo Application
 The "Symfony Demo Application" is a reference application created to show how
 to develop Symfony applications following the recommended best practices.
 
-[![Build Status](https://travis-ci.org/symfony/symfony-demo.svg?branch=master)](https://travis-ci.org/symfony/symfony-demo)
-
 Requirements
 ------------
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,76 +4,42 @@ Symfony Demo Application
...
-access the application in your browser at <http://localhost:8000>. You can
-stop the built-in web server by pressing `Ctrl + C` while you're in the
-terminal.
-
-> **NOTE**
->
-> If you want to use a fully-featured web server (like Nginx or Apache) to run
-> Symfony Demo application, configure it to point at the `web/` directory of the project.
-> For more details, see:
-> https://symfony.com/doc/current/cookbook/configuration/web_server_configuration.html
```
stof: Should this note be removed entirely ? I think it might make sense to keep it. People using the symfony-demo to benchmark Symfony should hopefully do it with Apache/Nginx and PHP-FPM rather than using the single-threaded builtin webserver.

README.md
```diff
@@ -4,76 +4,42 @@ Symfony Demo Application
...
-if you haven't already. Then, install the Symfony Demo Application executing
-this command anywhere in your system:
+Execute this command to install the project:
 
 ```bash
-$ symfony demo
-
-# if you're using Windows:
-$ php symfony demo
+$ composer create symfony/symfony-demo
```
stof: I suggest using the actual name (`create-project`) rather than relying on the symfony/console shortcut feature. If someone has a global plugin installed providing a `create` command, this would run the wrong command.

README.md
```diff
@@ -4,76 +4,42 @@ Symfony Demo Application
...
-Testing
--------
+Tests
+-----
 
 Execute this command to run tests:
 
 ```bash
-$ cd symfony_demo/
+$ cd symfony-demo/
```
bocharsky-bw: It was on purpose because Symfony installer creates `symfony_demo` folder instead, i.e. when `$ symfony demo`

javiereguiluz: Yes! But now with "composer create symfony/symfony-demo" it generates "symfony-demo/" so I think we must update this.

bocharsky-bw: If we remove Symfony installer from docs then 👍

javiereguiluz: It's removed :) This is how the new README looks: https://github.com/javiereguiluz/symfony-demo/blob/5b68ac7205f7ef3c19115593170ba132279f48c6/README.md

PR #679, opened 2017-10-17
Branche clem

clemq:


clemq:
Sorry, this was an exercice and I failed !

PR #691, opened 2017-11-03
Remove the project ID, which should be different per project

fabpot:
When using the demo, people should get a new project id instead of a generic one.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -73,7 +73,7 @@
     },
     "extra": {
         "symfony": {
-            "id": "01BP22MQ00QH5WD6NN6HHD3EX4",
+            "id": "",
             "allow-contrib": true
         }
     }


javiereguiluz:
Thanks Fabien.

yceruto:
I'm afraid this change will [break the test all time](https://travis-ci.org/symfony/demo/builds/296847531#L643) (because `composer validate` check) and the lock file is not up to date with the latest changes in `composer.json` at the moment of cloning this project directly from github + `composer install`.

We should revert this one or there is a workaround for that?

javiereguiluz:
I f I execute `composer update --lock` the problem solves ... except for Flex adding the ID back again ... and making `composer validate` fail again when you remove the ID :(

yceruto:
What about to execute `composer update --lock` after `composer install` and before `composer validate` in Travis?

fabpot:
You can set `SYMFONY_SKIP_REGISTRATION` to 1 before running composer update.

yceruto:
This means that we must remember do the same during development to avoid commit a new project ID / lock file outdated (what could be a bit annoying imho).

See https://github.com/symfony/demo/pull/693.

fabpot:
@yceruto Sure, but this is a very special project.

stof:
> What about to execute composer update --lock after composer install and before composer validate in Travis?

this would totally defeat the point of the test (as it would still allow to commit an out-of-sync lock file)

PR #693, opened 2017-11-06
Update lock and skip project ID registration on Travis

yceruto:
Fix https://github.com/symfony/demo/pull/691 with https://github.com/symfony/demo/pull/691#issuecomment-342175350.

This should fix the tests on master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ cache:
 env:
   global:
     - SYMFONY_PHPUNIT_DIR="$HOME/symfony-bridge/.phpunit"
+    - SYMFONY_SKIP_REGISTRATION=1
 
 matrix:
   fast_finish: true
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,7 @@
     "minimum-stability": "beta",
     "extra": {
         "symfony": {
-            "id": "01BY8QJHBGTYRP06WQFNY99EKP",
+            "id": "",
             "allow-contrib": true
         }
     }
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "50f921380572e84d1bdb54e59c3e315e",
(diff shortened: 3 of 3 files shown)


stof:
you should also update the CONTRIBUTING.md with a warning about having to skip registration when working on this project

javiereguiluz:
Just asking: what if we define a Flex ID for this project after all? That'd be wrong in a real project (each user must have their own ID) but as @fabpot said, this is a very special project. Moreover, using the same ID would allow us to get stats about the usage of this project. What do you think?

stof:
@javiereguiluz should each *user* actually have a flex id, or each project ? If it has to be a different id per user, storing it into the composer.json is plain wrong in Flex, as the composer.json is not per user but per project.

javiereguiluz:
@stof yes, the ID is a "user ID", not a "project ID". However, and only for this particular and so special project, we could bend this rule and "hardcode" the ID so every user that clones/downloads this project uses the same ID.

Advantages:

* This PR and all the other mentioned problems disappear.
* We could use this ID to gather (anonymous and very generic) stats about the usage of this project.

stof:
if the id is a user id, the architecture of Flex is broken: it cannot store it in the project-level composer.json file.

stof:
@javiereguiluz  the PR https://github.com/symfony/demo/pull/691 is saying `project id`, so it looks like you and @fabpot are not agreeing on what the id is about.

javiereguiluz:
If I'm wrong (probably I am!!) then this problem solved itself: the ID is per project; this is a project; we can set up a project ID and avoid all issues.

yceruto:
> project ID should be different per project

As far as I understand in https://github.com/symfony/demo/pull/691 "project ID" == "user project ID" I guess we're talking about the same thing :/

fabpot:
The ID is indeed a **project** ID as there is no notion of users in Flex.

If this project is **never** used as a basis for new projects, then having a hardcoded id is fine. What I fear is people using this as a base for their new projects and so inheriting the ID.

javiereguiluz:
Let's close this because the Flex ID is a project ID not a user ID, so it's OK to leave it. Fabien is right about developers using this for their project, but we always said that this is a demo project, not a real project.

PR #694, opened 2017-11-07
Removed jQuery from auto provided variables in webpack config

mathop:
`window.jQuery` is already defined in `autoProvidejQuery` function, there's no need to override:

https://github.com/symfony/webpack-encore/blob/v0.9.1/lib/WebpackConfig.js#L271

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -6,7 +6,6 @@ Encore
     .cleanupOutputBeforeBuild()
     .autoProvidejQuery()
     .autoProvideVariables({
-        "window.jQuery": "jquery",
         "window.Bloodhound": require.resolve('bloodhound-js'),
         "jQuery.tagsinput": "bootstrap-tagsinput"
     })


javiereguiluz:
You are right ... but when we did the transition to Webpack Encore, this didn't work. In fact, the `autoProvidejQuery()` option is the only Encore feature that never works for me. I'm sure it's not Encore's fault but legacy jQuery code fault ... but still, we must be careful. Could you please double check that everything works in all browsers after removing this? Thanks!

stof:
@javiereguiluz at some point, Encore reverted this change, before adding it again (after figuring out why something was not working). I think this explicit stuff might have been added in the demo during the reverting phase.

> In fact, the `autoProvidejQuery()` option is the only Encore feature that never works for me. I'm sure it's not Encore's fault but legacy jQuery code fault ... but still, we must be careful.

there is 2 kind of legacy code handling:

- legacy libraries not supporting CommonJS `require` to get `jquery`, which are part of your webpack-processed assets. This is the case that `autoProvidejQuery` is covering
- legacy application code in your project, which is **not** processed by webpack (for instance inline JS in your template) which expects to have a `jQuery` global variable (with plugins registered in it). `autoProvidejQuery` is **not** covering this case (`autoProvideVariables` is not either).

mathop:
@javiereguiluz Sorry for the delay, it seems working fine with different browsers.

javiereguiluz:
Thanks @mathop.

PR #697, opened 2017-11-10
Using monolog.yaml configuration

yceruto:
See https://github.com/symfony/recipes/pull/243/files


javiereguiluz:
Thanks @yceruto.

PR #700, opened 2017-11-20
Relying on default path for templates and translations directories

yceruto:
Also, minor fixes related to quotation in monolog.yaml configuration.


javiereguiluz:
Thanks Yonel.

PR #701, opened 2017-11-20
Make Doctrine repository as service and autowire/use it anywhere

yceruto:
Living in the edge of the [new awesome features like this](https://github.com/doctrine/DoctrineBundle/pull/727), I'm proposing before doctrine-bundle 1.8 release, show/test this great feature that simplify our lives and to see how it works! and because soon, best practices should adopt this approach by default. (In fact [maker-bundle will do it](https://github.com/symfony/maker-bundle/blob/master/src/Resources/skeleton/doctrine/Repository.php.txt) <3)

What do you think?

### DIFF
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -13,6 +13,7 @@ namespace App\Controller\Admin;
 
 use App\Entity\Post;
 use App\Form\PostType;
+use App\Repository\PostRepository;
 use App\Utils\Slugger;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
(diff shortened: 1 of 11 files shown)


src/Controller/Admin/BlogController.php
```diff
@@ -54,10 +55,9 @@ class BlogController extends AbstractController
      * @Route("/", name="admin_post_index")
      * @Method("GET")
      */
-    public function index(): Response
+    public function index(PostRepository $postRepository): Response
```
javiereguiluz: Maybe unrelated to this PR, but in other projects I see this practice: `PostRepository $postRepository` -> `PostRepository $posts`. Instead of `$xxxRepository`, they use the plural of `$xxx` as the variable name. E.g. `UserRepository $users`, `ProductRepository $products`, etc. What do you think?

yceruto: Yes, I like that and feels better! so done.

src/Controller/BlogController.php
```diff
@@ -47,15 +48,14 @@ class BlogController extends AbstractController
...
     {
-        $em = $this->getDoctrine()->getManager();
-        $posts = $em->getRepository(Post::class)->findLatest($page);
+        $latestPosts = $posts->findLatest($page);
 
         // Every template name also has two extensions that specify the format and
         // engine for that template.
         // See https://symfony.com/doc/current/templating.html#template-suffix
-        return $this->render('blog/index.'.$_format.'.twig', ['posts' => $posts]);
+        return $this->render('blog/index.'.$_format.'.twig', ['posts' => $latestPosts]);
```
yceruto: What about just `['posts' => $posts->findLatest($page)]`?

javiereguiluz: Not sure ... the line would be a bit long.

yceruto:
Updated commands too. Ready for me! :)

yceruto:
doctrine-bundle 1.8.0 is out! We have repositories as services as a native built-in feature! :tada: :tada:

javiereguiluz:
Merged! I looooove this. @yceruto thanks for working on this!

PR #707, opened 2017-11-25
Allow PHP 7.2 & 7.3 and PHPUnit 6

VolCh:
Switch to PHPUnit 6 `TestCase::expectException()` & `TestCase::expectExceptionMessage()` from PHPUnit 4 `TestCase::setExpectedException()`

Add PHP 7.2 & 7.3 (nightly) to Travis matrix with temporary workarounds for disable `xdebug` & `./vendor/bin/php-cs-fixer` until 7.3 full support

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,13 @@ matrix:
   fast_finish: true
   include:
     - php: 7.1
+    - php: 7.2
+    - php: nightly
+  allow_failures:
+    - php: nightly
 
 before_install:
-  - phpenv config-rm xdebug.ini
+  - '[[ "$TRAVIS_PHP_VERSION" == "nightly" ]] || phpenv config-rm xdebug.ini'
   - composer self-update
 
 install:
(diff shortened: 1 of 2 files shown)


VolCh:
@javiereguiluz let's be ready for PHP 7.2 release

javiereguiluz:
This looks OK to me. @VolCh thanks for contributing it. I'll wait a bit before merging in case there are more comments. Thanks!

.travis.yml
```diff
@@ -14,9 +14,13 @@ matrix:
...
   include:
     - php: 7.1
+    - php: 7.2
+    - php: nightly
+  allow_failures:
+    - php: nightly
 
 before_install:
-  - phpenv config-rm xdebug.ini
+  - '[[ "$TRAVIS_PHP_VERSION" == "nightly" ]] || phpenv config-rm xdebug.ini'
```
bocharsky-bw: Why don't we remove Xdebug config for `nightly`?

VolCh: It is absent in `nightly` now

stof: nightly does not have XDebug on Travis, because XDebug itself is often not compatible with nightly yet

PR #708, opened 2017-11-28
Updates links to docs regarding creating custom Repository Classes.

phpeek:
Appears this section of the book has been moved to separate page.


javiereguiluz:
@phpeek well spotted! Thanks for fixing it.

PR #709, opened 2017-11-28
Improved code thanks to PhpStorm and Php Inspections

javiereguiluz:


src/Command/AddUserCommand.php
```diff
@@ -199,7 +199,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $event = $stopwatch->stop('add-user-command');
         if ($output->isVerbose()) {
-            $this->io->comment(sprintf('New user database id: %d / Elapsed time: %.2f ms / Consumed memory: %.2f MB', $user->getId(), $event->getDuration(), $event->getMemory() / pow(1024, 2)));
+            $this->io->comment(sprintf('New user database id: %d / Elapsed time: %.2f ms / Consumed memory: %.2f MB', $user->getId(), $event->getDuration(), $event->getMemory() / (1024 ** 2)
+            ));
```
stof: this change looks suspicious to me

javiereguiluz: You mean `pow(1024, 2)` -> `(1024 ** 2)` change?

stof: no, the CS change

PR #710, opened 2017-11-28
Remove unnecessary configuration for controllers

yceruto:
This configuration is only necessary for service controllers that don't extend from `Controller` or `AbstractController`, which is not the case here.

It is not clear in the recipe (in this sense) but the documentation is clear enough:
> http://symfony.com/doc/current/controller.html#fetching-services-as-controller-arguments
> If you're not using the default configuration, you can tag your service manually with `controller.service_arguments`.


javiereguiluz:
Thank you Yonel.

PR #716, opened 2017-12-02
Removes explicit dependency on DoctrineBundle.

phpeek:
This dependency is implicitly provided by `symfony/orm-pack`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,6 @@
     "require": {
         "php": "^7.1.3",
         "ext-pdo_sqlite": "*",
-        "doctrine/doctrine-bundle": "^1.8",
         "doctrine/doctrine-fixtures-bundle": "^3.0",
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@phpeek thanks for this. I'm going to merge it because it's the right thing to do now. However, we're thinking of getting rid of "packs" on composer.json. You will still do "composer require ...pack", but we're thinking about "unfolding" the pack and adding the real dependencies on your composer.json, because that's easier to understand. If this idea is implemented, we'd need to do the opposite change ... add DoctrineBundle and remove the ORM pack. We'll see :)

phpeek:
@javiereguiluz I'm very glad to hear You guys are planning to turn `*-pack`s into unfolded deps. Sounds like a right thing to do :+1:. Thanks heads up !

PR #717, opened 2017-12-02
Adding debug pack.

phpeek:
I had to edit `symfony.lock` by hand in order to make diff not too scary.

It would be an awesome feature to have `symfony.lock` sorted alphabetically as it is happening with `composer.lock`. :(

There's still `TODO` item in monolog `EasyCorp\EasyLog\EasyLogHandler` handler on how to solve configuring it which I think might be a blocker for introducing `debug-pack` here. I've solved it in a way I'm using `debug-pack` in my project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 ###< symfony/framework-bundle ###
 
 ###> symfony/phpunit-bridge ###
+.phpunit
 /phpunit.xml
 ###< symfony/phpunit-bridge ###
 
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Execute this command to run tests:
 
 ```bash
 $ cd symfony-demo/
-$ ./vendor/bin/simple-phpunit
+$ php bin/phpunit
 ```
 
 [1]: https://symfony.com/doc/current/reference/requirements.html
--- a/bin/phpunit
+++ b/bin/phpunit
@@ -1,0 +1,18 @@
+#!/usr/bin/env php
+<?php
+
+if (!file_exists(dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit')) {
+    echo "Unable to find the `simple-phpunit` script in `vendor/symfony/phpunit-bridge/bin/`.\n";
+    exit(1);
+}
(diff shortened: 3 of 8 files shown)


phpeek:
Since my PR to `symfony/flex` has been accepted by @fabpot - https://github.com/symfony/flex/pull/222 and released in https://github.com/symfony/flex/releases/tag/v1.0.44, I've decided to regenerate `symfony.lock` so that it is now sorted. This is gonna make reviewing `symfony.lock` diffs easier.

However, in light of @javiereguiluz comment in my previous PR - https://github.com/symfony/demo/pull/716#issuecomment-348763017, I will close this PR if You decide to wait till unfolding `*-pack`s dependencies in place is released.

Also `composer.lock` was conflicting so I've issued `composer update` to rehash it and updated `symfony/flex` and `sensio/framework-extra-bundle`.

jkufner:
@phpeek Don't wait for the unfolding of the packs. Sorted lock files are useful anyway.

phpeek:
@jkufner Granted. :) However, my PR introduces more "opinionated" change which is configuring `easylog` handler for `dev`'s environment monolog configuration. This is currently marked to as `TODO` in `easy-log-handler`'s recipe - https://github.com/symfony/recipes/blob/master/easycorp/easy-log-handler/1.0/config/packages/dev/easy_log_handler.yaml#L6 :(

javiereguiluz:
@phpeek thanks for this. Don't worry about the lock file diff. About the other changes:

* Not sure about the EasyLog activation. If Symfony doesn't do that... Besides, what if we remove the Monolog dependency entirely and relay on [the new and simple built-in Symfony logger](https://symfony.com/blog/new-in-symfony-3-4-minimalist-psr-3-logger)?
* Abut PHPUnit, there's some active discussion on the Symfony Recipes server about that: https://github.com/symfony/recipes/pull/293

phpeek:
@javiereguiluz Very good points !

I'll close this PR and submit new one with `symfony.lock` reordering only.

Then, I'll look into removing `monolog` dependency. Having plain Symfony logger in demo app is much more natural approach IMO.

PR #718, opened 2017-12-04
Adds Doctrine's mapping validation on Travis.

phpeek:
Runs `doctrine:schema:validate` command with database sync skipping and
no interaction flags.

This is what I usually setup in CI for my projects to avoid common Doctrine's mapping errors e.g. missing `inversedBy` properties.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,3 +41,5 @@ script:
   - ./bin/console security:check --end-point=http://security.sensiolabs.org/check_lock
   # this checks that the composer.json and composer.lock files are valid
   - composer validate --strict
+  # this checks that Doctrine's mapping configurations are valid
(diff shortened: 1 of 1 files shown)


stof:
Note that for my work project, I even remove the `--skip-sync` option, so that forgetting to generate a DB migration (or making a mistake in it) also makes it red.

javiereguiluz:
@stof would you recommend to remove `--skip-sync` here too? Thanks!

stof:
it depends how you handle schema updates in the project.
Looking at the Travis config, it does not seem to be creating the schema from scratch itself with `doctrine:schema:create` or `doctrine:schema:update`, so it might make sense to ensure the sync too, if the update of the schema requires an action by the contributor.

javiereguiluz:
OK. Let's merge this one as is, and when we start working with migrations, we can tweak the command. Thanks!

PR #721, opened 2017-12-04
[Validators PL translation] Fixing declension.

piotrgradzinski:
Small fix on PL validators translation to support message declension.

Done in the same way as it was done in CS validators translation.

P.S. I've closed previous PR because I've used wrong account. Sorry for the problem.


javiereguiluz:
@piotrgradzinski you are right: in Symfony we use the same translation as you are proposing here. Thanks!

PR #724, opened 2017-12-05
Tags fixtures with `doctrine.fixture.orm`.

phpeek:
`^3.0` releases of `DoctrineFixturesBundle` treats services as fixtures when they're tagged with `doctrine.orm.fixture`.

Not sure if this is the most optimal way to do it.

Before:

```bash
php bin/console doctrine:fixtures:load -n

In LoadDataFixturesDoctrineCommand.php line 95:
                                                
  Could not find any fixture services to load.                                    
```
After:

```bash
php bin/console doctrine:fixtures:load -n  
  > purging database
  > loading App\DataFixtures\UserFixtures
  > loading App\DataFixtures\TagFixtures
  > loading App\DataFixtures\PostFixtures
```

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -28,7 +28,11 @@ services:
         resource: '../src/*'
         # you can exclude directories or files
         # but if a service is unused, it's removed anyway
-        exclude: '../src/{Entity,Migrations,Tests}'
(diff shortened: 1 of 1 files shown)


phpeek:
Closing since this is fixed in scope of - https://github.com/symfony/demo/pull/725

PR #726, opened 2017-12-05
Use rootform test to check by root form view

yceruto:
This is consistent with the latest changes in the form themes and avoids any future collision between form view properties and form fields.

See https://github.com/symfony/symfony/pull/25236


javiereguiluz:
Thanks Yonel.

PR #725, opened 2017-12-05
Fix Doctrine fixtures service auto-configuration

yceruto:
I've missed this during the update to DoctrineFixturesBundle 3.0

PR #727, opened 2017-12-05
Revamped the data fixtures

javiereguiluz:
After #724 and #725 I realized that our fixtures are not following the modern best practices recommended by DoctrineFixturesBundle ([explained here](https://symfony.com/doc/master/bundles/DoctrineFixturesBundle/index.html)):

* Use 1 class to load all fixtures unless you have lots of fixtures.
* Make the fixture class extend from Symfony's `Fixture` class instead of the complicated Doctrine's classes.
* Inject services in the constructor and rely on autowiring.
* etc.


yceruto:
I like this revamp, but we might need to keep the `FixturesTrait` into a separate file because it's required in `tests/Controller/Admin/BlogControllerTest.php`.

javiereguiluz:
@yceruto you are right. However, when I was going to do that change, I realized that we barely use the fixtures trait in tests. So, what do you think if we just replace the entire fixtures trait just by the the `generateRandomString()` method added in 7c3e865?

PR #730, opened 2017-12-11
Fixes loading of fixtures

### LINKED ISSUE #728: Could not find any fixture services to load.
After changing .env to mysql doctrine:schema:create finished successful.
doctrine:fixtures:load
Could not find any fixture services to load.

DavertMik:
Fixtures couldn't be located without declaring their service. Fixes #728

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -34,3 +34,7 @@ services:
     # 'arguments' key and define the arguments just below the service class
     App\EventSubscriber\CommentNotificationSubscriber:
         $sender: '%app.notifications.email_sender%'
+        
+    App\DataFixtures\:
+        resource: '../src/DataFixtures'
+        tags: [doctrine.fixture.orm]        


javiereguiluz:
@DavertMik thanks for providing this fix. However, it's probably not needed because DoctrineFixturesBundle adds this tag automatically thanks to the autoconfiguration. Here: https://github.com/doctrine/DoctrineFixturesBundle/blob/master/DependencyInjection/DoctrineFixturesExtension.php#L35-L36

I've just run `php bin/console doctrine:fixtures:load` without problems. I'm using `doctrine/doctrine-fixtures-bundle: 3.0.2`. Can you please check which is your version and if it works as expected? Thanks!

phpeek:
@DavertMik @javiereguiluz This has been fixed in scope of https://github.com/symfony/demo/pull/725.

DavertMik:
Ok. Maybe I used the older version!

PR #733, opened 2017-12-15
[Proposal] Adding phpstan checks.

phpeek:
This PR introduces `phpstan/phpstan` dependency.

More info on what `phpstan` is https://github.com/phpstan/phpstan.

For now I've enabled level 4 checks on `src/` dir.

Performing checks on `tests/` dir would probably require `phpstan-phpunit`, though this package requires `phpunit/phpunit` which conflicts with `simple-phpunit` :(

There are more extensions that we could use eg. Doctrine, but for now I wanted to keep it simple and see if You like this idea.

### DIFF
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -19,6 +19,7 @@ use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\Form\Extension\Core\Type\SubmitType;
+use Symfony\Component\Form\SubmitButton;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
@@ -100,7 +101,10 @@ class BlogController extends AbstractController
             // See https://symfony.com/doc/current/book/controller.html#flash-messages
             $this->addFlash('success', 'post.created_successfully');
 
-            if ($form->get('saveAndCreateNew')->isClicked()) {
+            /** @var SubmitButton $saveAndCreateNewButton */
+            $saveAndCreateNewButton = $form->get('saveAndCreateNew');
+
+            if ($saveAndCreateNewButton->isClicked()) {
                 return $this->redirectToRoute('admin_post_new');
             }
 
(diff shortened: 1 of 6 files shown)


src/Controller/Admin/BlogController.php
```diff
@@ -100,7 +101,10 @@ public function new(Request $request): Response
             // See https://symfony.com/doc/current/book/controller.html#flash-messages
             $this->addFlash('success', 'post.created_successfully');
 
-            if ($form->get('saveAndCreateNew')->isClicked()) {
```
javiereguiluz: Could we get more details about why the original code was not correct? Thanks!

phpeek: `phpstan` complained about this line with:

```
 ------ ---------------------------------------------------- 
  Line   src/Controller/Admin/BlogController.php             
 ------ ---------------------------------------------------- 
  104    Call to an undefined method                         
         Symfony\Component\Form\FormInterface::isClicked().  
 ------ ---------------------------------------------------- 
```

yceruto: I guess because the IDE does not found the referenced method `isClicked()` into `FormInterface` (displayed as warning), though we know that's a `ClickableInterface` instance, so I don't see nothing wrong in the original code, it's just a type-hint issue. IMO should be reverted.

phpeek: @yceruto This is an actual `phpstan` error. Without this change build will fail if I revert it.

stof: @yceruto the thing is, you know it is a ClickableInterface because you know that this field was added as a button. But static analysis does not know it.
So when adding static analysis as part of the CI, it becomes necessary to make the code work with it here.

yceruto: Yep, I understand, but type-hinting it as is does not guarantee that the method exists though. If we want do it correctly we might need to check if it's an instance of `ClickableInterface`, then overkill IMHO.

I'm in favor of use `phpstan` locally, it's great and help me found a lot of issues, but to have it as part of the CI seems excessive to me if it's necessary to modify an obvious code.

yceruto: @phpeek you could do the following to skip this analysis: `$form['saveAndCreateNew']->isClickable()` and keep it simple.

phpeek: @yceruto Sorry for the delay - I was away for holidays.

You're right with `$form['saveAndCreateNew']->isClicked()` array notation, `phpstan` is no longer complaining. However, I don't like this notation, nor I'd encourage people to use it, since with `->get()`, You're getting a benefit of `OutOfBoundsException` being thrown in case of non existent key.

phpeek:
Closing this for now, since getting `phpstan` to work here, would mean writing some boilerplate code and complicating things.

PR #735, opened 2017-12-19
Allow to search blog posts with unicode characters

### LINKED ISSUE #734: Support accented chars in search
Hello,

Currently query search sanitization replace all non alphanumeric chars :

```preg_replace('/[^[:alnum:] ]/', '', ...);```

and accented chars like éèà... are removed too, making posts not findable for accented articles. Maybe supporting accented chars would be nice for all non english people using this demo app as a base ?

javiereguiluz:
This fixes #734.

The search engine is now more strict because it's case sensitive (if the post title is `Lorém ipsum`, `lorém` and `Lorem` gives no results; you must search for `Lorém`). I'd say it's OK because we're not building a real search engine. Thoughts?

### DIFF
--- a/src/Repository/PostRepository.php
+++ b/src/Repository/PostRepository.php
@@ -94,7 +94,7 @@ class PostRepository extends ServiceEntityRepository
      */
     private function sanitizeSearchQuery(string $query): string
     {
-        return preg_replace('/[^[:alnum:] ]/', '', trim(preg_replace('/[[:space:]]+/', ' ', $query)));
+        return trim(preg_replace('/[[:space:]]+/', ' ', $query));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jlagneau:
Thanks for this.

Why did you remove `mb_strtolower` in `extractSearchTerms` though? Case insensitive searches works great with just your changes in `sanitizeSearchQuery`. It just won't work for uppercase non-ASCII characters in some DBMS like sqlite but it works out of the box with MySQL for example.

jkufner:
It is better to leave worrying about case sensitivity to database. Depending on configured collation it can compare strings according to local conventions. MySQL deals with this issue well, Sqlite not so much (at least in the default configuration) and it is ok ­— we should not expect much from the little Sqllite.

I think the strtolower is an ugly hack to fix incompetent database. I would not expect such a thing in example application.

javiereguiluz:
@jlagneau I prefer to not add this warning message yet because that requires a new content to translate in all languages and because by default we use SQLite and non-accented characters, so this always work as expected for most users.

@jkufner I know that your comments were well intended, but I'd like to add a minor clarification about this --> *"... to fix incompetent database"*  We, the Symfony project, don't believe that SQLite is incompetent and we respect SQLite profoundly, as we do with all the other databases.

PR #738, opened 2017-12-25
DoctrineFixturesBundle for dev and test only

ptrm04:
According to https://symfony.com/doc/current/bundles/DoctrineFixturesBundle/index.html this bundle "should" be only enabled in dev and test - any reason why symfony/demo needs in other envs?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,6 @@
     "require": {
         "php": "^7.1.3",
         "ext-pdo_sqlite": "*",
-        "doctrine/doctrine-fixtures-bundle": "^3.0",
         "erusev/parsedown": "^1.6",
         "ezyang/htmlpurifier": "^4.9",
         "sensio/framework-extra-bundle": "^5.0",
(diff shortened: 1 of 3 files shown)


javiereguiluz:
@ptrm04 thanks for this. However, it looks like we need this bundle in production to create the original application data. In previous Symfony versions, where we controlled the bundle list manually, we added a comment explaining this: https://github.com/symfony/demo/blob/3ac81645806f0e2b6d0b4418fff73d6a44bf5aa9/app/AppKernel.php#L24

But now we have the `config/bundles.php` file that is managed by Symfony and where we can't put any comment. So, I see two possible solutions:

* Do this change anyway and forget about data fixtures in production because this is not a real app and the way we distribute this app it will always work;
* Don't do this change but add some comment somewhere explaining this.

bocharsky-bw:
I'm leaning toward the 1st, probably we don't need fixtures in production since we committed populated SQLite DB to the repo which is enough on our side. If someone prefers to run it on another SQL DB - they need to perform more steps anyway like changing default DB credentials at least. So one more step to load fixtures manually is not a problem in my opinion

phpeek:
I'm with @bocharsky-bw.

We need to assume people are looking at `symfony/demo` as a way they should develop their own apps. We should not encourage people to have `DoctrineFixturesBundle` registered with production environment. Running `doctrine:fixtures:load` on production database by mistake could end up with a huge headache. :)

javiereguiluz:
Let's merge this ... but we'll do that in #805, which is the same as this PR but without conflicts. Thanks!

PR #740, opened 2017-12-28
Use Twig Bundle in require

MatthieuMota:
I think you must use twig-bundle directly in require section because it is needed for project to run with composer --no-dev for example. It can easily replace twig dependency because twig bundle depends of twig :)


javiereguiluz:
@MatthieuMota much better indeed! Thanks.

PR #745, opened 2018-01-07
Use Sf4 path for security.yaml file

mstrom:


javiereguiluz:
Good catch, thanks @mstrom.

PR #747, opened 2018-01-07
Correcting sf4 config path

mstrom:


config/packages/dev/swiftmailer.yaml
```diff
@@ -1,2 +1,2 @@
-#swiftmailer:
-    #delivery_address: me@example.com
+swiftmailer:
+    disable_delivery: true
```
javiereguiluz: I'm not sure about this change because in this app we try to be in sync as much as possible with the Symfony recipes. This option is used in the test config (https://github.com/symfony/recipes/blob/master/symfony/swiftmailer-bundle/2.5/config/packages/test/swiftmailer.yaml) but not in the dev config (https://github.com/symfony/recipes/blob/master/symfony/swiftmailer-bundle/2.5/config/packages/dev/swiftmailer.yaml).

javiereguiluz:
Thank you @mstrom.

PR #754, opened 2018-01-23
Throw a runtimeexception if class does not exist.

obozaxy:


javiereguiluz:
@obozaxy sorry for not having replied to you earlier. Although your pull request is not wrong, I'm afraid we must close it without merging. The main reason is that we don't modify in this project any file generated by Symfony itself (and this `console` file comes from https://github.com/symfony/recipes/blob/master/symfony/console/3.3/bin/console). This is done to avoid confusion to users that look at Symfony Demo and their own app and they behave differently.

The other reason is that this is considered correct by Symfony. I agree with you that this could be better, but Symfony's reasoning is: if you use env vars from `.env` file, you have DotEnv installed and this works. If you don't use the `.env` file, then you must set up these env vars yourself and `APP_ENV` is defined and this error doesn't happen.

Some users seem to disagree (https://github.com/symfony/recipes/issues/350, https://github.com/symfony/recipes/issues/235, https://github.com/symfony/recipes/issues/270) but that's something that Symfony may or may not ix itself instead of fixing it in this app. Thanks!

Issue #741, opened 2017-12-29
Validation error in edit form


nicolaspetitjean:
Hi

If i disable html validation in edit form and submit the form with empty value :
Type error: Argument 1 passed to App\Entity\Post::setTitle() must be of the type string, null given
<img width="1099" alt="capture d ecran 2017-12-29 a 21 42 27" src="https://user-images.githubusercontent.com/11570379/34447110-399dbcfa-ece1-11e7-8641-5b1a4539efa7.png">

bocharsky-bw:
Hm, nice catch! Sounds like a big in Symfony Validator. You can fix it with:
```php
public function setTitle(?string $title): void
```

i.e. also allow passing `null` to `setTitle()`, but I think it's just a workaround

nicolaspetitjean:
The most surprising thing is that the issue does not exists in the add form...

javiereguiluz:
I don't understand the issue well. You are visiting the password-protected backend, then disable the HTML validation using the browser code inspector and submit the form. The application responds with an error page and the wrong post is not added in the database. Is that correct or am I missing something? Thanks!

nicolaspetitjean:
According to entity def :
`/**
* @var string
*
* @ORM\Column(type="string")
* @Assert\NotBlank
*/
private $title;`
Empty value is not allowed, the edit form should act like the add form :
![capture du 2017-12-30 18-33-03](https://user-images.githubusercontent.com/11570379/34456141-ed706c42-ed8f-11e7-936d-b3c02b2b8335.png)
It's not the case, maybe it is me who is wrong

bocharsky-bw:
@javiereguiluz the problem only with edit form, since fields already have data. So if you clear the field, I.e. make it empty - the null value is set to the field, which fails due to strict type hint. It’s happened before validation.

With add form you won’t be able to reproduce it, because fields are set to null by default

phpeek:
Unfortunately validation doesn't work "well" with strict typing. :(

By strict typing Your entities, You're automatically preventing them from being in an invalid state.

I was not able to find reasonable workaround for those types of situations, this is why I'm looking into what are You going to suggest here. :)

What I'm doing in those kind of situations is either what @bocharsky-bw suggested - marking setter arguments as nullable `?string` or setting `'empty_data' => ''` when building form fields. Both of those approaches are not really elegant, though I find the first approach to cause less head scratching. :)

bocharsky-bw:
[...]

PR #757, opened 2018-01-30
Minor code fixes found with PhpStorm, Php Inspections and PHPStan

javiereguiluz:


src/Entity/Comment.php
```diff
@@ -126,12 +126,12 @@ public function setAuthor(User $author): void
...
     }
 
-    public function getPost(): Post
+    public function getPost(): ?Post
     {
         return $this->post;
     }
 
-    public function setPost(Post $post): void
+    public function setPost(?Post $post): void
```
yceruto: Btw, could we do the same in `Post::setTitle()` & Co. to fix https://github.com/symfony/demo/issues/741?

javiereguiluz: Good idea! Done.

PR #758, opened 2018-02-01
Removed exclusion of Tests folder in src

mickaelandrieu:


javiereguiluz:
Mickaël thanks for this contribution! The changes you are proposing are perfectly valid ... however, in this demo app we always try to keep the config files in sync with the default config files provided by Symfony flex recipes (to lower the learning curve).

Si, I propose a different thing. Instead of making this change, let's make other changes to this file to make it as similar as possible to https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/config/services.yaml. We've done that in #759 and that's why we're closing this. Thanks!

PR #765, opened 2018-02-12
Filter posts by selected tag

### LINKED ISSUE #762: Enhance Tag feature to work as a search for similar blog posts
Hello.  I think the Tag feature is great, but it seems incomplete as an example. When browsing the application, I would expect clicking a tag would show all other blog posts that have the same tag.

My request is enhancement to demonstrate best practices to implement this added ability.  Thank you.

yceruto:
Fixes #762

Even though this PR doesn't show any new Symfony feature, it makes the post tags have some sense.

### DIFF
--- a/src/Controller/BlogController.php
+++ b/src/Controller/BlogController.php
@@ -16,6 +16,7 @@ use App\Entity\Post;
 use App\Events;
 use App\Form\CommentType;
 use App\Repository\PostRepository;
+use App\Repository\TagRepository;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Cache;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
@@ -46,9 +47,13 @@ class BlogController extends AbstractController
      * Content-Type header for the response.
      * See https://symfony.com/doc/current/quick_tour/the_controller.html#using-formats
      */
-    public function index(int $page, string $_format, PostRepository $posts): Response
+    public function index(Request $request, int $page, string $_format, PostRepository $posts, TagRepository $tags): Response
     {
-        $latestPosts = $posts->findLatest($page);
+        $tag = null;
+        if ($request->query->has('tag')) {
+            $tag = $tags->findOneBy(['name' => $request->query->get('tag')]);
+        }
+        $latestPosts = $posts->findLatest($page, $tag);
 
         // Every template name also has two extensions that specify the format and
         // engine for that template.
--- a/src/Repository/PostRepository.php
+++ b/src/Repository/PostRepository.php
@@ -12,6 +12,7 @@
 namespace App\Repository;
 
 use App\Entity\Post;
+use App\Entity\Tag;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\Query;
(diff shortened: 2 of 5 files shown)


src/Repository/PostRepository.php
```diff
@@ -35,21 +35,27 @@ public function __construct(ManagerRegistry $registry)
...
+            ->setParameter('now', new \DateTime());
+
+        if (null !== $tag) {
+            $qb->andWhere('p IN (
+                    SELECT p2 
+                    FROM App:Post p2
+                    JOIN p2.tags t2
+                    WHERE t2.name = :tag)
+                ')
+                ->setParameter('tag', $tag);
```
voronkovich: @yceruto, I think it's better to use a "secret" DQL function `MEMBER OF` here :). See http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/d)ql-doctrine-query-language.html#dql-select-examples
```php
if (null !== $tag) {
    $qb->andwhere(':tag MEMBER OF p.tags')
       ->setParameter('tag', $tag);
}   
```
To do so, you need to get the tag in the controller:
```php
$tag = $tagRepository->findOneByName($request->query->get('tag'));
$latestPosts = $postRepository->findLatest($page, $tag);
```

templates/blog/_post_tags.html.twig
```diff
@@ -1,9 +1,9 @@
 {% if not post.tags.empty %}
     <p class="post-tags">
         {% for tag in post.tags %}
-            <span class="label label-default">
+            <a href="{{ path('blog_index', {'tag': tag.name}) }}" class="label label-default">
```
voronkovich: @yceruto, What about to change a style of the current selected tag by adding a different CSS class for him (`label-primary` or `label-info`)? See https://getbootstrap.com/docs/3.3/components/#available-variations

src/Repository/PostRepository.php
yceruto: Excellent idea! That would avoid the subquery, thanks!

templates/blog/_post_tags.html.twig
yceruto: I like that, done.

yceruto:
This way selected tag is highlighted:

highlight-tag

Credit to @voronkovich

src/Repository/PostRepository.php
yceruto: However, I don't like Doctrine's magic methods `findOneByName`, the IDE can't guess it ;)

bocharsky-bw: Just FYI, you can add an annotation to help your IDE like:
```
/**
  * @method Tag findOneByName(string $name)
  */
TagRepository
{
    // ...
}
```

yceruto: Yes, I know :) I meant intuitively, thanks you for share it.

yceruto:
Is this something we want to improve? or does this seem irrelevant for demo purpose?

voronkovich:
This feature:

- looks great;
- requires so little changes in the codebase;
- shows the use case for the Doctrine's operator `MEMBER OF`.

IMHO, this PR should be merged.

yceruto:
Rebased and `label-success` applied.

src/Controller/BlogController.php
```diff
@@ -46,9 +47,10 @@ class BlogController extends AbstractController
      * Content-Type header for the response.
      * See https://symfony.com/doc/current/quick_tour/the_controller.html#using-formats
      */
-    public function index(int $page, string $_format, PostRepository $posts): Response
+    public function index(Request $request, int $page, string $_format, PostRepository $posts, TagRepository $tags): Response
     {
-        $latestPosts = $posts->findLatest($page);
+        $tag = $tags->findOneBy(['name' => $request->query->get('tag')]);
```
stof: I think this should be wrapped in a check whether the query string param is used, to avoid doing a DB query to load a tag when you don't have a tag name.

src/Controller/BlogController.php
```diff
@@ -46,9 +47,13 @@ class BlogController extends AbstractController
      * Content-Type header for the response.
      * See https://symfony.com/doc/current/quick_tour/the_controller.html#using-formats
      */
-    public function index(int $page, string $_format, PostRepository $posts): Response
+    public function index(Request $request, int $page, string $_format, PostRepository $posts, TagRepository $tags): Response
```
sebastianblum: why do you call the repository $tags instead of $repository?

javiereguiluz: `$repository` is too broad ... is it a tag repository, a post repository, a user repository? It should be `$tagRepository` .... but we decided to try using this way of naming vars related to repository: instead of `$fooInSingularRepository` we call it `$fooInPlural`. This decision is not definitive. We're still thinking about this (by the way, other frameworks do this too).

OlivierToussaint: tagRepository is perfect for me
but if we rename $tags, we need to rename $posts

javiereguiluz:
Thanks a lot Yonel ... and I'm sorry it took me so long to merge it!

PR #766, opened 2018-02-16
[Fix] Move FormType in the right folder

orions:
Hello,

Why `PostType` and `CommentType` are under the namespace `App\Form` and other formType are under the namespace `App\Form\Type` ?


jzawadzki:
:-1: I think current way is correct
- in `App\Form` we keep forms (technically form types, yes) that are entity-level e.g. form for Post (such types won't be used as types in other forms)
- in `App\Form\Type` we keep custom form types that can be reused across many different forms (e.g. for DateTimePicker)

orions:
@jzawadzki i think all types should be in `App\Form\Type` but when i read the document
* https://symfony.com/doc/current/forms.html => `App\Form`
* https://symfony.com/doc/current/form/create_custom_field_type.html => `App\Form\Type`

javiereguiluz:
@yceruto what do you think about this proposal?

yceruto:
I'd say they are two groups of forms: Primary form types (root forms with at least one field, compound) are usually used in controllers, it should be placed in `Form` directory; Secondary form types (form field types) are generally used into other forms, it should be placed in `Form\Type` directory.

That's the case here, `PostType` and `CommentType` are primary forms, `DateTimePickerType` and `TagsInputType` are secundary forms. So their current positions are fine to me.

PR #767, opened 2018-02-17
Update ValidatorTest.php

appointmind:
The method name testValidateEmailFullName() seemed to be inconsistent with the purpose and the rest of the method names.


javiereguiluz:
@appointmind much better now! thanks for your first contribution to this project.

PR #768, opened 2018-02-17
update dama/doctrine-test-bundle to ^5.0

dmaicher:
Upgrades `dama/doctrine-test-bundle` to latest version which is compatible with PHPUnit 7 as well.

Updated travis config to not use legacy PHPUnit 5 anymore.


javiereguiluz:
@dmaicher cool! let's keep using cutting-edge versions in this project :)

PR #782, opened 2018-04-19
Updated Doctrine repositories for DoctrineBundle 1.9

### LINKED ISSUE #783: not working after composer update
after composer update command I receive this error:

Cannot autowire service "App\Repository\PostRepository": argument "$registry" of method "__construct()" references interface "Doctrine\Common\Persistence\ManagerRegistry" but no such service exists. You should maybe alias this interface to the existing "doctrine" service.

could you explain what is wrong, please

javiereguiluz:
DoctrineBundle 1.9 (https://github.com/doctrine/DoctrineBundle/releases/tag/1.9.0) was released yesterday and it changes one of the interfaces used in the new repositories.

This fixes #783.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,23 +1,23 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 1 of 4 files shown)


javiereguiluz:
@xabbuh I can't make tests work after this change. But errors don't have anything to do with the change. In my local computer, some tests return a 500 page and the error message says "DoctrineBundle is not enabled in the app" :astonished:

javiereguiluz:
Changes are finally being reverted (see https://github.com/doctrine/DoctrineBundle/pull/805) so we won't need these changes for now. Closing.

stof:
The 500 error is due to https://github.com/doctrine/DoctrineBundle/issues/804

PR #784, opened 2018-04-21
"Usage" (first use)

java1cprog:
After you create the project, you need to set the vendor dependencies by "composer install" command. Only after that you can run the application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ browser at <http://localhost:8000>:
 
 ```bash
 $ cd symfony-demo/
+$ composer install
 $ php bin/console server:run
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@java1cprog thank for this contribution but I'm afraid there's been some misunderstanding. The official installation command that we recommend and the only one we show in the README is `composer create-project symfony/symfony-demo`  That command installs all the dependencies, so there's no need to execute `composer install`.

That's why we need to close this unmerged. Cheers!

java1cprog:
Many thanks for such a detailed answer.

PR #790, opened 2018-04-23
disable the deprecated sensio_framework_extra.router.annotations

### LINKED ISSUE #786: Our current framework_extra.yaml config is deprecated since Symfony 4.1
After upgrading to Symfony 4.1@dev you can see this deprecation:

> Enabling the "sensio_framework_extra.router.annotations" configuration is deprecated since version 5.2. Set it to false and use the "Symfony\Component\Routing\Annotation\Route" annotation from Symfony itself.

christickner:
fixes #786 

`famework.annotations` still needs to be true due to the `validation: { enable_annotations: true }` in `framework.yaml`. 

Should I move the `framework.annotations: true` to the `framework.yaml` file instead? It seems more appropriate.

### DIFF
--- a/config/packages/framework_extra.yaml
+++ b/config/packages/framework_extra.yaml
@@ -1,6 +1,3 @@
-framework:
-    annotations: true
-
 sensio_framework_extra:
         router:
             annotations: false
--- a/config/packages/framework_extra.yaml
+++ b/config/packages/framework_extra.yaml
@@ -1,2 +1,6 @@
 framework:
     annotations: true
+
+sensio_framework_extra:
+        router:
+            annotations: false


javiereguiluz:
Not sure what's the best choice. Let's ask for help to some expert developers who are active in this repo: @voronkovich @bocharsky-bw @yceruto. Thanks!

stof:
Why do we have `framework_extra.yml` configuring `framework` ? this looks weird (and useless, as FrameworkBundle already enables the annotation layer properly based on the fact that doctrine/annotations is installed)

voronkovich:
I agree with @stof. I think we should just delete the `framework_extra.yml` file.

christickner:
agreed. removed the "framework" config from the file here: https://github.com/symfony/demo/pull/790/commits/a0638623271155d9a70da1828cc6bc30be7676af

@voronkovich note that we can't delete the file entirely, as we have to explicitly disable `sensio_framework_extra.router.annotations` for this issue to be solved.

All: should I squash, rebase on master, and force-push, or are we okay as is?

javiereguiluz:
@christickner thanks for your work on this! There's no need to squash or rebase because the tool we use to merge things do that automatically. Cheers!

Issue #792, opened 2018-04-29
Use @IsGranted annotation instead of @Security


voronkovich:
Recently a new `@IsGranted` annotation was introduced. It's just a shortcut for the well-known `@Security` annotation. See: https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/Configuration/IsGranted.php

javiereguiluz:
I think this could be a good idea to show the new annotation in action.

PR #797, opened 2018-05-10
Use @IsGranted instead of @Security.

DevMDamien:
As asked in issue #792, I replaced some `@Security("is_granted()")` annotations by `@IsGranted`.

### DIFF
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -15,6 +15,7 @@ use App\Entity\Post;
 use App\Form\PostType;
 use App\Repository\PostRepository;
 use App\Utils\Slugger;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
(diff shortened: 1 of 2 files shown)


src/Controller/Admin/BlogController.php
```diff
@@ -130,11 +131,10 @@ public function show(Post $post): Response
      * Displays a form to edit an existing Post entity.
      *
      * @Route("/{id}/edit", requirements={"id": "\d+"}, methods={"GET", "POST"}, name="admin_post_edit")
+     * @IsGranted("edit", subject="post", message="Posts can only be edited by their authors.")
      */
     public function edit(Request $request, Post $post): Response
     {
-        $this->denyAccessUnlessGranted('edit', $post, 'Posts can only be edited by their authors.');
```
yceruto: It would be good to keep this as another way of doing the same but without the annotation one.

DevMDamien: I let an example in the show action ([see the code](https://github.com/DevMDamien/demo/blob/137b484b25e788d43668ce352da420993d62529c/src/Controller/Admin/BlogController.php#L121-L123))

PR #807, opened 2018-05-30
Update to Symfony 4.1

javiereguiluz:


yceruto:
Javier, did you already think about some features of 4.1 that can be useful to implement here?

javiereguiluz:
I haven't thought about that actually. Which features would you like to see implemented? (in other pull requests, not this one 😅)

PR #810, opened 2018-06-07
Fix bug when loading bundles classes by environment

NAYZO:
| Q             | A
| ------------- | ---
| License       | MIT

Original PR: https://github.com/symfony/recipes/pull/417

Bundle classes are not loaded correctly, for example:
```php
// config/bundles.php
<?php

return [
Symfony\Bundle\FrameworkBundle\FrameworkBundle::class => ['dev' => false, 'prod' => true],
```
The FrameworkBundle will be loaded for both **dev** and for **prod** environments even with `'dev' => false` !!
  
Same thing for:
```php
// config/bundles.php
<?php

return [
Symfony\Bundle\FrameworkBundle\FrameworkBundle::class => ['all' => false],
```

### DIFF
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -38,7 +38,7 @@ class Kernel extends BaseKernel
     {
         $contents = require $this->getProjectDir().'/config/bundles.php';
         foreach ($contents as $class => $envs) {
-            if (isset($envs['all']) || isset($envs[$this->environment])) {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I'm afraid that we must close this because the related change in the Symfony recipes was also closed. @NAYZO I agree with you that this may be a bit confusing, but Fabien gave a compelling reason for doing this: performance. Thanks!

Issue #811, opened 2018-06-17
Error when running TestSuite


siewert87:
A fresh installation of this project gives me this when running the tests:

```bash
bash-4.3$ vendor/bin/simple-phpunit 
PHP Fatal error:  Declaration of DAMA\DoctrineTestBundle\PHPUnit\PHPUnitListener::startTestSuite(PHPUnit\Framework\TestSuite $suite) must be compatible with PHPUnit_Framework_TestListener::startTestSuite(PHPUnit_Framework_TestSuite $suite) in /var/www/html/vendor/dama/doctrine-test-bundle/src/DAMA/DoctrineTestBundle/PHPUnit/PHPUnitListener.php on line 9
PHP Stack trace:
PHP   1. {main}() /var/www/html/vendor/symfony/phpunit-bridge/bin/simple-phpunit:0
PHP   2. include() /var/www/html/vendor/symfony/phpunit-bridge/bin/simple-phpunit:280
PHP   3. PHPUnit_TextUI_Command::main() /var/www/html/vendor/bin/.phpunit/phpunit-5.7/phpunit:17
PHP   4. PHPUnit_TextUI_Command->run() /var/www/html/vendor/bin/.phpunit/phpunit-5.7/src/TextUI/Command.php:116
PHP   5. PHPUnit_TextUI_TestRunner->doRun() /var/www/html/vendor/bin/.phpunit/phpunit-5.7/src/TextUI/Command.php:186
PHP   6. Symfony\Bridge\PhpUnit\Legacy\TestRunnerForV5->handleConfiguration() /var/www/html/vendor/bin/.phpunit/phpunit-5.7/src/TextUI/TestRunner.php:163
PHP   7. PHPUnit_TextUI_TestRunner->handleConfiguration() /var/www/html/vendor/symfony/phpunit-bridge/Legacy/TestRunnerForV5.php:28
PHP   8. class_exists() /var/www/html/vendor/bin/.phpunit/phpunit-5.7/src/TextUI/TestRunner.php:931
PHP   9. spl_autoload_call() /var/www/html/vendor/bin/.phpunit/phpunit-5.7/src/TextUI/TestRunner.php:931
PHP  10. Composer\Autoload\ClassLoader->loadClass() /var/www/html/vendor/bin/.phpunit/phpunit-5.7/src/TextUI/TestRunner.php:931
[...]

PR #814, opened 2018-06-18
require phpunit 7.1 via SYMFONY_PHPUNIT_VERSION env variable

dmaicher:
Fixes https://github.com/symfony/demo/issues/811

When running the test suite on php 7.1 the wrong phpunit (non-maintained version 5) is required by the simple-phpunit-bridge which is not compatible with https://github.com/dmaicher/doctrine-test-bundle-demo.


javiereguiluz:
Thanks for fixing this bug David.

Issue #808, opened 2018-05-30
[RFC] Add Messenger component example


yceruto:
Now that 4.1 stable is happily released I'd like to put in the board this nice component and think about how we can introduce it here.

**My Proposal**

![newslatter-subscription-box](https://user-images.githubusercontent.com/2028198/40734078-65d817e8-6405-11e8-8f4f-2d2e89964700.png)

"Blog posts subscription" or whatever you want to call it. The goal is to show an input box where any user can submit their email to subscribe for new articles, when a new article is created, we would dispatch an async message through the bus to send all emails to the subscribed users.

**About The Implementation**

To keep it simple, we might need a fake transport to simulate the async process and go without the amqp extension. Other details later if this example is approved.

---

If you have a different idea where this new component can be used or how it should be used, please put it here. Thanks.

ogizanagi:
Fake transport could delay handling to `kernel.terminate` by storing messages in-memory and process messages once the request is served. That's actually perhaps a cheap "transport" we could provide in the core. I already rely sometimes on such a system with the EventDispatcher to delay some events (of course it has limitations).

Or if we want to go the worker way (which probably makes more sense to fully sample async), a [filesystem](https://github.com/php-enqueue/enqueue-dev/blob/master/docs/transport/filesystem.md) or [doctrine](https://github.com/php-enqueue/enqueue-dev/blob/master/docs/transport/dbal.md) transport with https://github.com/php-enqueue/messenger-adapter.

cc @sroze

ogizanagi:
@yceruto : Don't know, but I don't think so. I don't see much benefits from re-implementing what is already available through php-enqueue and already usable thanks to the messenger-bridge.
The AMQP transport was only motivated by providing a sensible built-in adapter out-of-the-box AFAIK.
Perhaps we'll add other transports, but that'll probably be very limited.

sroze:
IMHO I think that we could have a Doctrine one, but that’s basically it :)

thePanz:
@yceruto I implemented a filesystem transport here: https://github.com/thePanz/messenger-filesystem-transport
It started before the Messenger component was "official" and I just wanted to play with it.
In the end I enjoyed the implementation and ended up adding some "compression" features and using it in a pre-production environment :)

yceruto:
With 4.3 we have a Doctrine transport, ideal for this example.

@javiereguiluz what do you think about this one? 
Do you think that an example of this feature could make the demo too complex?

[...]

PR #817, opened 2018-06-20
Added more great Symfony 4.1 features

javiereguiluz:
The last remaining big features are:

* Messenger component: is being discussed in #808.
* Advanced Console Output: I plan to send another PR with some changes in some commands.

### DIFF
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -114,7 +114,7 @@ class BlogController extends AbstractController
     /**
      * Finds and displays a Post entity.
      *
-     * @Route("/{id}", requirements={"id": "\d+"}, methods={"GET"}, name="admin_post_show")
+     * @Route("/{id<\d+>}", methods={"GET"}, name="admin_post_show")
      */
     public function show(Post $post): Response
     {
@@ -130,7 +130,7 @@ class BlogController extends AbstractController
     /**
      * Displays a form to edit an existing Post entity.
      *
-     * @Route("/{id}/edit", requirements={"id": "\d+"}, methods={"GET", "POST"}, name="admin_post_edit")
+     * @Route("/{id<\d+>}/edit",methods={"GET", "POST"}, name="admin_post_edit")
      * @IsGranted("edit", subject="post", message="Posts can only be edited by their authors.")
      */
     public function edit(Request $request, Post $post): Response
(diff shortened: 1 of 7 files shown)


src/Controller/BlogController.php
```diff
@@ -39,7 +39,7 @@ class BlogController extends AbstractController
     /**
      * @Route("/", defaults={"page": "1", "_format"="html"}, methods={"GET"}, name="blog_index")
      * @Route("/rss.xml", defaults={"page": "1", "_format"="xml"}, methods={"GET"}, name="blog_rss")
-     * @Route("/page/{page}", defaults={"_format"="html"}, requirements={"page": "[1-9]\d*"}, methods={"GET"}, name="blog_index_paginated")
+     * @Route("/page/{page<[1-9]\d*>}", defaults={"_format"="html"}, methods={"GET"}, name="blog_index_paginated")
```
stof: I would even change it to `\d*+`. Using possessive quantifiers in routing requirements makes matching faster

src/Controller/Admin/BlogController.php
```diff
@@ -130,7 +130,7 @@ public function show(Post $post): Response
     /**
      * Displays a form to edit an existing Post entity.
      *
-     * @Route("/{id}/edit", requirements={"id": "\d+"}, methods={"GET", "POST"}, name="admin_post_edit")
+     * @Route("/{id<\d+>}/edit",methods={"GET", "POST"}, name="admin_post_edit")
```
stof: I suggest using a possessive quantifier here (`\d++`), to improve the route matching performance (Symfony generates possessive quantifiers for its builtin requirements, but custom ones need to do that themselves, as using non-possessive quantifiers can still be a valid use case)

src/Entity/User.php
```diff
@@ -42,20 +43,24 @@ class User implements UserInterface, \Serializable
...
+     * @Assert\NotBlank()
      */
     private $fullName;
 
     /**
      * @var string
      *
      * @ORM\Column(type="string", unique=true)
+     * @Assert\NotBlank()
+     * @Assert\Length(min = 2, max = 50)
```
stof: The existing annotations are not putting spaces around the `=`

javiereguiluz:
@stof I've merged without making the changes in the routing requirements ... but I've opened #818 to discuss about your proposal. Cheers!

PR #821, opened 2018-06-23
Update BlogController.php

taochangle:
fix 

Controller "App\Controller\BlogController::index()" requires that you provide a value for the "$page" argument. Either the argument is nullable and no null value has been provided, no default value has been provided or because there is a non optional argument after this one.

### DIFF
--- a/src/Controller/BlogController.php
+++ b/src/Controller/BlogController.php
@@ -46,7 +46,7 @@ class BlogController extends AbstractController
      * Content-Type header for the response.
      * See https://symfony.com/doc/current/quick_tour/the_controller.html#using-formats
      */
-    public function index(int $page, string $_format, PostRepository $posts): Response
+    public function index(int $page=1, string $_format='html', PostRepository $posts=null): Response
     {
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@taochangle thanks for opening this PR to fix the project.

However, in this case these changes should be not necessary. In fact, the app works without changing this. There are two things to consider:

1) We already define the default values in the `@Route` annotations (and they can be different for each route, the the `_format` argument, which sometimes is `XML` and sometimes `HTML`).
2) This error happens when the [Symfony autowiring](https://symfony.com/doc/current/service_container/autowiring.html) is not working as expected. The arguments of the controller can't be injected (specifically, the `PostRepository $posts` argument) and Symfony displays that error. If you use the default autowiring config, the controllers are applied the `controller.service_arguments` tag, and autowiring works. Please, check that the `config/services.yaml` config in your project is the same as the default one and also make sure that you are using Symfony 4.1.

So, I'm closing this as "won't fix" but I hope you understand the reasons. Thanks!

PR #823, opened 2018-06-25
Use PHP_OS_FAMILY constant thanks to Symfony's PHP 7.2 polyfill

javiereguiluz:


tests/Command/AddUserCommandTest.php
```diff
@@ -33,9 +33,7 @@ protected function setUp()
         exec('stty 2>&1', $output, $exitcode);
         $isSttySupported = 0 === $exitcode;
 
-        $isWindows = '\\' === \DIRECTORY_SEPARATOR;
-
-        if ($isWindows || !$isSttySupported) {
+        if ('Windows' === PHP_OS_FAMILY || !$isSttySupported) {
```
stof: you should fully-qualify it, so that it can still be resolved at compile-time when the constant is supported natively.

PR #830, opened 2018-07-04
Reset tag filter if it's link was clicked again

voronkovich:


javiereguiluz:
@voronkovich a nice little feature! I love it. Thanks!

PR #831, opened 2018-07-13
Add missing but unused logic to PostVoter

codedmonkey:
Adds a small bit of code to accommodate the missing logic when trying to view other users' posts. Although the demo never triggers any security checks when trying to view a post, the class describes this attribute itself, so I figured we might as well add it.

### DIFF
--- a/src/Security/PostVoter.php
+++ b/src/Security/PostVoter.php
@@ -46,6 +46,11 @@ class PostVoter extends Voter
      */
     protected function voteOnAttribute($attribute, $post, TokenInterface $token): bool
     {
+        // all users are allowed to view posts
+        if (self::SHOW === $attribute) {
+            return true;
+        }
+
         $user = $token->getUser();
 
         // the user must be logged in; if not, deny permission


src/Security/PostVoter.php
```diff
@@ -48,6 +48,11 @@ protected function voteOnAttribute($attribute, $post, TokenInterface $token): bo
     {
         $user = $token->getUser();
 
+        // all users are allowed to view posts
+        if ($attribute === self::SHOW) {
```
yceruto: I think this check was already done in the `supports()` method, isn't it?

codedmonkey: No, `supports()` only checks whether the voter should be invoked, which in this case, it does.

yceruto:
> Although the demo never triggers any security checks when trying to view a post

Actually it does:
https://github.com/symfony/demo/blob/559b771967a048920781df76c5cf6ffa8bf7bbd8/src/Controller/Admin/BlogController.php#L123

All users can see any article (out of admin zone), into admin zone all articles are filtered by the logged user and he is the only one authorized to see them. Also, there is no action that allows an user to see an article of another user in admin zone (unless you try it directly through the browser's entry, which doesn't make sense for me to allow it).

codedmonkey:
Welp, no idea how I missed that one. Sorry about that.

Issue #802, opened 2018-05-21
Create the best possible docker-compose.yml for this project


javiereguiluz:
Calling for help from Docker experts. We need to create the best possible `docker-compose.yml` file for this project. The application requirements are well defined (we use env vars, Webpack Encore, PHP 7.1, Symfony 4.1, SQLite database, etc.) so it should be possible to create that file.

javiereguiluz:
An important note mentioned to me by a person on Symfony Slack (I don't remember exactly who, I'm sorry). In the normal Symfony Demo app, we must use SQLite because it's our only option to use a database that requires zero setup.

However, when using Docker, we can use any database with zero setup. So, let's use PostgreSQL in the Docker config file. (I suggest PosgreSQL because to me it looks more popular among Symfony devs than MySQL).

stof:
the hard part is that our `config/packages/doctrine.yml` file is currently incompatible with Postgresql as it defines a mysql-only charset (the proper UTF-8 charset  is the `utf8` on postgresql, as they got it right directly instead of creating a second charset like `utf8mb4` for proper UTF-8 support)

javiereguiluz:
@stof could we run a quick `sed` command on that config file to replace the wrong value while setting up things in Docker?

sebastianblum:
@javiereguiluz we can remove ```charset: utf8mb4``` in this file (https://github.com/symfony/demo/blob/master/config/packages/doctrine.yaml) and add it as a parameter ```?charset=utf8mb4``` in the env.dist file (https://github.com/symfony/demo/blob/master/.env.dist) 

```
DATABASE_URL=sqlite:///%kernel.project_dir%/var/data/blog.sqlite?charset=utf8mb4
```

Syther101:
Hay all, subscribed to this issue when it was being talked about in slack. Was toying around with docker the other day and got reminded of this. Not the most experienced with Docker but can try and help where I can 😄

Do with have any idea what base image we would want to use. `7.1-jessie` maybe?

OlivierToussaint:
we can take this : https://github.com/dunglas/symfony-docker ?

frastel:
I would vote for sqlite because you do not have to start an additional container for this database.

About the requirements:
* mailing should also be possible? Or should we ignore this feature as it is disabled by default?
* how about the webserver? Which one should be used or should the build-in webserver be used?
[...]

PR #833, opened 2018-07-16
Implementation of the Docker environment

ajardin:
Here is a first implementation of what has been requested in #802. 

I made the pull request earlier in order to receive your feedbacks about the design itself. The documentation still has to be updated. Basically, I took the decision to create something very simple (SQLite without HTTPS) and that can be used to test Symfony by directly editing the project.

The installation process is currently composed of three steps:
- `git clone git@github.com:ajardin/demo.git symfony-demo`
- `cd symfony-demo`
- `make install`

And that's all! The browser will be automatically open once the application is ready.

@javiereguiluz, could you have a look at this and tell me what you think?


docker/php/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+    docker-php-ext-install \
+        intl \
+        opcache
+
+# Install Composer globally
+ENV COMPOSER_ALLOW_SUPERUSER 1
+RUN \
+    curl -sS https://getcomposer.org/installer | php && \
+    mv composer.phar /usr/local/bin/composer && \
+    composer global require "hirak/prestissimo:dev-master" --no-suggest --optimize-autoloader --classmap-authoritative
```
stof: not needed IMO. Flex already optimizes the downloading of packages to do it in parallel (and it might now play well with `hirak/prestissimo` if they both try to do it)

ajardin: I made several tests on my local environment and Composer is still faster with `hirak/prestissimo`. You will find the results below. Did I miss something?

```
# without hirak/prestissimo
make uninstall ; rm -rf vendor/ ; time make build start composer
[...]
make build start composer  1,25s user 0,51s system 1% cpu 1:33,85 total
```

```
# with hirak/prestissimo
make uninstall ; rm -rf vendor/ ; time make build start composer
[...]
make build start composer  1,25s user 0,49s system 3% cpu 49,488 total
```

And I'm not sure, but it seems for me that the incompatibility between `symfony/flex` and `hirak/prestissimo` was fixed with https://github.com/symfony/flex/issues/381.

Makefile
```diff
@@ -0,0 +1,81 @@
...
+.PHONY: build install logs logs-full ps restart start stats stop uninstall
+
+##
+## ----------------------------------------------------------------------------
+##   Project
+## ----------------------------------------------------------------------------
+##
+
+composer: ## Install Composer dependencies from the "php" container
+	$(DOCKER_COMPOSE) exec php sh -c "composer install --optimize-autoloader"
```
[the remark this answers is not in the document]
ajardin: From my understanding, there is no real issue with `--optimize-autoloader` as its purpose is to generate a classmap file to avoid filesystem checks. A thing quite interesting in term of performances while you are using OPcache.

If a class is not present in the classmap, the autoload will still request the filesystem to load it. You can find more details a little bit further in the documentation you linked (https://getcomposer.org/doc/articles/autoloader-optimization.md#trade-offs).

javiereguiluz:
I've tested this and i's great! I tested most of the `make` commands and they all worked great.

A minor comment: the first-time installation was too long because I think it compiled PHP. If this is true, could we instead install a minimal PHP binary for a modern PHP 7.2 version instead of building it from source? Thanks!

frastel:
@javiereguiluz It doesn't compile. Docker has to build all required images before the application could be started. For the very first run this build takes quite some time as no layer exists in the cache yet and all needed PHP packages have to be downloaded (think of `apt-get update && apt-get install php-cli` when you want to provision a standard debian/ubuntu server). The rebuilds however could re-use layers and thus the installation of the required packages is ultra fast.

ajardin:
Hi @javiereguiluz, and thanks for your feedbacks!

As @frastel nicely explained, what you see is the "image build process" performed by Docker when required cache layers are not available your local machine. This is completely normal to have thousands of lines during that step:
- the base image needs to be built as describe in [its Dockerfile](https://github.com/docker-library/php/blob/85af0c14e3f23689f0851d3164ab3b630e7f016f/7.2/alpine3.7/fpm/Dockerfile).
- our custom image needs to be built as describe in [our Dockerfile](https://github.com/ajardin/demo/blob/1b0cc5a50d980029c20aa8f356f52b24f9f0e7a7/docker/php/Dockerfile).

The `php` service is built on top of an Alpine image which is the [smallest PHP image](https://github.com/docker-library/docs/blob/master/php/README.md#phpalpine) available on the Docker Store (~80MB). And with the addition of Symfony/Composer requirements, the result is an image of ~180MB.

Here is the list of PHP compiled modules.
```
[PHP Modules]
Core
ctype
curl
date
dom
fileinfo
filter
ftp
hash
iconv
intl
json
libxml
mbstring
mysqlnd
openssl
pcre
PDO
pdo_pgsql
pdo_sqlite
Phar
posix
readline
Reflection
session
SimpleXML
sodium
SPL
sqlite3
standard
tokenizer
xml
xmlreader
xmlwriter
Zend OPcache
zip
zlib

[Zend Modules]
Zend OPcache
```

Finally, I made some additional adjustments to reduce the installation time (from ~4:20m to ~2:35m) by removing the assets compilation (already present in the repository) and moving to the preview version of Composer (to take advantage of the latest improvements).

@maxhelias : I don't think a h2 proxy is required here, but I let @javiereguiluz answer. 😉

.env.dist
```diff
@@ -10,9 +14,9 @@ APP_SECRET=67d829bf61dc5f87a73fd814e2c9f629
 
 ###> doctrine/doctrine-bundle ###
 # Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url
-# For a sqlite database, use: "sqlite:///%kernel.project_dir%/var/data.db"
+# For a sqlite database, use: "sqlite:///%kernel.project_dir%/var/data.db?charset=utf8mb4"
```
stof: `utf8mb4` does not make sense for sqlite AFAIK. That's a MySQL thing

docker/php/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+        pdo_pgsql \
+        zip && \
+    apk del .build-deps
+
+# Install Composer globally
+ENV COMPOSER_ALLOW_SUPERUSER 1
+RUN \
+    curl -sS https://getcomposer.org/installer | php && \
+    mv composer.phar /usr/local/bin/composer && \
+    /usr/local/bin/composer self-update --preview
```
stof: all that can be simplify: `curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer --preview` (and this will avoid downloading the phar twice)

ajardin:
Do you have additional questions or concerns? If it's good enough for you, I'll start to write a small documentation.

javiereguiluz:
I'm sorry this PR stalled. I was waiting for more approvals from Docker experienced users. May I ask why did you close it? Note that we're still super interested in having Docker enabled for this repo. Thanks!

ajardin:
I closed it for several reasons. 😉 

- Because I thought you'll prefer to focus on SymfonyCloud for the local environment since the pull request hasn't received any feedback since September.
- I'm not satisfied with this implementation as the Docker environment will be tightly coupled with the application. I personally prefer an approach where the environment is retrieved like a dependency. 

Long story short, that's not a renouncement as I'm still looking for the best solution.

Issue #820, opened 2018-06-22
feature request


lxregistry:
hello,
could you add a function to edit user profile, change email, name, password, etc.

javiereguiluz:
Implemented in #834.

PR #834, opened 2018-07-17
Add ability to edit user's informations and password

ker0x:
As requested in #820

### DIFF
--- a/src/Controller/UserController.php
+++ b/src/Controller/UserController.php
@@ -23,7 +23,7 @@ use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
 /**
  * Controller used to manage current user.
  *
- * @Route("/user")
+ * @Route("/profile")
  * @Security("has_role('ROLE_USER')")
  *
  * @author Romain Monteil <monteil.romain@gmail.com>
@@ -31,9 +31,9 @@ use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
 class UserController extends AbstractController
 {
     /**
-     * @Route("/", methods={"GET", "POST"}, name="user_index")
+     * @Route("/edit", methods={"GET", "POST"}, name="user_edit")
      */
-    public function index(Request $request): Response
+    public function edit(Request $request): Response
     {
         $user = $this->getUser();
 
@@ -45,19 +45,19 @@ class UserController extends AbstractController
 
             $this->addFlash('success', 'user.updated_successfully');
 
-            return $this->redirectToRoute('user_index');
+            return $this->redirectToRoute('user_edit');
         }
 
-        return $this->render('user/index.html.twig', [
+        return $this->render('user/edit.html.twig', [
             'user' => $user,
             'form' => $form->createView(),
         ]);
     }
 
     /**
-     * @Route("/password", methods={"GET", "POST"}, name="user_password")
+     * @Route("/change-password", methods={"GET", "POST"}, name="user_change_password")
      */
-    public function password(Request $request, UserPasswordEncoderInterface $encoder): Response
+    public function changePassword(Request $request, UserPasswordEncoderInterface $encoder): Response
     {
         $user = $this->getUser();
 
@@ -67,13 +67,12 @@ class UserController extends AbstractController
         if ($form->isSubmitted() && $form->isValid()) {
             $user->setPassword($encoder->encodePassword($user, $form->get('newPassword')->getData()));
 
-            $em = $this->getDoctrine()->getManager();
-            $em->flush();
+            $this->getDoctrine()->getManager()->flush();
 
             return $this->redirectToRoute('security_logout');
         }
 
-        return $this->render('user/password.html.twig', [
+        return $this->render('user/change_password.html.twig', [
             'form' => $form->createView(),
         ]);
     }
--- a/src/Controller/UserController.php
+++ b/src/Controller/UserController.php
@@ -1,0 +1,80 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Controller;
+
+use App\Form\Type\ChangePasswordType;
+use App\Form\UserType;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
+
+/**
+ * Controller used to manage current user.
+ *
+ * @Route("/user")
+ * @Security("has_role('ROLE_USER')")
+ *
+ * @author Romain Monteil <monteil.romain@gmail.com>
+ */
+class UserController extends AbstractController
+{
+    /**
+     * @Route("/", methods={"GET", "POST"}, name="user_index")
+     */
+    public function index(Request $request): Response
+    {
+        $user = $this->getUser();
+
+        $form = $this->createForm(UserType::class, $user);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $this->getDoctrine()->getManager()->flush();
+
+            $this->addFlash('success', 'user.updated_successfully');
+
+            return $this->redirectToRoute('user_index');
+        }
+
+        return $this->render('user/index.html.twig', [
+            'user' => $user,
+            'form' => $form->createView(),
+        ]);
+    }
+
+    /**
+     * @Route("/password", methods={"GET", "POST"}, name="user_password")
+     */
+    public function password(Request $request, UserPasswordEncoderInterface $encoder): Response
+    {
+        $user = $this->getUser();
+
+        $form = $this->createForm(ChangePasswordType::class);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $user->setPassword($encoder->encodePassword($user, $form->get('newPassword')->getData()));
+
+            $em = $this->getDoctrine()->getManager();
+            $em->flush();
+
+            return $this->redirectToRoute('security_logout');
+        }
+
+        return $this->render('user/password.html.twig', [
+            'form' => $form->createView(),
+        ]);
+    }
+}
--- a/src/Form/Type/ChangePasswordType.php
+++ b/src/Form/Type/ChangePasswordType.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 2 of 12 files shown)


src/Controller/UserController.php
```diff
@@ -0,0 +1,90 @@
...
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $user->setPassword($encoder->encodePassword($user, $form->get('newPassword')->getData()));
+
+            $em = $this->getDoctrine()->getManager();
+            $em->persist($user);
+            $em->flush();
+
+            return $this->redirectToRoute('security_logout');
```
OlivierToussaint: I should remplace redirect to $this->redirectToRoute('user_index') and I add a flashbag message

translations/messages.en.xlf
```diff
@@ -169,6 +177,14 @@
...
                 <target>Browse backend</target>
             </trans-unit>
+            <trans-unit id="action.edit_user">
+                <source>action.edit_user</source>
+                <target>Edit user</target>
+            </trans-unit>
+            <trans-unit id="action.change_password">
+                <source>action.change_password</source>
+                <target>Change password</target>
+            </trans-unit>
```
voronkovich: The `action.update_user_info` is missing.

src/Form/Type/ChangePasswordType.php
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('currentPassword', PasswordType::class, [
+                'constraints' => [
+                    new NotBlank(),
+                    new UserPassword(),
```
yceruto: I think `NotBlank` constraint isn't necessary since a "Not blank" value is already verified in `UserPassword` constraint.

src/Form/Type/ChangePasswordType.php
```diff
@@ -0,0 +1,72 @@
...
+                    'autocomplete' => 'off',
+                ],
+            ])
+            ->add('newPassword', RepeatedType::class, [
+                'type' => PasswordType::class,
+                'constraints' => [
+                    new NotBlank(),
+                    new Length([
+                        'min' => 5,
+                    ]),
```
yceruto: I also suggest adding a `callback()` constraint and check by `BasePasswordEncoder::isPasswordTooLong($password)` otherwise we could get an uncatched error `BadCredentialsException` [here](https://github.com/symfony/demo/pull/834/files#diff-2e6355de36175249110ad2b48562407aR77).

src/Controller/UserController.php
```diff
@@ -0,0 +1,90 @@
...
+            'user' => $user,
+            'form' => $form->createView(),
+        ]);
+    }
+
+    /**
+     * @Route("/password", methods={"GET", "POST"}, name="user_password")
+     *
+     * @param \Symfony\Component\HttpFoundation\Request                             $request
+     * @param \Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface $encoder
```
yceruto: All these PHPDoc can be removed too.

src/Controller/UserController.php
```diff
@@ -0,0 +1,90 @@
...
+        $user = $this->getUser();
+
+        $form = $this->createForm(ChangePasswordType::class);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $user->setPassword($encoder->encodePassword($user, $form->get('newPassword')->getData()));
+
+            $em = $this->getDoctrine()->getManager();
+            $em->persist($user);
```
ogizanagi: `persist` is unnecessary here as the entity is not a new one and already managed by doctrine.

src/Form/Type/ChangePasswordType.php
ogizanagi: That's not something that would happen in a real-life scenario unless really trying to make it fail. No sensible user would try setting a 4096+ chars long password. So I would not showcase it here or instead arbitrary set a reasonable `max` value in `Length` constraint above.

src/Form/UserType.php
```diff
@@ -0,0 +1,64 @@
...
+
+        // By default, form fields include the 'required' attribute, which enables
+        // the client-side form validation. This means that you can't test the
+        // server-side validation errors from the browser. To temporarily disable
+        // this validation, set the 'required' attribute to 'false':
+        // $builder->add('title', null, ['required' => false, ...]);
+
+        $builder
+            ->add('username', TextType::class, [
+                'attr' => ['readonly' => true],
```
ogizanagi: setting the `readonly` attribute will not enforce this field cannot be submit and username not changed.
You must use [disabled](http://symfony.com/doc/master/reference/forms/types/form.html#disabled) instead.

src/Form/Type/ChangePasswordType.php
yceruto: It's [72 chars for `BCryptPasswordEncoder`](https://github.com/symfony/symfony/blob/ce73293532f6df48b2cac2aad8d040817429cea3/src/Symfony/Component/Security/Core/Encoder/BCryptPasswordEncoder.php#L22) which is the current configured encoder :P but you're right, setting it (`BCryptPasswordEncoder::MAX_PASSWORD_LENGTH`) in `max` `Length` constraint is simple.

src/Form/UserType.php
ker0x: If I use **disabled**, I got an error `Argument 1 passed to App\Entity\User::setUsername() must be of the type string, null given`.

So, what should I do ?
* leave readonly
* add the nullable type to `setUsername()` in `App\Entity\User`
* allow user to change his username
* remove the username input

translations/messages.en.xlf
ker0x: Actually, it was the translation key in the template file that was wrong.

src/Controller/UserController.php
ker0x: I find it safer to logout the user after he has changed his password, but I can be wrong. I would like to get feedback from the community about that.

src/Form/UserType.php
voronkovich: @ker0x, He means a form option, not an HTML attribute:
```php
$builder
    ->add('username', TextType::class,
        'label' => 'label.username',
        'disabled' => true,
    ])
```

src/Controller/UserController.php
```diff
@@ -0,0 +1,80 @@
...
+        return $this->render('user/index.html.twig', [
+            'user' => $user,
+            'form' => $form->createView(),
+        ]);
+    }
+
+    /**
+     * @Route("/password", methods={"GET", "POST"}, name="user_password")
+     */
+    public function password(Request $request, UserPasswordEncoderInterface $encoder): Response
```
ogizanagi: changePassword ? (same for the template name & route name)

src/Controller/UserController.php
```diff
@@ -0,0 +1,80 @@
...
+ * @Security("has_role('ROLE_USER')")
+ *
+ * @author Romain Monteil <monteil.romain@gmail.com>
+ */
+class UserController extends AbstractController
+{
+    /**
+     * @Route("/", methods={"GET", "POST"}, name="user_index")
+     */
+    public function index(Request $request): Response
```
ogizanagi: profile / editProfile ? (same for the template name & route name)

src/Form/UserType.php
```diff
@@ -0,0 +1,65 @@
...
+        // By default, form fields include the 'required' attribute, which enables
+        // the client-side form validation. This means that you can't test the
+        // server-side validation errors from the browser. To temporarily disable
+        // this validation, set the 'required' attribute to 'false':
+        // $builder->add('title', null, ['required' => false, ...]);
+
+        $builder
+            ->add('username', TextType::class, [
+                'label' => 'label.username',
+                'disabled' => true,
```
yceruto: If the intention is not to modify the username, how about removing this field and print `{{ user.username }}` directly in the template?

src/Controller/UserController.php
```diff
@@ -0,0 +1,80 @@
...
+        $user = $this->getUser();
+
+        $form = $this->createForm(ChangePasswordType::class);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $user->setPassword($encoder->encodePassword($user, $form->get('newPassword')->getData()));
+
+            $em = $this->getDoctrine()->getManager();
+            $em->flush();
```
yceruto: `$this->getDoctrine()->getManager()->flush()`?

src/Controller/UserController.php
```diff
@@ -0,0 +1,80 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
+
+/**
+ * Controller used to manage current user.
+ *
+ * @Route("/user")
```
yceruto: `/user-profile`, `/profile/user`, `user-account`, `account/user`... ?

src/Controller/UserController.php
ker0x: Yep, already change :+1:

templates/user/index.html.twig
```diff
@@ -0,0 +1,27 @@
...
+{% block body_id 'user_index' %}
+
+{% block main %}
+    <h1>{{ 'title.edit_user'|trans }}</h1>
+
+    {{ form_start(form, {attr: attr|default({})}) }}
+        {{ form_widget(form) }}
+
+        <button type="submit" class="{{ button_css|default("btn btn-primary") }}">
+            <i class="fa fa-save" aria-hidden="true"></i> {{ button_label|default('action.save'|trans) }}
```
yceruto: same here `{{ 'action.save'|trans }}`

templates/user/index.html.twig
```diff
@@ -0,0 +1,27 @@
...
+
+{% block body_id 'user_index' %}
+
+{% block main %}
+    <h1>{{ 'title.edit_user'|trans }}</h1>
+
+    {{ form_start(form, {attr: attr|default({})}) }}
+        {{ form_widget(form) }}
+
+        <button type="submit" class="{{ button_css|default("btn btn-primary") }}">
```
yceruto: why not `btn btn-primary` directly? same for `password.html.twig` template.

templates/user/index.html.twig
```diff
@@ -0,0 +1,27 @@
+{% extends 'base.html.twig' %}
+
+{% block body_id 'user_index' %}
+
+{% block main %}
+    <h1>{{ 'title.edit_user'|trans }}</h1>
+
+    {{ form_start(form, {attr: attr|default({})}) }}
```
yceruto: `{attr: attr|default({})}` can be removed.

src/Form/UserType.php
ker0x: As this is a demo application, maybe we can keep the disabled option to show people how to disable an input in a form ?

src/Controller/UserController.php
ker0x: I changed it for `/profile/edit`

src/Controller/UserController.php
ker0x: I changed it for `/profile/change-password`

sebastianblum:
@bobdenotter I think I can answer your question.
Root Types should be in `App\Form`
Child Types should be in `App\Form\Type`
so in my opinion this behavior is the current symfony best practise

javiereguiluz:
@ker0x you did it really great. This is a ver well done feature. Thanks for working on this and thanks for your patience during the review. And thanks to all reviewers too!

PR #837, opened 2018-07-19
Enabled the no_superfluous_phpdoc_tags fixer rule

javiereguiluz:
Our code was already good 😎 ... but this rule removed some unneeded PHPdoc.


tests/Form/DataTransformer/TagArrayToStringTransformerTest.php
```diff
@@ -121,8 +119,6 @@ private function getMockedTransformer(array $findByReturnValues = []): TagArrayT
      * This helper method creates a Tag instance for the given tag name.
      *
      * @param string $name
```
stof: should this be moved to a typehint too ?

PR #841, opened 2018-08-06
Fix installation when using composer 1.7+

fbourigault:
This avoid the following error while installing the demo application with composer 1.7.

```
Declaration of Symfony\Flex\ParallelDownloader::getRemoteContents($originUrl, $fileUrl, $context) should be compatible with Composer\Util\RemoteFilesystem::getRemoteContents
  ($originUrl, $fileUrl, $context, ?array &$responseHeaders = NULL)
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "sensiolabs/security-checker": "^4.1",
         "symfony/asset": "^4.1",
         "symfony/expression-language": "^4.1",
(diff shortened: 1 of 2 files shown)


fbourigault:
Travis error is unrelated (security checker failure).

javiereguiluz:
Thank you Fabien.

PR #845, opened 2018-08-12
Update dutch language file.

bobdenotter:
- Add some missing strings
 - Fix archaic and corny translations (formal "u" -> informal "je", etc)

PR #846, opened 2018-08-12
Updating `.lock` files for security reasons

bobdenotter:
Currently, `composer install` as well as Travis CI fails, because of symfony/http-foundation (v4.1.1): 

https://symfony.com/blog/cve-2018-14773-remove-support-for-legacy-and-risky-http-headers

screen shot 2018-08-12 at 15 14 37

This PR updates the `composer.lock` and `symfony.lock` files to use the latest Symfony 4.1.3. Should also fix Travis breakage on #845.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 7.1
+    - php: 7.1.18
     - php: 7.2
     - php: nightly
   allow_failures:
--- a/composer.lock
+++ b/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "composer/ca-bundle",
-            "version": "1.1.1",
+            "version": "1.1.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/composer/ca-bundle.git",
-                "reference": "d2c0a83b7533d6912e8d516756ebd34f893e9169"
+                "reference": "46afded9720f40b9dc63542af4e3e43a1177acb0"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/composer/ca-bundle/zipball/d2c0a83b7533d6912e8d516756ebd34f893e9169",
-                "reference": "d2c0a83b7533d6912e8d516756ebd34f893e9169",
+                "url": "https://api.github.com/repos/composer/ca-bundle/zipball/46afded9720f40b9dc63542af4e3e43a1177acb0",
+                "reference": "46afded9720f40b9dc63542af4e3e43a1177acb0",
                 "shasum": ""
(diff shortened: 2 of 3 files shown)


bobdenotter:
The build now breaks on this: 

```
PHPUnit 6.5.11 by Sebastian Bergmann and contributors.

Testing Project Test Suite
..................................................                50 / 50 (100%)

Time: 1.03 minutes, Memory: 56.25MB

OK (50 tests, 92 assertions)

Remaining deprecation notices (1)

  1x: Doctrine\Common\ClassLoader is deprecated.
    1x in BlogControllerTest::testAdminBackendHomePage from App\Tests\Controller\Admin
```

Which is something that (i think) needs to be fixed upstream: 

https://github.com/symfony/symfony/issues/28119

stof:
a solution to avoid this issue would be to update Symfony packages, but not Doctrine ones (until Doctrine ORM makes another release removing the usage of the deprecated API)

stof:
@bshaffer the demo app works. It reports being affected by a security vulnerability, but that's not "not working".
And this PR cannot be merged as is, as it turns the CI red.

bobdenotter:
@stof 

> a solution to avoid this issue would be to update Symfony packages, but not Doctrine ones

I've updated the PR like that, but now Travis passes on PHP 7.2, but not on PHP 7.1: 

```
Executing script security-checker security:check [KO]
 [KO]
Script security-checker security:check returned with error code 1
!!                                                                     
!!    An error occurred: gnutls_handshake() failed: Handshake failed.  
!!                                                                     
!!  
Script @auto-scripts was called via post-install-cmd
The command "composer install" failed and exited with 1 during .
Your build has been stopped.
```

Suggestions on how to proceed?

bshaffer:
I meant it doesn't work because it throws an error on `create-project`. You're right, if you ignore the error it works. But it breaks build scripts unless you suppress the error with ` || true`

bobdenotter:
Update: That took a little bit of trial and error, but the tests are green again. 

 - Updated `symfony/*` components to 4.1.3
 - Updated `composer/ca-bundle` to 1.1.2
 - Set Travis to use PHP 7.1.18, instead of 7.1.6 or something

See here for details on the TLS issue workaround: https://github.com/travis-ci/travis-ci/issues/6339

If/when Travis fixes this properly, i'll revert travis.yml to use plain `7.1` again.

javiereguiluz:
@bobdenotter I warmly appreciate your work to fix this issue and to investigate all these problems. Sorry for the late merge.

bobdenotter:
@javiereguiluz You're welcome! I noticed some breakage in the NPM/Encore worklfow as well. Now that this has been merged, expect a PR to straighten that one out, too.

PR #854, opened 2018-09-01
Composer update + leverage symfony/flex ~1.1

nicolas-grekas:


javiereguiluz:
Nicolas, I'm sorry but merging other pending PRs introduced conflicts in this one. Can you please take a look? Thanks!

bobdenotter:
This PR will also fail tests, because when the `doctrine/*` packages get updated, you'll bump into the same issue as I did in #846 :

screen shot 2018-08-12 at 15 14 37

I think we need to either selectively update `symfony/*` components, or lock the versions of `doctrine/*`, or do something similar as you did in https://github.com/symfony/symfony/pull/28038

stof:
The drawback of using `*` as Symfony constraint is that `composer validate` now complains about using unbounded constraint (and it is perfectly right about it).
Thus, any time you use `composer require` for new components, they would still get an actual constraint (unless you explicitly force `*`, but then you're responsible for the unbounded constraint).
I suggest keeping bound constraints.

PR #855, opened 2018-09-06
Updating composer deps, don't fail tests on vendor deprecations

bobdenotter:
Follows #846, replaces #854 

I've taken the liberty of updating @nicolas-grekas' PR to update flex to 1.1, whilst keeping bound constraints for the symfony components. 

Tests would fail on a deprecation in doctrine/common. I have also added a temporary `SYMFONY_DEPRECATIONS_HELPER=weak_vendors`, so it'll pass tests. After Symfony 4.2 has been released, I'll remove it.

### DIFF
--- a/bin/phpunit
+++ b/bin/phpunit
@@ -15,7 +15,8 @@ if (false === getenv('SYMFONY_PHPUNIT_DIR')) {
     putenv('SYMFONY_PHPUNIT_DIR='.__DIR__.'/.phpunit');
 }
 if (false === getenv('SYMFONY_DEPRECATIONS_HELPER')) {
-    putenv('SYMFONY_DEPRECATIONS_HELPER=weak_vendors');
+    // see https://symfony.com/doc/current/components/phpunit_bridge.html#making-tests-fail
+    putenv('SYMFONY_DEPRECATIONS_HELPER=999999');
 }
 
 require dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit';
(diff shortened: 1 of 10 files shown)


bin/phpunit
```diff
@@ -14,5 +14,8 @@ if (false === getenv('SYMFONY_PHPUNIT_VERSION')) {
 if (false === getenv('SYMFONY_PHPUNIT_DIR')) {
     putenv('SYMFONY_PHPUNIT_DIR='.__DIR__.'/.phpunit');
 }
+if (false === getenv('SYMFONY_DEPRECATIONS_HELPER')) {
+    putenv('SYMFONY_DEPRECATIONS_HELPER=weak_vendors');
```
nicolas-grekas: Weak_vendor is only for libraries. Let's borrow from https://github.com/symfony/recipes/pull/442/files instead

composer.json
```diff
@@ -15,7 +15,7 @@
         "sensiolabs/security-checker": "^4.1",
         "symfony/asset": "^4.1",
         "symfony/expression-language": "^4.1",
-        "symfony/flex": "^1.0.86",
+        "symfony/flex": "^1.1",
```
nicolas-grekas: You have to borrow also the extra config to benefit from v1.1

bobdenotter: Could you elaborate on this? Which extra config?

nicolas-grekas: https://github.com/symfony/demo/pull/854/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R83

bobdenotter: Thanks, both suggestions have been applied.

javiereguiluz:
Thanks Bob! We created #858 to not forget about tweaking this once Symfony 4.2 is released.

PR #856, opened 2018-09-06
Made some tweaks in the README

javiereguiluz:


README.md
```diff
@@ -24,16 +22,16 @@ Usage
 -----
 
 There's no need to configure anything to run the application. Just execute this
-command to run the built-in web server and access the application in your
+command to start the built-in web server and access the application in your
```
bobdenotter: If you change this to `start`,  should we also change to `server:start` a few lines below? This feels inconsistent.

PR #857, opened 2018-09-06
Updating Yarn dependencies, making build process work again

bobdenotter:
Some housekeeping on the Yarn / Encore build process: 

 - Updated `symfony/encore` to `0.20.1` from `0.9.1`
 - Update a few other deps
 - Build process is now working again
 - Changed file naming strategy

Note: I replaced this 

```
    .enableVersioning(false)
```

with this:

```
    .configureFilenames({
        js: '[name].js',
        images: 'images/[name].[ext]',
        fonts: 'fonts/[name].[ext]'
    })
```

I did this to keep the delta minimal. Since https://github.com/symfony/webpack-encore/pull/110 `.enableVersioning(false)` is not doing what it says (for understandable reasons), but in this case I chose to keep the filenames without a versioning hash. If you think we should introduce filenames with version hashes instead, i'll happily update this PR.


bobdenotter:
PR updated, Travis is green.

javiereguiluz:
@weaverryan quick ping: do you agree replacing `enableVersioning()` by `configureFilenames()` as a best practice in Webpack Encore? Thanks!

weaverryan:
Hi guys!

Hmm, I don't fully agree actually. If we were to have one "default" setting that we should encourage most people to use, it is to USE `enableVersioning()`. This is the default setting you get when you install the `encore-pack`, which works because the `json_manifest_path ` is setup for you to read the `manifest.json` file. So, it's free versioning support.

Configuring the filenames so that the images do not include a hash is something that I think only a minority of users want. If you're using the `json_manifest_path`, then the image filename is read from the `manifest.json` file and is transformed into the versioned filename.

So actually, I think we should go the other direction - *enable* versioning :).

Cheers!

weaverryan:
(but I like all the other changes - things do need to be updated, from time to time)

bobdenotter:
> I think we should go the other direction - enable versioning :).

That's fine with me, too. PR updated accordingly.

javiereguiluz:
Thanks Bob.

PR #859, opened 2018-09-29
Fix session starting issue on flash messages usage

andrerom:
Use of flash messages triggers session start, which as of last year
triggers Symfony to mark response as private, and not cached.

Fix the issue by checking if there is a session first as advised on:
https://symfony.com/doc/current/session/avoid_session_start.html

### DIFF
--- a/templates/default/_flash_messages.html.twig
+++ b/templates/default/_flash_messages.html.twig
@@ -5,19 +5,27 @@
    A common practice to better distinguish between templates and fragments is to
    prefix fragments with an underscore. That's why this template is called
    '_flash_messages.html.twig' instead of 'flash_messages.html.twig'
+
+   We check if we have session before reading flashes as it otherwise triggers session start:
+   https://symfony.com/doc/current/session/avoid_session_start.html
+
+   TIP: With FOSHttpCache you can also adapt this to make it cache safe:
+   https://foshttpcachebundle.readthedocs.io/en/latest/features/helpers/flash-message.html
 #}
 
-<div class="messages">
-    {% for type, messages in app.flashes %}
-        {% for message in messages %}
-            {# Bootstrap alert, see http://getbootstrap.com/components/#alerts #}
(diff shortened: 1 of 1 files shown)


sebastianblum:
In the symfony docs, both can be find:
What it the symfony best practise?

* `{% if app.request.hasPreviousSession %}`
* `{% if app.session.started %}`

https://symfony.com/blog/new-in-symfony-3-3-improved-flash-messages Normally symfony should do it automatically or?

andrerom:
@sebastianblum I think your question is more for @javiereguiluz or maybe rather @nicolas-grekas.

I just picked this from the doc, as part of my [talk on Symfony Live London](https://twitter.com/andrerom/status/1045806806298185729) on Http Caching, where I used symfony-demo as the base for showing of FosHttpCache usage _(with Varnish xkey)_. And symfony-demo just happens to follow what ever doc says is best practice.

But to be honest I agree, the framework should not trigger session start on things like this.

andrerom:
> The only way I can think of to fix this once for all is by using a separate request for flash messages (i.e. either XHR or IFRAME or ESI).

True, that would probably be cleaner then how FosHttpCache works around this:
https://foshttpcachebundle.readthedocs.io/en/latest/features/helpers/flash-message.html

javiereguiluz:
Thank you André.

nicolas-grekas:
@javiereguiluz sorry but this is wrong as explained...

javiereguiluz:
It looks correct according to Symfony Docs 😕  Please, someone send a new pull request to fix this merge.

But please keep in mind that this is the Symfony Demo app ... so we cant't complicate it with things that are only relevant for apps with lots of traffic. For example we can't display the flashes with a sub-request and enable Varnish, etc. Thanks!

andrerom:
Sure, but docs should probably be updated here, and demo should follow best practice in docs right?
In other words imho this merge is correct as long as doc is like it is :P

javiereguiluz:
Thanks Nico! See #863.

PR #861, opened 2018-09-30
Fix /var configuration in .gitignore

fabpot:
A better fix would probably be to move the DB elsewhere.

But at least, this change is required to make the demo work out of the box. If not, the `var/` directory is not added when doing `git add .` after creating the project with `composer create-project`, which is not what we want.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,6 @@ yarn-error.log
 ###> symfony/framework-bundle ###
 /.env
 /public/bundles/
-/var/
 /vendor/
 ###< symfony/framework-bundle ###
 


.gitignore
```diff
@@ -20,7 +20,6 @@ yarn-error.log
 ###> symfony/framework-bundle ###
 /.env
 /public/bundles/
-/var/
```
fabpot: `/var/*` is already excluded a few lines before (but with some other rules afterwards). This one overrides the work done above.

javiereguiluz:
Thanks Fabien.

PR #863, opened 2018-10-04
Fixed the handling of flash messages

javiereguiluz:
Fixes the #859 merge.

### DIFF
--- a/templates/default/_flash_messages.html.twig
+++ b/templates/default/_flash_messages.html.twig
@@ -5,15 +5,16 @@
    A common practice to better distinguish between templates and fragments is to
    prefix fragments with an underscore. That's why this template is called
    '_flash_messages.html.twig' instead of 'flash_messages.html.twig'
+#}
 
-   We check if we have session before reading flashes as it otherwise triggers session start:
+{#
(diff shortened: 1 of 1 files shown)


xabbuh:
I do not understand this change. I mean the flash entries will probably be created when processing a POST request, but the reason we register them is that we redirect at the end and we need to display them in response for the subsequent event which will be a GET request then. So with this change the flash messages are likely to never be shown.

sebastianblum:
Hello @javiereguiluz and @xabbuh 

This fix does only work with 307 redirects, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307

Submit form: POST /url - store flash message and redirect with 307 to /thank-you
307 redirect: POST /thank-you - shows the flash message 

With an 302 redirect, the redirect will switch to an GET-Request and so the flash message will be hidden.
Submit form: POST /url - store flash message and redirect with 302 to /thank-you
302 redirect: GET /thank-you - doesn't show the flash message

stof:
Posting to a thank-you page looks wrong to me btw (that page should probably not even accept POST requests)

PR #871, opened 2018-10-06
Change null !== to is_null function

diegograssato:


bocharsky-bw:
Hey @diegograssato, could you provide a bit more information... what is the reason behind these changes? Is it faster this way, or something else? Otherwise, I don't see the value in this PR. Someone else later can open a PR that will do exactly the reverse.

Cheers!

diegograssato:
Ok, I'll close this PR

PR #874, opened 2018-10-11
Replaced the "transchoice" filter by "trans"

### LINKED ISSUE #872: Twig Filter "transchoice" is deprecated since version 4.2
Twig Filter "transchoice" is deprecated since version 4.2. Use "trans" with parameter "%count%" instead in /var/www/vhosts/symfony/templates/blog/post_show.html.twig at line 37.

javiereguiluz:
Fixes #872.

### DIFF
--- a/translations/messages.ru.xlf
+++ b/translations/messages.ru.xlf
@@ -266,7 +266,7 @@
             </trans-unit>
             <trans-unit id="post.num_comments">
                 <source>post.num_comments</source>
-                <target>%count% комментарий|%count% комментария|%count% комментариев</target>
+                <target>{count, plural, one {# комментарий} many {# комментария} other {# комментариев}}</target>
             </trans-unit>
             <trans-unit id="post.commented_on">
                 <source>post.commented_on</source>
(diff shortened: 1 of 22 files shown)


stof:
you also need to migrate to the ICU pluralization format in messages. This is *not* finished.

translations/messages.ru.xlf
```diff
@@ -266,7 +266,7 @@
             </trans-unit>
             <trans-unit id="post.num_comments">
                 <source>post.num_comments</source>
-                <target>%count% комментарий|%count% комментария|%count% комментариев</target>
+                <target>{count, plural, one {# комментарий} many {# комментария} other {# комментариев}}</target>
```
stof: The 3 Russian variants are `one`, `few` and `many`: http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#ru-comp

(and yes, using `other` would have been better than `many`, but they cannot fix it anymore for BC reasons: http://cldr.unicode.org/index/cldr-spec/plural-rules#TOC-Choosing-Plural-Category-Names)

javiereguiluz:
@stof thanks for the review and the fixes. I made some changes. Hopefully this is ready now.

PR #878, opened 2018-10-20
Fixed exemple in alternative use of isGranted annotation

mickaelandrieu:
Not really important, but replacing one but the alternative don't produce the same result :+1:


javiereguiluz:
Great attention to detail! I love it. Thanks Mickaël.

PR #882, opened 2018-11-01
Delete duplicate sensio_framework_extra config

Tobion:
It exists already as https://github.com/symfony/demo/blob/bbe5180a8c3b6886ea59227df83de59a4573e586/config/packages/sensio_framework_extra.yaml


javiereguiluz:
Tobias, thanks a lot for this and all the other fixes you sent for the Symfony Demo app.

Issue #889, opened 2018-11-06
Translation broke after upgrading to Symfony 4.2 beta1


javiereguiluz:
Translated messages using the new format no longer work:

![i18n-error](https://user-images.githubusercontent.com/73419/48053722-ae9d4600-e1ab-11e8-9e25-0faf84a975e4.png)

This is probably because we need to add the suffix in the translation messages to enable the new format explicitly.

ogizanagi:
I think you just need to rename translations files with the `+intl-icu` suffix:

```diff
- messages.en.xlf
+ messages+intl-icu.en.xlf
```

(https://github.com/symfony/symfony/pull/28952)

gmsantos:
@javiereguiluz I will work on that

javiereguiluz:
@gmsantos great! Thanks a lot.

javiereguiluz:
Fixed by #890.

PR #890, opened 2018-11-06
Sufix translations files with +intl-icu

gmsantos:
~~Also added intl extension to composer require because when I was testing on my machine, this extension was missing.~~

Reverting intl extension require in favor of #895

Maybe we could require [`polyfill-intl-normalizer`](https://github.com/symfony/polyfill-intl-normalizer) instead of the extension.


gmsantos:
resolves #889

javiereguiluz:
Thanks for fixing this bug Gabriel.

PR #893, opened 2018-11-10
removed md2html from summary on blog index template

hapheus:
The description says that neither HTML nor Markdown is allowed. This is also the only place where markdown is interpreted on the summary.


javiereguiluz:
@hapheus good catch! We've merged this ... but tweaked it a bit in 412c79c53b041cffff8ec05485edfb4140f99a20 to make it look better. Thanks!

PR #895, opened 2018-11-12
Require php-intl instead of crashing on IntlTimeZone

aurelijusb:
PHP 7.2 on Ubuntu does not have `php7.2-intl` installed by default.
With `composer install` there are no check for **`intl`** extension
and first page shows misleading error:

Attempted to load class "IntlTimeZone" from the global namespace. Did you forget a "use" statement?

It is better to show warning during `composer install`
(or composer create-project):

Your requirements could not be resolved to an installable set of packages. Problem 1 - The requested PHP extension ext-intl * is missing from your system. Install or enable PHP's intl extension.

So it would be **easier to google**, that the original problem could be solved with:

```bash
sudo apt-get install php7.2-intl
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,6 +7,7 @@
     "require": {
         "php": "^7.1.3",
         "ext-pdo_sqlite": "*",
+        "ext-intl": "*",
         "doctrine/doctrine-bundle": "^1.6.10",
         "doctrine/doctrine-migrations-bundle": "^1.3",
         "doctrine/orm": "^2.5.11",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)


stof:
@javiereguiluz as discussed in the issue, that might not be the preferred way there.

javiereguiluz:
@stof I'm afraid I don't understand your comment. Do you like the change proposed in this pull request?

stof:
@javiereguiluz in the issue, I suggested that this issue should be fixed in Twig-extensions instead, to keep it compatible with the polyfill, instead of requiring the extension in the demo.

bocharsky-bw:
Does this issue still exist for the latest version of the project on Ubuntu?

aurelijusb:
Seems to be solved :slightly_smiling_face: Thanks. Closing.

> Tested with:
> Ubuntu 20.04.2 LTS
> PHP 7.4.3
> Symfony 5.3.1
> Symfony CLI version v4.25.4

bocharsky-bw:
Thank you for reporting it! ❤️

PR #898, opened 2018-11-19
Refactor security_checker.yaml for using autowire and autoconfigure options.

atamansv:
- Enabled autowire and autoconfigure in _default section and refactor for using it
- Removed parameter public: false - false is default value
- Removed config/packages/dev/security_checker.yaml - no longer needed

We can also discuss about move it to config/services.yaml as it already use autowiring


javiereguiluz:
Sergey, we appreciate your contribution. However, for the `config/packages/*` files we always copy+paste the official recipes (and update those files when the related recipes are updated).

In https://github.com/symfony/recipes/tree/master/sensiolabs/security-checker/4.0/config/packages I see that this recipe has indeed been updated so we must do some changes in our app. But this is different than your changes ... so could you please update this pull request to use the same config as the official recipe? Thanks a lot!

atamansv:
If I use same cofig it's just revert all changes)
Except deleting dev/security_checker.yaml - don't understand why it exists.

So maybe it's better to move this PR to `symfony/recipes` repository?

javiereguiluz:
Yes, I think it's better to propose these changes to the official recipe and if they are accepted, then make those changes here too. Thanks!

PR #899, opened 2018-11-20
Improving entities contracts

yceruto:
For example:
```php
$post = new Post();

// ...

if (!$post->getAuthor()) { // it's null initially so it should be fine allow null return
    // ...
} 

// ...

$post->setAuthor(null);
// shouldn't be allowed since the column is not null -> on flush() it'll fail anyway
```
Also it's consistent with `make:entity` command.

### DIFF
--- a/src/Entity/Comment.php
+++ b/src/Entity/Comment.php
@@ -90,7 +90,7 @@ class Comment
         return !$containsInvalidCharacters;
     }
 
-    public function getId(): int
+    public function getId(): ?int
     {
         return $this->id;
     }
@@ -115,7 +115,7 @@ class Comment
         $this->publishedAt = $publishedAt;
     }
 
-    public function getAuthor(): User
+    public function getAuthor(): ?User
     {
         return $this->author;
     }
@@ -130,7 +130,7 @@ class Comment
         return $this->post;
     }
 
-    public function setPost(?Post $post): void
+    public function setPost(Post $post): void
     {
         $this->post = $post;
     }
--- a/src/Entity/Post.php
+++ b/src/Entity/Post.php
@@ -128,7 +128,7 @@ class Post
         $this->tags = new ArrayCollection();
     }
 
-    public function getId(): int
+    public function getId(): ?int
     {
         return $this->id;
     }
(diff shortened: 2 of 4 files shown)


src/Entity/Comment.php
```diff
@@ -130,7 +130,7 @@ public function getPost(): ?Post
         return $this->post;
     }
 
-    public function setPost(?Post $post): void
+    public function setPost(Post $post): void
```
stof: that breaks `Post::removeComment` (but the proper way there is not to set it as `null` here anyway, as the Comment will be removed entirely by the `orphanRemoval` feature, not be kept orphan)

yceruto: Thanks @stof, done.

yceruto:
@javiereguiluz what do you think about this proposal?

src/Entity/Comment.php
```diff
@@ -105,7 +105,7 @@ public function setContent(string $content): void
         $this->content = $content;
     }
 
-    public function getPublishedAt(): \DateTime
+    public function getPublishedAt(): ?\DateTime
```
stof: not needed, as the field is not nullable in the DB and the constructor initializes it

src/Entity/Post.php
```diff
@@ -158,27 +158,27 @@ public function getContent(): ?string
...
     }
 
-    public function setContent(?string $content): void
+    public function setContent(string $content): void
     {
         $this->content = $content;
     }
 
-    public function getPublishedAt(): \DateTime
+    public function getPublishedAt(): ?\DateTime
```
stof: same here

yceruto:
@stof comments addressed, thanks.

src/Entity/Post.php
```diff
@@ -198,7 +198,6 @@ public function addComment(?Comment $comment): void
 
     public function removeComment(Comment $comment): void
     {
-        $comment->setPost(null);
```
voronkovich: @yceruto, Why did you remove this line?

yceruto: See https://github.com/symfony/demo/pull/899#discussion_r235046319

voronkovich: @yceruto, Ok, I got it.

javiereguiluz:
Thank you Yonel.

PR #902, opened 2018-11-26
update security check endpoint to security.symfony.com

bobdenotter:
As suggested by @fabpot, here: https://twitter.com/fabpot/status/1067085006542118913 😇

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ script:
   # this checks that the XLIFF translations contain no syntax errors
   - ./bin/console lint:xliff translations
   # this checks that the application doesn't use dependencies with known security vulnerabilities
-  - ./bin/console security:check --end-point=http://security.sensiolabs.org/check_lock
+  - ./bin/console security:check --end-point=http://security.symfony.com/check_lock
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -38,7 +38,7 @@ script:
   # this checks that the XLIFF translations contain no syntax errors
   - ./bin/console lint:xliff translations
   # this checks that the application doesn't use dependencies with known security vulnerabilities
-  - ./bin/console security:check --end-point=http://security.sensiolabs.org/check_lock
+  - ./bin/console security:check --end-point=http://security.symfony.com/check_lock
```
fabpot:
```suggestion
  - ./bin/console security:check
```

bobdenotter: I wondered about that too. The _default_ is actually using https, instead of http: 

https://github.com/sensiolabs/security-checker/blob/master/SensioLabs/Security/Crawler.php#L23

Since this is in `.travis.yml`, we can assume there's no need to fall back to plain http.. PR updated!

javiereguiluz:
Thanks Bob.

PR #906, opened 2018-11-30
Disable HTTP caching for the pages where flashes are being shown

voronkovich:
For now demo app shows flashes only when an HTTP method POST is used to avoid page caching. Maybe we could fix it by adding [no-store](https://tools.ietf.org/html/rfc7234#section-5.2.1.5) directive for all pages where the flashes are being shown?

### DIFF
--- a/src/EventSubscriber/DisableHttpCachingSubscriber.php
+++ b/src/EventSubscriber/DisableHttpCachingSubscriber.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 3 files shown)


javiereguiluz:
@voronkovich I appreciate your contribution a lot ... but I'm not sure about merging it in the Symfony Demo app.

This app is extremely important for Symfony, because it gives a first impression to newcomers and people evaluating Symfony. Now, dealing with "flash messages" should be trivial in a web framework. It's such a simple, limited and trivial feature, that it cannot be complicated in any way. It should work perfectly without any effort on the developer.

That's why "we can't" merge this feature. It gives the wrong impression that Symfony is overengineered, verbose and complicated. But of course you are right that we must solve the current problem ... but in a super simple and alternative way, please 🙏 

Thank you for understanding this decision and for all your contributions to this repo. We warmly appreciate them ❤️

PR #908, opened 2018-12-04
Fix translations files with ICU message formater

ker0x:
As pointed by @stof in #890, translations files does no longer take in account variables because of the `+intl-icu` suffix in translation files. This PR fix that!

### DIFF
--- a/translations/messages+intl-icu.bg.xlf
+++ b/translations/messages+intl-icu.bg.xlf
@@ -20,15 +20,15 @@
             </trans-unit>
             <trans-unit id="http_error.name">
                 <source>http_error.name</source>
-                <target>Грешка %status_code%</target>
+                <target>Грешка {status_code, number}</target>
             </trans-unit>
             <trans-unit id="http_error.description">
                 <source>http_error.description</source>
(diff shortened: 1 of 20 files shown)


translations/messages+intl-icu.bg.xlf
```diff
@@ -20,15 +20,15 @@
             </trans-unit>
             <trans-unit id="http_error.name">
                 <source>http_error.name</source>
-                <target>Грешка %status_code%</target>
+                <target>Грешка {status_code, number}</target>
```
ker0x: I'm not sure if the type is needed here! I added it by principe but feedback will be appreciated

translations/messages+intl-icu.bg.xlf
```diff
@@ -80,7 +80,7 @@
             </trans-unit>
             <trans-unit id="title.edit_post">
                 <source>title.edit_post</source>
-                <target>Редактиране на публикация #%id%</target>
+                <target>Редактиране на публикация #{id, number}</target>
```
ker0x: Here too, not sure if needed !

javiereguiluz:
Thanks for fixing this bug Romain.

PR #915, opened 2018-12-11
Use bootstrap.php for tests and environment-specific settings

yceruto:
Latest changes according to:
 * https://symfony.com/doc/current/configuration/dot-env-changes.html
 * https://symfony.com/blog/new-in-symfony-4-2-define-env-vars-per-environment
 * https://symfony.com/blog/improvements-to-the-handling-of-env-files-for-all-symfony-versions
 * and [phpunit recipe](https://github.com/symfony/recipes/blob/6c6ac5bf41c4952e1eeb2bdd2644ca2b4d8ba594/phpunit/phpunit/4.7/phpunit.xml.dist#L8)

### DIFF
--- a/.env
+++ b/.env
@@ -2,6 +2,8 @@
 # Set variables here that may be different on each deployment target of the app, e.g. development, staging, production.
 # https://symfony.com/doc/current/best_practices/configuration.html#infrastructure-related-configuration
 
+DEFAULT_FORMAT=xml
+
 ###> symfony/framework-bundle ###
 APP_ENV=dev
 APP_DEBUG=1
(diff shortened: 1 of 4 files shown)


.env
```diff
@@ -2,6 +2,8 @@
 # Set variables here that may be different on each deployment target of the app, e.g. development, staging, production.
 # https://symfony.com/doc/current/best_practices/configuration.html#infrastructure-related-configuration
 
+DEFAULT_FORMAT=xml
```
javiereguiluz: Just asking: what does this option define? Which default format will be XML? Thanks.

yceruto: Not sure why this is here... :) my bad.

javiereguiluz:
Very nice changes as always! We now follow all the latest Symfony best practices. Thanks Yonel!

PR #920, opened 2018-12-22
Encapsulate user creation with a factory

voronkovich:
I found that we use the one and the same piece of code in two places to create User instances. I suggest to encapsulate those code with a factory.


seb-jean:
Maybe use https://github.com/zenstruck/foundry

setineos:
Any news?

Issue #910, opened 2018-12-05
Tricky error related to 500 error pages


javiereguiluz:
Yesterday I merged #909 as an (ugly) temporary solution to a very tricky error.

How to reproduce the error:

* Use `prod` and `debug=0` in `.env`
* Comment the `DATABASE_URL` config to trigger a 500 error
* The 500 error page sometimes is styled with CSS and sometimes isn't:
  * The 500 error for this page shows styles: `/en/blog/posts/lorem-ipsum-dolor-sit-amet-consectetur-adipiscing-elit`
  * The 500 error for this page doesn't: `/en/blog/`

Can anyone spot the reason and provide a pull request to fix it? Thanks!

ckrack:
The reason is, that webpack-encore-bundle's `EntrypointLookup` already has the css file in the returnedFiles in this line:
https://github.com/symfony/webpack-encore-bundle/blob/a3382c840f74db83c3df42cd811e06934c422bd9/src/Asset/EntrypointLookup.php#L65

ckrack:
```diff
// webpack-encore-bundle/src/Asset/EntrypointLookup.php
    public function getCssFiles(string $entryName): array
    {
+       dump('returned files:', $this->returnedFiles);
        return $this->getEntryFiles($entryName, 'css');
    }
// ...
    private function getEntryFiles(string $entryName, string $key): array
    {
        $this->validateEntryName($entryName);
        $entriesData = $this->getEntriesData();
        $entryData = $entriesData['entrypoints'][$entryName];
        if (!isset($entryData[$key])) {
            // If we don't find the file type then just send back nothing.
            return [];
        }

        // make sure to not return the same file multiple times
        $entryFiles = $entryData[$key];
+       dump('entryfiles:', $entryFiles);
+       dump('diff entry and existing returned', array_diff($entryFiles, $this->returnedFiles));
        $newFiles = array_values(array_diff($entryFiles, $this->returnedFiles));
        $this->returnedFiles = array_merge($this->returnedFiles, $newFiles);
+       dump('wrote returned files');
+       dump('newfiles', $newFiles);
        return $newFiles;
    }
```

Output from calling `/en/blog/`
```
"returned files:"
array:1 [
  0 => "/build/css/app.4aa95248.css"
]
"entryfiles:"
array:1 [
  0 => "/build/css/app.4aa95248.css"
]
"diff entry and existing returned"
[]
"wrote returned files"
"newfiles"
[]
```

[...]

PR #923, opened 2018-12-28
Update error500.html.twig

20uf:
Useless, already included with ```{{encore_entry_link_tags ('css / app')}}```


javiereguiluz:
@20uf thanks for this contribution! In normal circumstances you would be right and this would be a mistake ... however, this is a temporary ugly solution to fix a very tricky error related to Encore. It's explained here: https://github.com/symfony/demo/issues/910

This is not fixed yet, so I'm afraid we can't merge this yet and that's why I'm closing it.

PR #927, opened 2019-01-09
Fixed transitive dependencies usage

javiereguiluz:
I was playing with the "transitive dependencies usage" inspection and lots of errors were reported. Our code uses some classes that belong to packages not required explicitly in our `composer.json`.

For example, we have lots of console commands with lots of `use Symfony\Component\Console\...` import but we never required `symfony/console` ... so it worked thanks to some package requiring symfony/console indirectly.

Same happens to lots of other missing dependencies. However, should we add all of them?

* These seem obvious to me that we should add them: Console, Intl, OptionsResolver.
* These could be added or not: Twig (added indirectly by TwigBundle), all the security/* micro-dependencies (added indirectly by SecurityBundle), EventDispatcher, Doctrine Common, Doctrine Persistence, etc.
* I think we should not add these: HttpFoundation, HttpKernel, Routing, etc.

What do you think?


composer.json
```diff
@@ -14,20 +14,25 @@
...
         "sensiolabs/security-checker": "^5.0",
         "symfony/asset": "*",
+        "symfony/console": "*",
         "symfony/expression-language": "*",
         "symfony/flex": "^1.1",
         "symfony/form": "*",
         "symfony/framework-bundle": "*",
+        "symfony/intl": "*",
         "symfony/monolog-bundle": "^3.1",
+        "symfony/options-resolver": "*",
```
yceruto: About `OptionsResolver` it's a hard dep for `symfony/form` (which is already required), so there is no effect.

voronkovich:
I agree with @yceruto, we don't need this changes

javiereguiluz:
@voronkovich I agree that we should not add most of these deprecations ... but what about `symfony/console` and `symfony/intl` which are imported explicitly from our code but never included in `composer.json` ?

voronkovich:
@javiereguiluz, to be honest, I have no strong opinion about the matter. Personally, I never worry about transitive dependencies. BTW, there is a tool that can be helpful to check such dependencies: [ComposerRequireChecker](https://github.com/maglnet/ComposerRequireChecker).

javiereguiluz:
Let me show the most clear example in my opinion: `symfony/console`. Our app absolutely depends on it because we have several commands created with it. But we don't require it; we're just lucky that these packages require it indirectly:

```cli
$ composer why symfony/console

doctrine/doctrine-bundle     1.10.1   requires  symfony/console (~2.7|~3.0|~4.0)
doctrine/migrations          v1.8.1   requires  symfony/console (~3.3|^4.0)
doctrine/orm                 v2.6.3   requires  symfony/console (~3.0|~4.0)
friendsofphp/php-cs-fixer    v2.14.0  requires  symfony/console (^3.4.17 || ^4.1.6)
sensiolabs/security-checker  v5.0.3   requires  symfony/console (~2.7|~3.0|~4.0)
symfony/web-server-bundle    v4.2.2   requires  symfony/console (~3.4|~4.0)
```

This one looks "wrong" to me, so I'm in favor of adding symfony/console. We just need to get the list of must-have dependencies according to what we do in our code.

stof:
yeah, that one looks wrong indeed.

OptionsResolver is fine however: we get it because Form depends on it. But we only use it in form types AFAIK. So removing forms would also remove the need for OptionsResolver.

PR #929, opened 2019-01-12
Fix dynamic translation with ICU

matthieumota:
I've run a simple Symfony demo and if you try to edit a post with 5 tags, you get an error due to [MessageFormatter](http://php.net/manual/fr/class.messageformatter.php) from icu used in [IntlFormatter](https://github.com/symfony/translation/blob/master/Formatter/IntlFormatter.php). It's due to ```{{ limit }}``` in message.

No documentation about that but we can see in a [test](https://github.com/symfony/translation/blob/master/Tests/Formatter/IntlFormatterTest.php#L88) that ```{ variable }``` can be processed. So we must use ```{ limit }``` and no ```{{ limit }}``` in message translate with icu.

I think to submit a PR to translation component for we can use ```{{ limit }}``` or ```{ limit }``` syntax.

### DIFF
--- a/translations/validators+intl-icu.bg.xlf
+++ b/translations/validators+intl-icu.bg.xlf
@@ -12,11 +12,11 @@
             </trans-unit>
             <trans-unit id="post.too_short_content">
                 <source>post.too_short_content</source>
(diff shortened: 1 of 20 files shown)


javiereguiluz:
Thanks for fixing this bug Matthieu.

PR #931, opened 2019-01-13
Implement docker

mpiot:
This PR provides a Docker implementation for the Symfony Demo project:

- [X] Main multistage Dockerfile for the app
- [X] Dockerfiles for: Nginx and MySQL
- [X] docker-compose.yml (with target usage) and docker-compose.override.yml.dist
- [X] Makefile for: install, start, stop, db interraction, assets, etc...
- [X] Update doctrine.yaml config
- [X] Add a Doctrine migration
- [X] Remove /public/build, /var from Git
- [ ] Update the README.md
- [ ] Add a part that explain how to build a prod image ? Like `docker build --target app-prod -t symfony-demo:0.1.0 .`


mpiot:
@javiereguiluz okay, I see many possibilities:
 - implement MySQL and Yarn in docker (add SQLite to docker), keep build assets and the SQLite db
 - replace MySQL by SQLite
 - replace MySQL by SQLite, and remove yarn

I think the first proposal is better, then I also can add a comment in docker near the PDO extension ?

javiereguiluz:
@mpiot I'm afraid I can't help you because I don't know much about Docker. But others will read this and will help you.

In any case, the basic premise is: we can do anything inside Docker ... but we can't change anything in the existing app.

mpiot:
@javiereguiluz I've fix it :-)
I've just keep a Doctrine Migration file, then if a user want to use MySQL he can execute the migration.

javiereguiluz:
Thanks a lot for reworking this. To me it's OK because it doesn't change the existing app ... but I 'll let Docker experts review it before merging. Thanks.

docker/NginxDockerfile
```diff
@@ -0,0 +1,41 @@
...
+        echo '        fastcgi_param SCRIPT_FILENAME $realpath_root$fastcgi_script_name;'; \
+        echo '        fastcgi_param DOCUMENT_ROOT $realpath_root;'; \
+        echo '        internal;'; \
+        echo '    }'; \
+        echo ''; \
+        echo '    location ~ \.php$ {'; \
+        echo '        return 404;'; \
+        echo '    }'; \
+        echo '}'; \
+    } > /etc/nginx/conf.d/default.conf
```
stof: I suggest to have a `default.conf`  file in this `docker` and to add it instead. The nginx config file split into `echo` commands for each line is not really readable. Having a config file would be easier to maintain IMO.

mpiot: @stof I agrre with you, before I did it like that. I don't remember why I've change it to use this echo...

Makefile
```diff
@@ -0,0 +1,165 @@
...
+
+
+##
+## Database
+##---------------------------------------------------------------------------
+
+wait-for-db:
+	$(EXEC) php -r "set_time_limit(60);for(;;){if(@fsockopen('db',3306)){break;}echo \"Waiting for MySQL\n\";sleep(1);}"
+
+db-diff: vendor wait-for-db                                                                            ## Generate a migration by comparing your current database to your mapping information
```
voronkovich: So, If we don't use the MySql all these command will wait 60 seconds?

docker/nginx-default.conf
```diff
@@ -0,0 +1,35 @@
+ server {
+    listen 80;
+    server_name _;
+    root /app/public;
+
+    add_header X-Content-Type-Options nosniff;
+    add_header X-XSS-Protection "1; mode=block";
+    add_header X-Frame-Options SAMEORIGIN;
+
+    if ($http_user_agent ~* "WordPress") {
```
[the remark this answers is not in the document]
mpiot: Mmm, it's about security, but useless in our case too, I've really forget a lot of parts... :-(

docker/nginx-default.conf
```diff
@@ -0,0 +1,35 @@
...
+        return 403;
+    }
+
+    location / {
+        try_files $uri /index.php$is_args$args;
+    }
+
+    location /protected-files/ {
```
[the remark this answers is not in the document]
mpiot: It's to store protected files and use X-Accel-Redirect, idem, forget...

Dockerfile
```diff
@@ -0,0 +1,181 @@
...
+##               APP               ##
+#####################################
+FROM php:${PHP_VERSION}-fpm as app
+
+ARG ICU_VERSION
+ARG APCU_VERSION
+
+# Used for the ICU compilation
+ENV PHP_CPPFLAGS="${PHP_CPPFLAGS} -std=c++11"
+ENV APP_VERSION=0.0.0
```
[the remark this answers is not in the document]
mpiot: I agree, it's interesting :-) I do it ;-)

Makefile
```diff
@@ -0,0 +1,165 @@
...
+clear-cache: perm
+	$(EXEC) $(CONSOLE) cache:clear --no-warmup
+	$(EXEC) $(CONSOLE) cache:warmup
+
+tty:                                                                                                   ## Run app container in interactive mode
+	$(EXEC) /bin/bash
+
+clear: perm                                                                                            ## Remove all the cache, the logs, the sessions and the built assets
+	$(EXEC) rm -rf var/cache/*
+	$(EXEC) $(CONSOLE) redis:flushall -n
```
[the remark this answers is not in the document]
mpiot: Idem...

Makefile
mpiot: Yes, in fact... I can remove all the db part, or just remove the waiting part and the db-reset part.

Makefile
```diff
@@ -0,0 +1,165 @@
...
+uninstall: stop                                                                                        ## Remove docker containers
+	$(DOCKER_COMPOSE) rm -vf
+
+reset: uninstall install                                                                               ## Remove and re-create docker containers
+
+clear-cache: perm
+	$(EXEC) $(CONSOLE) cache:clear --no-warmup
+	$(EXEC) $(CONSOLE) cache:warmup
+
+tty:                                                                                                   ## Run app container in interactive mode
```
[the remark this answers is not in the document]
mpiot: This part open a terminal in the container, then I've named it tty:
https://en.wikipedia.org/wiki/Computer_terminal#Text_terminals

I think sh is not really correct because there is a difference between /bin/sh and /bin/bash

mpiot:
@voronkovich I've fixed some parts, as mentioned above, I've forget some parts (I use files from a personnal project).

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+            context: .
+            target: app-dev
+        networks:
+            - frontend
+        volumes:
+            - .:/app
+
+
+networks:
+    frontend:
```
voronkovich: This network is useless, we already have a `default` one. https://docs.docker.com/compose/networking/#configure-the-default-network

Makefile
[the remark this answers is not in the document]
voronkovich: So, maybe a `bash` would be more appropriate name? BTW, I think you forgot to add a `-it` flags to allocate tty.

mpiot:
@voronkovich In the makefile the `make tty` use `docker-compose exec` then we don't need to add `-it` options.

I wait some other comments for this part, `make tty`, `make bash` or `make terminal` ?

Makefile
jkufner: This is actually quite nice oneliner. Wait until services are ready is very useful, so I would like to see something like this in the makefile, even if commented out as we use sqlite here.

mpiot: Yes, but it's difficult to add it in the Makefile whithout MySQL... Basically I've add all the Makefile db part in the case the user would like to use MySQL, but then we've removed all MySQL trace...

mpiot:
@jkufner why not, now: 3 persons => 3 choices :-)

voronkovich:
`make shell` sounds nice to me too.

mpiot:
THis is ok, I've rename the `tty` part in `shell` :-)

mpiot:
What do you think about the naming for the `php-cs-fixer` part ? Because we name `lint-yaml`, `lint twig`, `lint xliff`, why not `lint php` ?

I'm pretty divided about it.

Dockerfile
```diff
@@ -0,0 +1,181 @@
...
+        echo 'opcache.max_accelerated_files = 20011'; \
+        echo 'opcache.fast_shutdown = 1'; \
+        echo 'realpath_cache_size = 4096K'; \
+        echo 'realpath_cache_ttl = 600'; \
+    } > /usr/local/etc/php/php.ini
+
+RUN { \
+        echo 'date.timezone = Europe/Paris'; \
+        echo 'short_open_tag = off'; \
+        echo 'memory_limit = 8192M'; \
```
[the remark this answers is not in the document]
Pierstoval: By  the way, `date.timezone` can be set to UTC so everyone will have same dates on the app, WDYT?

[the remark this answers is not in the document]
mpiot: @Pierstoval Because I'm in France I've configure it like that, but you're totaly right.

mpiot:
Test fails because there is a vulnerability (Twig 2.7 required, but it will be fixed when #955 is merged)

Dockerfile
```diff
@@ -0,0 +1,179 @@
...
+    ; \
+    # Compile ICU (required by intl php extension)
+    curl -L -o /tmp/icu.tar.gz http://download.icu-project.org/files/icu4c/${ICU_VERSION}/icu4c-$(echo ${ICU_VERSION} | sed s/\\./_/g)-src.tgz; \
+    tar -zxf /tmp/icu.tar.gz -C /tmp; \
+    cd /tmp/icu/source; \
+    ./configure --prefix=/usr/local; \
+    make clean; \
+    make; \
+    make install; \
+    #Install the PHP extensions
```
athlan: I'd split it as separate `RUN` step, any mistake in app-specific set of `docker-php-ext` requires full rebuild of this later which takes ages.

mpiot:
@athlan I've added a .dockerignore file.

Dockerfile
mpiot: I've add all of it in a single layout to follow Dockerfile best practices that recommend to do the less number of layer as possible. Then, when you want to add something in your image, you can split it to test before, then regroup it after ?

I think, it's not totaly usefull to split this part, because we just need the split sometimes when you want to change something, but otherwise we never change anything.

That's my opinion. But I understand what you say, because when I edit this part I split it for try, because this part (with ICU) is very long.

Wait coments, but if more people prefer, we can split after the ICU install.

Dockerfile
```diff
@@ -0,0 +1,179 @@
...
+
+COPY . /app
+WORKDIR /app
+
+RUN yarn install && yarn build && rm -R node_modules
+
+#####################################
+##       PROD VENDOR BUILDER       ##
+#####################################
+FROM composer:${COMPOSER_VERSION} as vendor-builder
```
athlan: Here I'd inherit from `app` step, because if application relies on some extension installed before (in app step) (example bcmath) here this dependency will be missing. Proposal:

```
FROM app as vendor-builder

ARG COMPOSER_VERSION

# Install Composer
RUN set -ex; \
    EXPECTED_SIGNATURE="$(curl -L https://getcomposer.org/download/${COMPOSER_VERSION}/composer.phar.sha256sum)"; \
    curl -L -o composer.phar https://getcomposer.org/download/${COMPOSER_VERSION}/composer.phar; \
    ACTUAL_SIGNATURE="$(sha256sum composer.phar)"; \
    if [ "$EXPECTED_SIGNATURE" != "$ACTUAL_SIGNATURE" ]; then >&2 echo 'ERROR: Invalid installer signature' && rm /usr/local/bin/composer && exit 1 ; fi; \
    chmod +x composer.phar && mv composer.phar /usr/local/bin/composer; \
    RESULT=$?; \
    exit $RESULT;

```

mpiot: Previously, I was doing like it, but I've change to avoid have composer in the final image.

At the moment on all my projects, I've never had this problem (but, sure, I've not test all dependencies). The goal of this part is to only install vendors and avoid to have composer in the final image.

athlan: Final image won't have this dependency, because `app-prod` inherits from `app`, not `vendor-builder`. You wisely used only copy of `/app` files from `vendor-builder` while composer is located under `/usr/local/bin/composer`.

mpiot: Yes but if we install composer on `app` then, you have `composer` on `app-prod` and finaly on the image.

.dockerignore
```diff
@@ -0,0 +1,20 @@
...
+.gitignore
+.php_cs.dist
+.travis.yml
+appveyor.yml
+docker-compose.override.yml.dist
+docker-compose.yml
+Dockerfile
+Makefile
+phpunit.xml.dist
+README.md
```
athlan: I suggest add also:
* `/docker`
* `/node_modules` (might appear while development, where drive is mounted in docker-compose)

Dockerfile
athlan: Yes yes, that's why my proposal is to have it only on `vendor-builder`:

```
FROM app as vendor-builder
```

.dockerignore
mpiot: Poissible to miss some files yes, I've just copy a file from my repository, but I run the build in Travis, then some folder/files are never present.

For the docker folder, I can't because the Dockerfile need to access it.

athlan: Ah, yes, in case of CI/CD those folders probably wont appear. Tests outputs are also isolated in your makefiles. Just missed it checking build locally.

mpiot:
You want to install composer in vendor-builder that extend app, that's it ?

Then we install composer in the app-dev and vendor-builder ?

That's possible, but, I'm not totally convinced. Because at the moment on
some projects, I've never had a problem on vendor install because a php
extension missed.

Le ven. 2 août 2019 à 17:01, Piotr Pelczar <notifications@github.com> a
écrit :

Makefile
```diff
@@ -0,0 +1,166 @@
...
+test-all: lint test-schema security-check tests                                                        ## Lint all, run schema and security check, then unit and functionnal tests
+
+
+##
+## Dependencies
+##---------------------------------------------------------------------------
+
+deps: vendor assets                                                                                    ## Install the project dependencies
+
+
```
mstrom: What do you think do we need to add command to update dependancies?
`deps-update:
## Update the project dependencies
	$(EXEC) composer update -n
        $(EXEC) yarn upgrade
`

Makefile
```diff
@@ -0,0 +1,166 @@
...
+## Assets
+##---------------------------------------------------------------------------
+
+watch: node_modules                                                                                    ## Watch the assets and build their development version on change
+	$(EXEC) yarn watch
+
+assets: node_modules                                                                                   ## Build the development version of the assets
+	$(EXEC) yarn dev
+
+assets-build: node_modules                                                                             ## Build the production version of the assets
```
mstrom: `assets` and `assets-build` both building assets
maybe use `assets-prod` instead of `assets-build`?
```suggestion
assets-prod: node_modules                                                                             ## Build the production version of the assets
```

ker0x:
The project already encourages to use the Symfony binary, either to create the project or to launch the local server. Why not use the advantages it offers with Docker integration (https://symfony.com/doc/current/setup/symfony_server.html#docker-integration)?

landsman:
It would be nice to finalize this and merge it. Can I help?

setineos:
I think we should use the basic Docker configuration if we want to implement it in this repository.

landsman:
It would be nice to merge at least something. It's here from 2019. 😢

PR #933, opened 2019-01-16
Fixed ru translation

### LINKED ISSUE #932: Opening any post throws Exception
I open http://127.0.0.1:8000/ru/blog/ and then when I try to click on any post header, I get the Exception.

image

yceruto:
Fix #932

### DIFF
--- a/translations/messages+intl-icu.ru.xlf
+++ b/translations/messages+intl-icu.ru.xlf
@@ -306,7 +306,7 @@
             </trans-unit>
             <trans-unit id="post.num_comments">
                 <source>post.num_comments</source>
-                <target>{count, plural, one {# комментарий} few {# комментария} many {# комментариев}}</target>
+                <target>{count, plural, one {# комментарий} other {# комментариев}}</target>
             </trans-unit>
             <trans-unit id="post.commented_on">
                 <source>post.commented_on</source>


javiereguiluz:
I've been able to validate in my local app that this fixes the reported error, so let's merge it. Thanks Yonel.

translations/messages+intl-icu.ru.xlf
```diff
@@ -306,7 +306,7 @@
             </trans-unit>
             <trans-unit id="post.num_comments">
                 <source>post.num_comments</source>
-                <target>{count, plural, one {# комментарий} few {# комментария} many {# комментариев}}</target>
+                <target>{count, plural, one {# комментарий} other {# комментариев}}</target>
```
stof: this change is wrong. Russian has 3 differents forms. `few` should not have been removed

yceruto: @stof The previous format doesn't work, what is the correct solution that you suggest?

stof: well, `many` should have been `other`. But the `few` variant is still necessary for proper Russian pluralization AFAIK.

yceruto: Got it! fixed in https://github.com/symfony/demo/pull/934

PR #935, opened 2019-01-17
Reverted the handling of flash messages

Matvien:
Reverted the handling of flash messages


javiereguiluz:
Thanks for fixing this bug Evgeny.

PR #937, opened 2019-01-18
Use a Guard authenticator for user login

### LINKED ISSUE #924: New Feature request
https://symfony.com/doc/current/security/form_login_setup.html

please add also Login with FormBuilder

it will be very useful!

javiereguiluz:
This fixes #924.

This app always follows the latest Symfony recommendations ... and form login is now recommended to implement it with a Guard authenticator (see https://symfony.com/doc/current/security/form_login_setup.html). So, let's update our form login feature.

### DIFF
--- a/src/Security/LoginFormAuthenticator.php
+++ b/src/Security/LoginFormAuthenticator.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace App\Security;
+
+use App\Entity\User;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
+use Symfony\Component\Security\Core\Exception\InvalidCsrfTokenException;
+use Symfony\Component\Security\Core\Security;
+use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+use Symfony\Component\Security\Guard\Authenticator\AbstractFormLoginAuthenticator;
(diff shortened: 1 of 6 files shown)


src/Security/LoginFormAuthenticator.php
```diff
@@ -0,0 +1,92 @@
...
+use Symfony\Component\Security\Core\Exception\InvalidCsrfTokenException;
+use Symfony\Component\Security\Core\Security;
+use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+use Symfony\Component\Security\Guard\Authenticator\AbstractFormLoginAuthenticator;
+use Symfony\Component\Security\Http\Util\TargetPathTrait;
+
+class LoginFormAuthenticator extends AbstractFormLoginAuthenticator
```
yceruto: I am not against promoting the recommended form, but this class has many complex steps and concepts compared to the previous code. Unless you need custom behavior in the login process, I think this class is very difficult for newcomers.

javiereguiluz: You are right that this is more complex than before ... but if we consider that this is "too complex" ... then the problem lies in Symfony, which proposes a solution that is too complex.

@weaverryan what's your opinion? Should we leave this login form authenticator outside of this demo app? Thanks!

yceruto: Yep, this feels like a lot of *manual* statements (and a lot of knowledge) that are automatic with previous config, IMHO.
> Unless you need custom behavior in the login process

src/Security/LoginFormAuthenticator.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public function getUser($credentials, UserProviderInterface $userProvider)
+    {
+        $token = new CsrfToken('authenticate', $credentials['csrf_token']);
+        if (!$this->csrfTokenManager->isTokenValid($token)) {
+            throw new InvalidCsrfTokenException();
+        }
+
+        $user = $this->entityManager->getRepository(User::class)->findOneBy(['username' => $credentials['username']]);
```
mbabker: Is there a reason to use the EntityManager over the injected UserProvider here?

javiereguiluz:
Let's close this for now for being too complicated for newcomers. Thanks!

PR #943, opened 2019-02-05
Migrate from HTMLPurifier to html-sanitizer

tgalopin:
I would like to propose the usage of html-sanitizer in remplacement of HTMLPurifier.

html-sanitizer is well integrated in modern Symfony applications (autowiring / autoconfiguration), it is better suited than HTMLPurifier to sanitize user inputs such as Markdown, it is faster and it creates HTML that's simpler and easier to understand. It is also IMO easier to extends and configure.

### DIFF
--- a/src/Utils/Markdown.php
+++ b/src/Utils/Markdown.php
@@ -11,6 +11,8 @@
 
 namespace App\Utils;
 
+use HtmlSanitizer\SanitizerInterface;
+
 /**
  * This class is a light interface between an external Markdown parser library
  * and the application. It's generally recommended to create these light interfaces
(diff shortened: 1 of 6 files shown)


src/Utils/Markdown.php
```diff
@@ -22,22 +24,18 @@
 class Markdown
 {
     private $parser;
-    private $purifier;
+    private $sanitizer;
 
-    public function __construct()
+    public function __construct(SanitizerInterface $sanitizer)
     {
         $this->parser = new \Parsedown();
```
gmsantos: it's just a nail-care, but `Parsedown` could be moved to constructor too?

tgalopin: I thought about it and thought it actually shouldn't be part of this PR, as it's not the same aim. I can definitely do another PR with a service for the parser though.

javiereguiluz:
Thanks Titouan for this proposal. The new library looks indeed better!

I made two quick profiles before/after making this change and performance improvements are nice too:

* Blog index (`/en/blog`): 19% improvement (https://blackfire.io/profiles/compare/731fce4c-3aa5-41cb-988c-420d5e01c4e4/graph)
* Blog post (`/en/blog/posts/lorem-ipsum-dolor-sit-amet-consectetur-adipiscing-elit`): 27% improvement (https://blackfire.io/profiles/compare/d12e27d9-a911-498c-a611-50e76dcd891d/graph)

javiereguiluz:
Thank you Titouan.

Issue #946, opened 2019-02-15
use Voter constants instead of plain explicit string


inmarelibero:
using `PostVoter::SHOW` instead of `'show'` is preferable in https://github.com/symfony/demo/blob/master/src/Controller/Admin/BlogController.php#L122

so, 
` $this->denyAccessUnlessGranted(PostVoter::SHOW, $post, 'Posts can only be shown to their authors.');`
instead of 
` $this->denyAccessUnlessGranted('show', $post, 'Posts can only be shown to their authors.');`

I can produce a PR if needed

javiereguiluz:
I think we should do this change. In PHP code using these classes is easy and the resulting code is concise. It's not the same for the Twig templates, but we don't have to change this in the, (if any of the templates uses it in `is_granted()`).

Thanks for your proposal @inmarelibero!

inmarelibero:
PR was sent https://github.com/symfony/demo/pull/948/files

waiting for feedbacks/review

javiereguiluz:
Closing as fixed by #948.

PR #948, opened 2019-02-19
#946 use Voter constants instead of plain explicit string

inmarelibero:
address issue #946


src/Controller/Admin/BlogController.php
```diff
@@ -160,7 +161,7 @@ public function edit(Request $request, Post $post): Response
      */
     public function delete(Request $request, Post $post): Response
     {
-        if (!$this->isCsrfTokenValid('delete', $request->request->get('token'))) {
+        if (!$this->isCsrfTokenValid(PostVoter::DELETE, $request->request->get('token'))) {
```
inmarelibero: usage is right, but I don't know if we want here to use the original string or the class constant

doc is https://symfony.com/doc/3.4/controller/csrf_token_validation.html, but imho the doc is not clear about what a developer should use as `token_id`

src/Security/PostVoter.php
```diff
@@ -28,9 +28,9 @@ class PostVoter extends Voter
 {
     // Defining these constants is overkill for this simple application, but for real
     // applications, it's a recommended practice to avoid relying on "magic strings"
-    private const SHOW = 'show';
+    const DELETE = 'delete';
     private const EDIT = 'edit';
```
inmarelibero: it's ugly here that only the `EDIT` constant is left as private

solutions:
1) leave as is, ugly but correct
2) make constant public, even if not needed
3) make `EDIT` constant public, and use `{% if is_granted(constant('App\\Security\\PostVoter::EDIT'), post) %}` in https://github.com/symfony/demo/blob/master/templates/blog/post_show.html.twig#L62 instead of `{% if is_granted('edit', post) %}`

src/Controller/Admin/BlogController.php
stof: you use what you want, as long as you use the same id when validating the token here and when generating it in the template

src/Security/PostVoter.php
yceruto: I'd suggest (2) `make constant public, even if not needed` just for consistency.

javiereguiluz: Yes, I also like the option (2).

Besides, I think we should add the `public` modifier explicitly to the constants to let developers know that we are making them public consciously. Thanks!

src/Controller/Admin/BlogController.php
inmarelibero: ah ok it has nothing to do with PostVoter then.. I'll revert here

src/Security/PostVoter.php
inmarelibero: visibility updated: now all constants are `public`

inmarelibero:
work done here, waiting for review/merge

do I have to squash commits?

javiereguiluz:
@inmarelibero thanks for this contribution! This is now merged. By the way, there's no need to squash the commits because the tool we use to merge pull requets does that automatically. Thanks!

PR #949, opened 2019-02-21
Redirect to the right page after changing the locale in the login page

javiereguiluz:
This fixes the problem reported in https://github.com/symfony/symfony-docs/issues/10864

I don't think this is a Symfony bug or a Demo app bug ... I think it's an edge case for the reasons explained in the code ... and that's why I propose this solution.

### DIFF
--- a/src/Controller/SecurityController.php
+++ b/src/Controller/SecurityController.php
@@ -12,9 +12,11 @@
 namespace App\Controller;
 
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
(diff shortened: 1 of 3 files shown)


yceruto:
Removing the possibility to change the locale in login page feel bad to me :( I think this should be solved in other way around.

What do you think about regenerate the target path (if any) on `login` method?
```php
/**
 * @Route("/login", name="security_login")
 */
public function login(AuthenticationUtils $helper, Session $session): Response
{
    if ($session->has('_security.main.target_path')) {
        $session->set('_security.main.target_path', $this->generateUrl('admin_index'));
    }

    // ...
}
```
Therefore, we are updating the uri with the correct locale if it was changed in the login page.

yceruto:
Note that `admin_index` is the only secure zone in this app, so we can hardcoded it safely.

javiereguiluz:
@yceruto I agree that your solution is much better. We could use the TargetPathTrait to simplify the code a bit. Thanks.

PR #951, opened 2019-02-26
Block login page for logged in users

Carrancio:
This PR blocks the login page for logged in users.

### DIFF
--- a/src/Controller/SecurityController.php
+++ b/src/Controller/SecurityController.php
@@ -37,7 +37,7 @@ class SecurityController extends AbstractController
     {
         // Block login page for logged in users
         if ($this->container->get('security.authorization_checker')->isGranted('ROLE_USER')) {
-            return new RedirectResponse('/');
+            return $this->redirectToRoute('blog_index');
         }
 
         // this statement solves an edge-case: if you change the locale in the login
--- a/src/Controller/SecurityController.php
+++ b/src/Controller/SecurityController.php
@@ -12,6 +12,7 @@
 namespace App\Controller;
 
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@Carrancio thanks for this contribution.

As you know, this app is "the reference Symfony app" and lots of people (not only newcomers) look into it to check how things are done. So it's very important that we only do what Symfony really promotes as a best practice.

Although your proposal is technical correct ... I wonder if this is something we should do and promote. I've never seen this in a Symfony app ... and we don't recommend this in Symfony Docs. But that doesn't mean that your proposal is wrong. I just want to wait and read more opinions about this. Thanks.

mbabker:
I'd say this is a good practice to follow, even if not promoted in documentation.  As an authenticated user, if I visit the login page and see a login form I would be mildly confused as to why I'm seeing it.  So this is a check I implement in my apps.

FWIW Laravel ships [a middleware](https://github.com/laravel/laravel/blob/4997f08105d6fb941b1ed2c749ebd2f55fba5a74/app/Http/Middleware/RedirectIfAuthenticated.php) in their skeleton application, enabled by default, that allows you to define routes and/or controllers that should be accessed by guests only (and this is applied to the default [`LoginController`](https://github.com/laravel/laravel/blob/4997f08105d6fb941b1ed2c749ebd2f55fba5a74/app/Http/Controllers/Auth/LoginController.php)).  Another practical example would be trying to reach https://github.com/login while logged in here on GitHub.

src/Controller/SecurityController.php
```diff
@@ -34,6 +35,11 @@ class SecurityController extends AbstractController
      */
     public function login(Request $request, AuthenticationUtils $helper): Response
     {
+        // Block login page for logged in users
+        if ($this->container->get('security.authorization_checker')->isGranted('ROLE_USER')) {
+            return new RedirectResponse('/');
```
mbabker: To be consistent with other parts of the demo, use `return $this->redirectToRoute('blog_index');` instead of directly creating the RedirectResponse object.

javiereguiluz:
@Carrancio this is now merged! Thanks ... and congrats on your first Symfony Demo contribution.

Note: while merging, we refactored the code to use the `Symfony\Component\Security\Core\Security` class, which is the recommended in the Symfony Docs.

PR #955, opened 2019-03-13
Update for Twig 2.7: `Template` -> `TemplateWrapper`

bobdenotter:
Since Twig 2.7, `$twig->resolveTemplate($template)` doesn't return `Twig\Template` but `Twig\TemplateWrapper`. This PR simply updates the typehints accordingly.

PR #960, opened 2019-03-29
Update Encore and tweak some of its settings

Lyrkan:
This PR:

* Updates the version of Encore to `^0.26`
* Updates the related bundle to `^1.4`
* Fixes the imported font paths that were not working anymore due to [the recent `resolve-url-loader` update](https://github.com/symfony/webpack-encore/pull/540)
* Enables integrity hashes by default
* Adds `core-js@3` and configures Babel properly so polyfills are automatically added to the resulting build
* Disables file versioning when not in production mode (which prevented from using the dev-server)


Lyrkan:
Failing tests should be fixed by #955.

javiereguiluz:
Thanks a lot for working on all this! I tried to merge this and solve the conflicts myself, but I couldn't. Could you please solve the conflicts? Thanks.

Lyrkan:
It should be ok now :)

webpack.config.js
```diff
@@ -10,13 +10,18 @@ Encore
...
     .enableSassLoader()
-    .enableVersioning()
+    .enableVersioning(Encore.isProduction())
     .addEntry('app', './assets/js/app.js')
     .addEntry('login', './assets/js/login.js')
     .addEntry('admin', './assets/js/admin.js')
     .addEntry('search', './assets/js/search.js')
     .splitEntryChunks()
     .enableSingleRuntimeChunk()
+    .enableIntegrityHashes()
```
stof: should be enabled only in production. In dev, it would cause issue when using the watcher, as I'm not sure the updated `entrypoints.json` gets dumped again by the watcher (and then, that would make it an integrity violation)

Lyrkan: It actually works fine in both `--watch` and `dev-server` mode.

For the first one, the `entrypoints.json` does get dumped again with updated hashes, and since `--watch` implies a browser refresh the `integrity` attribute gets updated accordingly.

Here is an example with some additional logs after changing the `app.js` file:

image

And with the `dev-server` enabled the integrity hashes are not computed at all since the files aren't dumped. This results in an empty `integrity` object in the `entrypoints.json` and no `integrity` attributes in the generated tags:

```js
{
  "entrypoints": {
    // (...)
    "search": {
      "js": [
        "http://localhost:8080/build/runtime.js",
        "http://localhost:8080/build/vendors~admin~app~login~search.js",
        "http://localhost:8080/build/vendors~admin~app~search.js",
        "http://localhost:8080/build/vendors~app~search.js",
        "http://localhost:8080/build/vendors~search.js",
        "http://localhost:8080/build/search.js"
      ]
    }
  },
  "integrity": {}
}
```

javiereguiluz:
Thank you Vincent.

PR #961, opened 2019-03-31
Remove stylesheet entries from Encore config

voronkovich:
The documentation says:

> **Using addStyleEntry() is supported, but not recommended**. A better option is to use follow the pattern above: use addEntry() to point to a JavaScript file, then require the CSS needed from inside of that.

> With Encore, think of your app.js file like a standalone JavaScript application: it will require all of the dependencies it needs (e.g. jQuery or React), including any CSS.

See https://symfony.com/doc/current/frontend/encore/simple-example.html#compiling-only-a-css-file


javiereguiluz:
Much better indeed! Thanks Oleg!

PR #966, opened 2019-04-10
Add MakerBundle as a dev dependency

bocharsky-bw:
What about to add a maker bundle to this Demo distribution? I think a lot of people are using Demo for some testing, bug reproducing, etc. and have to install it maker bandle themselves - it would be cool to have it installed already. Also, I think it's good to reveal Maker bundle to newcomers as well.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,6 +41,7 @@
         "symfony/css-selector": "*",
         "symfony/debug-bundle": "*",
         "symfony/dotenv": "*",
+        "symfony/maker-bundle": "^1.11",
         "symfony/phpunit-bridge": "*",
         "symfony/stopwatch": "*",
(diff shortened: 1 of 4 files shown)


javiereguiluz:
Thinking about this ... what if we make it a real dependency, instead of a "dev" dependency? MakerBundle is great and having installed in Symfony Demo would be a great way to showcase it.

bocharsky-bw:
Hm, but what's the point of using it in prod? I just don't want to make devs think that it's a good idea to use MakerBundle in prod environment

javiereguiluz:
Yes, you are right! It wasn't a very smart suggestion 😅

javiereguiluz:
Merged! Thanks.

PR #968, opened 2019-04-26
Fix getting template source and update deps

### LINKED ISSUE #967: 500 error after update
when I log on admin area I receive this error

Argument 1 passed to App\Twig\SourceCodeExtension::getTemplateSource() must be an instance of Twig\TemplateWrapper, instance of __TwigTemplate_2cbf7fc5651a791395a7872583630b929ff0fd2ae3fa9cc8e5edc91b3a4ac77e given, called in /var/www/ ....

could you recheck in your side if this admin area it is working

bocharsky-bw:
Fix #967 and some vulnerability issues in deps

### DIFF
--- a/src/Twig/SourceCodeExtension.php
+++ b/src/Twig/SourceCodeExtension.php
@@ -93,17 +93,11 @@ class SourceCodeExtension extends AbstractExtension
         return new \ReflectionFunction($callable);
     }
 
+    /**
+     * @param TemplateWrapper|Template $template
+     */
     private function getTemplateSource($template): array
     {
-        if (!$template instanceof TemplateWrapper && !$template instanceof Template) {
-            throw new \Exception(sprintf(
-                'The given template should be instance of %s or %s, but %s given',
-                TemplateWrapper::class,
-                Template::class,
-                \get_class($template)
-            ));
-        }
-
         $templateSource = $template->getSourceContext();
 
         return [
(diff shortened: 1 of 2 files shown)


src/Twig/SourceCodeExtension.php
```diff
@@ -92,8 +93,17 @@ private function getCallableReflector(callable $callable): \ReflectionFunctionAb
         return new \ReflectionFunction($callable);
     }
 
-    private function getTemplateSource(TemplateWrapper $template): array
+    private function getTemplateSource($template): array
     {
+        if (!$template instanceof TemplateWrapper && !$template instanceof Template) {
+            throw new \Exception(sprintf(
```
yceruto: I'm not sure this exception is needed... this function is private/local and `$twig->resolveTemplate($template)` ony returns `TemplateWrapper` or `Template`.

I'd suggest adding a PHPDoc block instead.

yceruto:
Shouldn't we also update this line?
https://github.com/symfony/demo/blob/4bc69f63c950a31a046b82d7fef19de4c831c4af/composer.json#L33

src/Twig/SourceCodeExtension.php
bocharsky-bw: I'm fine with PHPDoc, just was coding defensively... Let's wait for more feedback then

bocharsky-bw:
> Shouldn't we also update this line?

Hm, don't think we needed. Updated to what?

src/Twig/SourceCodeExtension.php
javiereguiluz: I agree with Yonel on this one.

bocharsky-bw: OK, it's simpler now 👍

javiereguiluz:
Thank you Victor; it's merged now.

PR #971, opened 2019-05-06
Fix tag icons after FontAwesome 5.8 upgrade

### LINKED ISSUE #970: Tag icon after updating to FontAwesome 5.8
It looks that the new update to FontAwesome 5.8 has broken the tag icon.

image

bocharsky-bw:
Fix #970

Now it looks like:
Screen Shot 2019-05-06 at 12 30 41


javiereguiluz:
Thanks for fixing this bug Victor.

PR #974, opened 2019-05-09
Updated to Symfony 4.3

javiereguiluz:


phpunit.xml.dist
```diff
@@ -21,6 +22,8 @@
     </testsuites>
 
     <listeners>
+        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
```
noniagriconomie: What does this do? Thx

config/packages/security.yaml
```diff
@@ -1,8 +1,9 @@
 security:
     encoders:
         # Our user class and the algorithm we'll use to encode passwords
+        # 'auto' means to let PHP choose the best possible password hasher (Argon2 or Bcrypt)
```
noniagriconomie: Php or Symfony?

phpunit.xml.dist
javiereguiluz: Don't know exactly ... but the official recipe includes it: https://github.com/symfony/recipes/blob/3b19bd94886555df1997b441b6cf6a1fab088f3a/symfony/phpunit-bridge/4.1/phpunit.xml.dist#L31

config/packages/security.yaml
javiereguiluz: Updated.

config/packages/security.yaml
```diff
@@ -1,8 +1,9 @@
 security:
     encoders:
         # Our user class and the algorithm we'll use to encode passwords
+        # 'auto' means to let Symfony choose the best possible password hasher (Argon2 or Bcrypt)
```
yceruto: `argon2i` and `bcrypt` are deprecated since 4.3
The new 'auto' choice list is `sodium` or `native`

yceruto: ref: https://github.com/symfony/symfony/blob/c5772ae9bc3d9a238a9d03b97afc773ce16b93ee/src/Symfony/Component/Security/Core/Encoder/EncoderFactory.php#L87-L89

stof: which are themselves using bcrypt or argon2i as algorithm.

yceruto: Yes, I meant the encoder names used for this `auto` option.

javiereguiluz:
After updating to Symfony 4.3 beta2, I see this deprecation:

image

I can't find any occurrence of `for ... if` in our templates and the deprecation message doesn't tell us which template contains the error. Any clue? Thanks.

yceruto:
I guess it comes from the `vendor/symfony/swiftmailer-bundle/Resources/views/Collector/swiftmailer.html.twig` template.

yceruto:
Also I see another deprecation in tests: 
```
  1x: The "DAMA\DoctrineTestBundle\PHPUnit\PHPUnitListener" class extends "PHPUnit\Framework\BaseTestListener" that is deprecated Use TestListenerDefaultImplementation trait instead.
    1x in DebugClassLoader::loadClass from Symfony\Component\Debug
```

yceruto:
> I can't find any occurrence of for ... if in our templates and the deprecation message doesn't tell us which template contains the error. Any clue? Thanks.

@javiereguiluz I sent a PR to improve it https://github.com/twigphp/Twig/pull/3031

fabpot:
Updating SwiftmailerBundle will remove some deprecations.

javiereguiluz:
I no longer see deprecations after updating Symfony to 4.3 RC1 and the other dependencies too. @yceruto do you still see deprecations in tests? I can't see anything. Thanks.

yceruto:
I still see this one https://github.com/symfony/demo/pull/974#issuecomment-496197569 https://travis-ci.org/symfony/demo/jobs/538240998#L616

PR #976, opened 2019-05-11
Updated docblock links

miguelci:
Was following the documentation here and one of the links wasn't going anywhere.
Updated the link that wasn't working and also others that had redirects.
Not a big PR, but might help someone.


javiereguiluz:
Thanks for fixing all these outdated links!

Issue #975, opened 2019-05-09
Update the way events are used


javiereguiluz:
After #974 is merged, we should refactor the way we use events as explained in https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching

In practice, we should:

* Remove the `src/Events.php` and use event objects instead of event names
* Create a new event object to replace the GenericEvent used in BlogController
* Update the calls to dispatch() to remove the event name

ayshiff:
I would like to work on it 👍

javiereguiluz:
Fixed by #977.

PR #977, opened 2019-05-25
Feature: update events sf43

ayshiff:
This PR updates the way events are used since Symfony 4.3 Beta 2 [#975](https://github.com/symfony/demo/issues/975)

- removed Event class
- created a `CommentCreatedEvent` event object
- update calls to dispatch()
- used `Symfony\Contracts\EventDispatcher\Event` instead of the `Event` class

I have a problem when I update the calls to dispatch() to remove the event name. The `comment.created` event is inside the `not called listeners` but it works fine when I keep the event name.   
I don't know if it is an issue or not.


src/Controller/BlogController.php
```diff
@@ -109,14 +108,17 @@ public function commentNew(Request $request, Post $post, EventDispatcherInterfac
...
+            // https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching
+            $event = new CommentCreatedEvent($comment);
 
             // When an event is dispatched, Symfony notifies it to all the listeners
             // and subscribers registered to it. Listeners can modify the information
             // passed in the event and they can even modify the execution flow, so
             // there's no guarantee that the rest of this controller will be executed.
             // See https://symfony.com/doc/current/components/event_dispatcher.html
-            $eventDispatcher->dispatch(Events::COMMENT_CREATED, $event);
+            $eventDispatcher->dispatch($event, CommentCreatedEvent::NAME);
```
yceruto: The best part of this new feature is that we can omit the second argument here

src/EventSubscriber/CommentNotificationSubscriber.php
```diff
@@ -41,14 +40,14 @@ public function __construct(\Swift_Mailer $mailer, UrlGeneratorInterface $urlGen
     public static function getSubscribedEvents(): array
     {
         return [
-            Events::COMMENT_CREATED => 'onCommentCreated',
+            CommentCreatedEvent::NAME => 'onCommentCreated',
```
yceruto: using `CommentCreatedEvent::class` instead

src/Events/CommentCreatedEvent.php
```diff
@@ -0,0 +1,36 @@
...
+use App\Entity\Comment;
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * The comment.created event is dispatched each time a comment is created
+ * in the system.
+ */
+class CommentCreatedEvent extends Event
+{
+    public const NAME = 'comment.created';
```
yceruto: Now it's useless

yceruto:
The composer.json needs to be updated to requiere 4.3-beta2 and make tests pass

ayshiff:
Maybe I should wait that [#974](https://github.com/symfony/demo/pull/974) is merged.

javiereguiluz:
@ayshiff thanks for this contribution! We did some minor tweaks while merging, but your pull request was great. Thanks to reviewers too.

PR #981, opened 2019-05-31
Do not install assets with "--symlink --relative"

seb-jean:
By default do not install assets with "--symlink --relative".


bocharsky-bw:
Could you explain why not?

javiereguiluz:
We prefer to keep this unchanged. So far it's worked great. The current problem with assets it's more related to "non-built assets" so we're trying to fix it differently. Thanks.

PR #982, opened 2019-06-02
Rebuild assets and return back styles

### LINKED ISSUE #980: No assets are built on installation and there are no instructions for this
Doing what's currently in README.md is enough for running the page, but it doesn't install any assets, so page is completely without any css.

bocharsky-bw:
Fix #980

Looks like assets were not rebuilt/committed on a PR that was merged to master that's why we lost all styles after installing Demo from scratch following the README

I just rebuilt the assets and committed changes here.


javiereguiluz:
I don't understand this PR:

1) If I try to merge it, the tool we use says: *"Unable to merge branch pull/982. Resolve the conflict by hand and start the command again"* But there's no conflict.
2) If I try to recreate this PR myself, I can't. I rebuild assets ... but end with the same result as in the existing master branch of the repo.

So, I can't merge this ... and I can't reproduce this 🤔  Could you please double check what's going on here? Thanks.

bocharsky-bw:
Ah, you were faster than me :)

P.S. Not sure what was the problem though, glad you were able to merge it this way

PR #983, opened 2019-06-02
Add development section to explain working with assets

bocharsky-bw:
Add a few words about how to work with assets in Symfony Demo, basically to help with questions like #980


javiereguiluz:
@bocharsky-bw thanks for this! I like the idea ... but not the implementation. Adding this in the README looks like the project is complicated to setup. So, I've moved(and reworded a bit) your comments from README to webpack.config.js, adding also a link to the Symfony Docs article about front-end. Cheers!

PR #984, opened 2019-06-02
Fail CI if the repo is in a dirty state after building assets

bocharsky-bw:
In order to prevent things fixed in #982 I think we can check for missing assets on TravisCI

Tests should pass after merging #982

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: php
 sudo: false
 
 cache:
+  yarn: true
   directories:
     - $HOME/.composer/cache/files
     - ./bin/.phpunit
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Very nice! I like this idea a lot. I'm going to change `yarn build` by `yarn build production` to be more explicitly.

bocharsky-bw:
@javiereguiluz not sure about `yarn build production` command, because it fails even locally for me, does it work for you locally?

javiereguiluz:
It should be `yarn encore production` (and `yarn encore dev --watch`) Sorry! This is being fixed in #990.

PR #991, opened 2019-06-04
Switch to Doctrine's Pager

javiereguiluz:
Pagerfanta library could be soon be unmaintained: https://github.com/whiteoctober/Pagerfanta/issues/278

Let's switch to Doctrine Pager.

### DIFF
--- a/src/Repository/PostRepository.php
+++ b/src/Repository/PostRepository.php
@@ -15,7 +15,7 @@ use App\Entity\Post;
 use App\Entity\Tag;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
 use Doctrine\Common\Persistence\ManagerRegistry;
-use Doctrine\ORM\Query;
+use Doctrine\ORM\QueryBuilder;
 use Doctrine\ORM\Tools\Pagination\Paginator;
 
 /**
@@ -37,8 +37,6 @@ class PostRepository extends ServiceEntityRepository
 
     public function findLatest(int $page = 1, Tag $tag = null): array
     {
-        $page = $page < 1 ? 1 : $page;
-
         $qb = $this->createQueryBuilder('p')
             ->addSelect('a', 't')
             ->innerJoin('p.author', 'a')
@@ -46,15 +44,14 @@ class PostRepository extends ServiceEntityRepository
             ->where('p.publishedAt <= :now')
             ->orderBy('p.publishedAt', 'DESC')
             ->setParameter('now', new \DateTime())
-            ->setFirstResult(($page - 1) * Post::NUM_ITEMS)
-            ->setMaxResults(Post::NUM_ITEMS);
+        ;
 
         if (null !== $tag) {
             $qb->andWhere(':tag MEMBER OF p.tags')
                 ->setParameter('tag', $tag);
         }
 
-        return $this->createPaginator($qb->getQuery(), $page);
+        return $this->createPaginator($qb, $page);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/Repository/PostRepository.php
```diff
@@ -36,31 +35,24 @@ public function __construct(ManagerRegistry $registry)
...
     {
         $qb = $this->createQueryBuilder('p')
             ->addSelect('a', 't')
             ->innerJoin('p.author', 'a')
             ->leftJoin('p.tags', 't')
             ->where('p.publishedAt <= :now')
             ->orderBy('p.publishedAt', 'DESC')
-            ->setParameter('now', new \DateTime());
+            ->setParameter('now', new \DateTime())
+            ->setFirstResult(($page - 1) * Post::NUM_ITEMS)
```
stof: you no longer deal with negative page numbers properly here (Pagerfanta was validating that).

src/Twig/AppExtension.php
```diff
@@ -88,4 +91,21 @@ public function getLocales(): array
...
         return $this->locales;
     }
+
+    public function getPaginator(Paginator $paginator, int $currentPage, int $pageSize = Post::NUM_ITEMS): array
+    {
+        $hasPreviousPage = $currentPage > 1;
+        $hasNextPage = ($currentPage * $pageSize) < $paginator->count();
+
+        return [
+            'results' => $paginator,
```
stof: should do `$paginator->getIterator()` to have the right count of elements in the results.

src/Repository/PostRepository.php
```diff
@@ -36,31 +35,24 @@ public function __construct(ManagerRegistry $registry)
         parent::__construct($registry, Post::class);
     }
 
-    public function findLatest(int $page = 1, Tag $tag = null): Pagerfanta
+    public function findLatest(int $page = 1, Tag $tag = null): Paginator
```
stof: this is not a good idea to pass directly to Twig like that. The Countable implementation of the Paginator gives you the total number of results, not the number of items in the iterator itself. This will confuse twig (`loop.last` will become wunky for instance).

javiereguiluz:
@stof thanks for your review! I've changed some things thanks to your review.

src/Repository/PostRepository.php
```diff
@@ -110,4 +104,22 @@ private function extractSearchTerms(string $searchQuery): array
...
+
+        return [
+            'results' => $paginator->getIterator(),
+            'currentPage' => $currentPage,
+            'hasPreviousPage' => $hasPreviousPage,
+            'hasNextPage' => $hasNextPage,
+            'previousPage' => $hasPreviousPage ? $currentPage - 1 : null,
+            'nextPage' => $hasNextPage ? $currentPage + 1 : null,
+            'numPages' => (int) ceil($paginator->count() / $pageSize),
+            'haveToPaginate' => $paginator->count() > $pageSize,
```
bocharsky-bw: We call `$paginator->count()` 3 times here, will it execute 3 queries? Maybe put it on a var:
`$count = $paginator->count();`

src/Repository/PostRepository.php
```diff
@@ -36,15 +35,19 @@ public function __construct(ManagerRegistry $registry)
...
+
         $qb = $this->createQueryBuilder('p')
             ->addSelect('a', 't')
             ->innerJoin('p.author', 'a')
             ->leftJoin('p.tags', 't')
             ->where('p.publishedAt <= :now')
             ->orderBy('p.publishedAt', 'DESC')
-            ->setParameter('now', new \DateTime());
+            ->setParameter('now', new \DateTime())
+            ->setFirstResult(($page - 1) * Post::NUM_ITEMS)
```
bocharsky-bw: What about to move it to a function, like:
```
public static function getFirstResult($page)
{
    $page = $page < 1 ? 1 : $page;

    return ($page - 1) * Post::NUM_ITEMS;
}
```

templates/blog/index.html.twig
```diff
@@ -24,9 +24,29 @@
...
+                        <li class="active"><span>{{ i }} <span class="sr-only">(current)</span></span></li>
+                    {% else %}
+                        <li><a href="{{ path('blog_index_paginated', {page: i}) }}">{{ i }}</a></li>
+                    {% endif %}
+                {% endfor %}
+
+                {% if paginator.hasNextPage %}
+                    <li class="prev"><a href="{{ path('blog_index_paginated', {page: paginator.nextPage}) }}" rel="next">Next →</a></li>
+                {% else %}
+                    <li class="next disabled"><span>Next →</span></li>
```
bocharsky-bw: Maybe better to use FontAwesome arrows here instead to match the styles of icons we have?

Pierstoval:
Great change! Straightforward, clear, customizable…

Now to all questions like _"How to create a paginator?"_, just need to answer _"Look at Symfony Demo"_ 👍

JarJak:
It would be better if `createPaginator` would return a paginator object, not just an array. It will help with debugging, IDE suggesting, types control etc.

bocharsky-bw:
I think PhpStorm IDE should figure out those array keys and suggest autocompleting for them... Not sure about autocompleting in Twig templates but it should work in PHP files as far as I know. But agree,  paginator object will give us more strict format for this

PR #990, opened 2019-06-03
Updated the front-end dependencies

javiereguiluz:

PR #994, opened 2019-06-06
Rebuild assets

bocharsky-bw:
Somehow running `yarn build` on TravisCI  sets the repo in a dirty state:
https://travis-ci.org/symfony/demo/builds/541616536

Actually, the same for me when I run `yarn build` locally. I rebuilt the assets - let's check if tests pass now.

P.S. Basically, continue https://github.com/symfony/demo/pull/990#issuecomment-498306872


javiereguiluz:
This is still failing (The command "yarn install && yarn encore production && git add --all && git diff --staged --shortstat --exit-code" exited with 1.) but I'm going to merge it anyway to have the assets updated. Thanks.

bocharsky-bw:
Hm, I see it fails, and that's weird :/ Probably the output depends on Yarn/NPM version... and if so - it makes difficult (impossible?) to use this little trick to fail builds if `yarn encore production` generated a new output. It sounds like we should match the env on locally and on TravisCI. I'll think about this problem, not sure how to fix it properly yet

stof:
could the issue be caused by what is described in this comment: https://github.com/webpack-contrib/mini-css-extract-plugin/issues/308#issuecomment-445325727 ?

bocharsky-bw:
Thanks @stof ! Yes, I confirm that's the problem, the paths different on CI and on local computers where we generate and commit them to the repo. Though, not sure about possible workaround here, The only way I see so far - turn off versioning, then the trick with checking diffs on CI should work

Lyrkan:
Weird, I thought this issue only happened with recent versions of `mini-css-extract-plugin` (https://github.com/symfony/webpack-encore/pull/564 has been blocked for a while because of that)...

I couldn't find any workaround and I'm not sure fixing it in `mini-css-extract-plugin` will be as simple as what's described in the comment linked by @stof (using `dependency.content` for `this._identifier`) so, for now, disabling versioning for the commited files as suggested in #1004 is probably the best thing to do.

bocharsky-bw:
Thank you for your feedback @Lyrkan ! Actually, IIRC (I've tried many solutions) we have problems with changed hashes for JS files only, CSS ones are good.

Issue #992, opened 2019-06-05
SYMFONY_PHPUNIT_VERSION in the phpunit.xml.dist has no effect


tristanbes:
Hello,

It seems that `<env name="SYMFONY_PHPUNIT_VERSION" value="7.1"/>` contained in the [phpunit.xml.dist](https://github.com/symfony/demo/blob/master/phpunit.xml.dist#L15) has literally no effect on the PHPUnit version used instead of the version `7.1`

Let's see for yourself on your Travis builds: https://travis-ci.org/symfony/demo/jobs/540800942#L637 where version `6.5.14` is used.

[Commenting those 3 lines seems to fix this problem](https://github.com/symfony/demo/blob/master/bin/phpunit#L11-L13), reading the value from phpunit.xml.dist again.

--- 

Thanks @Kocal to help me tracking this down.

mbabker:
The `bin/phpunit` file (and probably `phpunit.xml.dist`) should be updated for the PHPUnit bridge changes in SF 4.3, which includes removing the lines you've highlighted.  But, that's probably blocked right now by #998.

dmaicher:
@javiereguiluz so this one can also be closed? :wink: Was fixed by https://github.com/symfony/demo/pull/1000

javiereguiluz:
Yes, closing as as fixed. Thanks!

PR #1000, opened 2019-06-11
use phpunit 7.5

dmaicher:
See https://github.com/symfony/demo/issues/992

And I also updated to phpunit 7.5

### DIFF
--- a/bin/phpunit
+++ b/bin/phpunit
@@ -8,9 +8,6 @@ if (!file_exists(dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-php
 if (false === getenv('SYMFONY_PHPUNIT_REMOVE')) {
     putenv('SYMFONY_PHPUNIT_REMOVE=');
 }
-if (false === getenv('SYMFONY_PHPUNIT_VERSION')) {
-    putenv('SYMFONY_PHPUNIT_VERSION=6.5');
-}
 if (false === getenv('SYMFONY_PHPUNIT_DIR')) {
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Thanks a lot David! Also, congrats on creating the 1,000th PR of this repository 🎉

Kocal:
Btw, I really think we should removes [those lines](https://github.com/symfony/demo/blob/master/bin/phpunit#L8-L13) too. 
We will face the same problem whenever we will use `SYMFONY_PHPUNIT_REMOVE` or `SYMFONY_PHPUNIT_DIR`

dmaicher:
@Kocal feel free to open a PR for it :+1:

Kocal:
Yep :+1: 

Also, does someone know why do we need to commit the `bin/phpunit` file? 
We had some internal discussion with @tristanbes and we could not decide who _is right_ or not... :smile:

javiereguiluz:
@Kocal about committing files ... keep in mind that this application is very special. We want everything to work without doing anything or running any command. That's why we do things (e.g. commit the built assets) that you shouldn't do in real apps.

Kocal:
> We want everything to work without doing anything or running any command

Hum, but how can it works since `vendor` folder is not committed? :confused: 

It requires a `composer install`, and running this command will copy binary files in the `bin` directory and so we don't need to commit it, right?

javiereguiluz:
@Kocal I meant you don't have to do anything ... when installing it via `create-project` or the `symfony` binary. We don't care about people cloning the repo because they know what to do.

Issue #998, opened 2019-06-10
Fix Travis failing tests


javiereguiluz:
This is our Travis CI test suite now:

![image](https://user-images.githubusercontent.com/73419/59186893-50d34100-8b74-11e9-98ab-20b41c129185.png)

The failing tests look like this:

![image](https://user-images.githubusercontent.com/73419/59186908-5c266c80-8b74-11e9-87c1-2b6d7e4fc83e.png)

Full details: https://travis-ci.org/symfony/demo/jobs/543634635

This is NOT related to #997. It was failing way before. Anyone can help us fix this? Thanks!

dmaicher:
Currently looking into this. Can reproduce it locally using php 7.1.30

dmaicher:
@javiereguiluz fixed in https://github.com/symfony/demo/pull/1001

PR #1001, opened 2019-06-11
revert fixtures db change

dmaicher:
This fixes https://github.com/symfony/demo/issues/998

The issue was that the fixture db data contained user passwords using `argon2` encoder algo which is not implemented in php 7.1 afaik. So authentication failed in the tests.

So this reverts the db changes done in https://github.com/symfony/demo/commit/6c32238022bb0666f06d04a2b527b803591f799b.

Build is green now :wink:


javiereguiluz:
Great detective work! Thank you so much David. It's great to see all tests green again 💚

PR #1002, opened 2019-06-11
chore(phpunit): remove lines for `SYMFONY_PHPUNIT_REMOVE`

Kocal:
It's a following of #1000.

The `phpunit` binary looks [like this](https://github.com/symfony/recipes/blob/master/symfony/phpunit-bridge/4.3/bin/phpunit) for Symfony 4.3, we don't need `SYMFONY_PHPUNIT_REMOVE` lines too.

I also think we don't need `SYMFONY_PHPUNIT_DIR` lines but I'm not really sure (it's working fine without those lines in one of our app).


javiereguiluz:
Thanks Hugo.

PR #1004, opened 2019-06-15
Tweak versioning and return back failing on CI

bocharsky-bw:
A possible fix for: https://github.com/symfony/demo/pull/994#issuecomment-499892242


bocharsky-bw:
Hm, still does not work, probably the problem is in something else, depends on how that "hash" is generated, because when I manually change the full path to the repo locally - it still works for me. Closing it

javiereguiluz:
@bocharsky-bw would this be fixed (or would it simplify the solution) if we disable versioning? I think we can do that for this demo app.

bocharsky-bw:
Hm, if we could consider disabling versioning on this app - I think it might help then. Let's figure it out

bocharsky-bw:
OK, this looks like a working example based on TravisCI

bocharsky-bw:
I think this one is ready

javiereguiluz:
Thanks Victor for working on this feature, this is much appreciated.

PR #1006, opened 2019-07-03
Added a Paginator object based on Doctrine's paginator

javiereguiluz:
The current code uses an array to provide the pagination information. We did that as part of the quick proof-of-concept replacement of Pagerfanta library.

This PR proposes to replace that array by a proper object. The new design was based on the one proposed by EasyAdmin to also replace Pagerfanta (https://github.com/EasyCorp/EasyAdminBundle/pull/2777).

### DIFF
--- a/src/Pagination/Paginator.php
+++ b/src/Pagination/Paginator.php
@@ -1,0 +1,110 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Pagination;
+
+use Doctrine\ORM\QueryBuilder as DoctrineQueryBuilder;
+use Doctrine\ORM\Tools\Pagination\CountWalker;
+use Doctrine\ORM\Tools\Pagination\Paginator as DoctrinePaginator;
+
+/**
+ * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ */
+class Paginator
+{
+    private const PAGE_SIZE = 10;
+    private $queryBuilder;
+    private $currentPage;
+    private $pageSize;
+    private $results;
+    private $numResults;
+
+    public function __construct(DoctrineQueryBuilder $queryBuilder, int $pageSize = self::PAGE_SIZE)
+    {
+        $this->queryBuilder = $queryBuilder;
+        $this->pageSize = $pageSize;
+    }
+
+    public function paginate(int $page = 1): self
+    {
+        $this->currentPage = max(1, $page);
+        $firstResult = ($this->currentPage - 1) * $this->pageSize;
+
+        $query = $this->queryBuilder
+            ->setFirstResult($firstResult)
+            ->setMaxResults($this->pageSize)
+            ->getQuery();
+
+        if (0 === \count($this->queryBuilder->getDQLPart('join'))) {
+            $query->setHint(CountWalker::HINT_DISTINCT, false);
+        }
+
+        $paginator = new DoctrinePaginator($query, true);
+
+        $useOutputWalkers = \count($this->queryBuilder->getDQLPart('having') ?: []) > 0;
+        $paginator->setUseOutputWalkers($useOutputWalkers);
+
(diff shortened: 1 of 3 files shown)


src/Pagination/Paginator.php
```diff
@@ -0,0 +1,100 @@
...
+        $query = $queryBuilder
+            ->setFirstResult($firstResult)
+            ->setMaxResults($pageSize)
+            ->getQuery();
+
+        if (0 === \count($queryBuilder->getDQLPart('join'))) {
+            $query->setHint(CountWalker::HINT_DISTINCT, false);
+        }
+
+        $paginator = new DoctrinePaginator($query, true);
```
stof: this second argument should be configurable. It is quite frequent to have paginated queries which don't fetch join a collection. and the pagination can be more efficient in that case (using a single query rather than 2)

src/Pagination/Paginator.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        return $this->numResults > $this->pageSize;
+    }
+
+    public function getNumResults(): int
+    {
+        return $this->numResults;
+    }
+
+    public function getResults(): ?\Traversable
```
stof: this cannot return `null` AFAICT.

src/Pagination/Paginator.php
stof: btw, if there is no join at all in the query (detected above), we are 100% sure that there is no fetch join of a collection (not all joins are collection fetch joins, so it is still worth it to allow configuring it for other cases)

src/Pagination/Paginator.php
```diff
@@ -0,0 +1,100 @@
...
+            ->setFirstResult($firstResult)
+            ->setMaxResults($pageSize)
+            ->getQuery();
+
+        if (0 === \count($queryBuilder->getDQLPart('join'))) {
+            $query->setHint(CountWalker::HINT_DISTINCT, false);
+        }
+
+        $paginator = new DoctrinePaginator($query, true);
+        $paginator->setUseOutputWalkers(false);
```
stof: This should detect whether the QueryBuilder has a `having` clause, as the tree walker cannot count those (the output worker is necessary in such case).
And it should also detect whether the root entity uses an association as identifier (in case we paginate with a fetch join condition), as it is not supported either by the tree walker.

src/Pagination/Paginator.php
javiereguiluz: I'd prefer to not make this configurable. The difference between making 1 or 2 queries should not be noticeable, mostly because the extra query will be very simple.

If not doing this introduces some bugs or problems, then I'm willing to make it configurable. Thanks.

src/Pagination/Paginator.php
javiereguiluz: Thanks for this review and for the explanation. Sadly this is way above my knowledge of Doctrine ... so I'd need the exact needed changes in the code to add them. Thanks.

src/Pagination/Paginator.php
bocharsky-bw: Agree, it should be `getResults(): \Traversable`. If `$this->results` may be null - I'd prefer to convert it to an empty array instead of null

mazarini:
Hi, 
Do you think it's possible to make it as a bundle (ie demo-bundle).  It can introduice the way to tell how to make and test a bundle with bests practices.

Thank for symfony/demo. I learn so many with it.

src/Pagination/Paginator.php
bocharsky-bw: Hm, IIRC we use this paginator in one spot only... if so - I don't think it should be configured unless it needed IMHO

imiroslavov: @javiereguiluz As we spoke extending doctrine's paginator is not an option. Maybe exposing it with a getter and leaving the developer to decide whether to turn this on/off is a better idea?

yceruto: even if this is a demo and there is no `having` clause currently in the code, I think we should make the adjustment according to Stof's comment, so anyone copying and pasting from here would have a correct version that covers this case.

Something like this could work:
```php
$paginator->setUseOutputWalkers(\count($this->queryBuilder->getDQLPart('having')) > 0);
```

PR #1007, opened 2019-07-03
Add symfony 4.4 and 5.0 to .travis.yml

mazarini:
Add 2 tests with failures allowed :
For this 2 adds only :
- change memory_limit of PHP,
- add 'composer install' a  for 4.4 and 5.0 only
- change stability to dev
- change 'composer install' to 'composer update'
- suppress last (composer.json change)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
   global:
     - SYMFONY_PHPUNIT_DIR=./bin/.phpunit
     - SYMFONY_DEPRECATIONS_HELPER=29
+    - ACTION="install"
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,15 +22,36 @@ matrix:
...
   - '[[ "$TRAVIS_PHP_VERSION" == "7.4snapshot" ]] || phpenv config-rm xdebug.ini'
   - composer self-update
+  # Set memory to max (memory fail)
+  - '[[ "$ACTION" == "install" ]] || echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini'
+  # Install before update because of memory usage
+  - '[[ "$ACTION" == "install" ]] || composer install'
+  # Set stability to dev to allow 4.4dev and 5.0dev
+  - '[[ "$ACTION" == "install" ]] || composer config minimum-stability dev'
+  # Choose version of symfony to use
+  - composer config extra.symfony.require $SYMFONY
```
bocharsky-bw: I suppose we want to execute it for not "install" as well, i.e. `[[ "$ACTION" == "install" ]] || composer config extra.symfony.require $SYMFONY `... and then probably we don't need global `SYMFONY="4.3.*"` at all for simplicity

mazarini: Yes, "composer config extra.symfony.require $SYMFONY" replace "4.3.*" by "4.3.*" for update  
It's just because :
- don't like  to use condition 
- prefere that all var are define for each case even if not use

Simplicity can be to have $ACTION and $SYMFONY always define. I had $STABILITY=dev (or stable) in the same spirit. I replace it by a condition because it works without it.

bocharsky-bw: Yeah, I understand it works, but it may just mislead other devs :/ My point is: if there's no effect - why execute those lines? Right now we would need to keep in sync both: `extra.symfony.require` in `composer.json` and global `$SYMFONY` in `.travis.yaml`, otherwise, it will put the repo in a dirty state on TravisCI. Yeah, probably still not a big deal, but why not simplify things and make it clearer for devs?

mazarini: Make sense : "Right now we would need to keep in sync both"
1) suppress global $SYMFONY
2) add condition on $ACTION 

So I have just to propose a new PR ?

bocharsky-bw: If we agree on it (probably good to wait for more opinions) - better to do changes in this PR  I think

javiereguiluz: I don't have a strong opinion about this ... but regarding the mention to the "extra.symfony.require in composer.json", we can't remove that. It's a "best practice" of Symfony and want to promote it.

mazarini: extra.symfony.require in composer.json only change depending on symfony release to test. The discution was just to not replace 4.3 by 4.3 who has no utility.

javiereguiluz:
@mazarini a very nice first contribution to this project! Thanks a lot for your help.

PR #1008, opened 2019-07-05
Be more specific in linting twig and yaml files

chr-hertel:
I would argue that adding those parameters ensures linting that is more similar to the framework runtime in production:
symfony will parse tags in config yaml files and twig runtime might differ between dev and prod environment.

for example we would catch a forgotten `{{ dump() }}` in a twig template that way

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,9 +36,9 @@ script:
   # this checks that the source code follows the Symfony Code Syntax rules
   - '[[ "$TRAVIS_PHP_VERSION" == "7.4snapshot" ]] || ./vendor/bin/php-cs-fixer fix --diff --dry-run -v'
   # this checks that the YAML config files contain no syntax errors
-  - ./bin/console lint:yaml config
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -36,9 +36,9 @@ script:
   # this checks that the source code follows the Symfony Code Syntax rules
   - '[[ "$TRAVIS_PHP_VERSION" == "7.4snapshot" ]] || ./vendor/bin/php-cs-fixer fix --diff --dry-run -v'
   # this checks that the YAML config files contain no syntax errors
-  - ./bin/console lint:yaml config
+  - ./bin/console lint:yaml config --parse-tags
```
bocharsky-bw: Probably lint Yaml in with `--env=prod` as well?

chr-hertel: there is no difference between `dev` or `prod` in this case

javiereguiluz:
Merged! Thank you.

PR #1010, opened 2019-07-07
Update license year

Arman-Hosseini:


javiereguiluz:
Good catch, thanks @Arman-Hosseini.

PR #1012, opened 2019-07-09
Add TravisCI build status badge

bocharsky-bw:
What about to add build status badge just to make it easier to notice when tests are failed in `master`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ Symfony Demo Application
 The "Symfony Demo Application" is a reference application created to show how
 to develop applications following the [Symfony Best Practices][1].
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
I prefer to not add this. First, because Travis alerts us when test don't pass, so we don't merge those PRs. Second, even if we ignore Travis, the proprietary tool we use to merge things displays a big red warning message when any GitHub requirement don't pass (e.g. Travis tests).

Besides, I don't think any of the active contributors of this repository sees the README of the project, so we won't see the warning. Finally, tests pass almost always on this repo ... so I don't think it's OK to show "tests failing" during the short periods of time that tests don't pass. Also, that may make some people think that this project is not a high quality one, which wouldn't be fair or true.

Thanks for understanding!

bocharsky-bw:
Thank you for explaining your point of view

PR #1015, opened 2019-07-23
Fix pagination class

ogizanagi:


javiereguiluz:
Good catch, thanks Maxime.

Issue #1018, opened 2019-08-03
[Slugger] blog_index stops working if the title contains certain characters


o-alquimista:
When you write a post with the following title: `/test`, you get an error when accessing the blog_index route:

```
An exception has been thrown during the rendering of a template ("Parameter "slug" for route "blog_post" must match "[^/]++" ("/test" given) to generate a corresponding URL.").
```

The slug `\test` (backward slash) doesn't break it.

I use the same slugger on my project and it breaks both under the same conditions. I'll be investigating this too, and share back any developments.

**EDIT:**
Oops, I just learned that adding an opening chevron `<` at the beginning of the title breaks blog_index too. I suspect more characters may also cause this problem. This may require a whitelist-like regex to solve.

o-alquimista:
At first I thought of using `stripslashes()`, but it only removes backward slashes. To remove the forward slashes from the slug, you have to use `str_replace()`.

This is how I fixed it:

```
    public static function slugify(string $string): string
    {
        $slug = preg_replace('/\s+/', '-', mb_strtolower(trim(strip_tags($string)), 'UTF-8'));
        $slug = str_replace("/", "", $slug);

        return $slug;
    }
```

The `preg_replace()` remains the same. I simply passed its result to the `str_replace()` and returned it. The app now works as expected.

o-alquimista:
This is my solution for all forbidden characters:
```
    public static function slugify(string $string): string
    {
        $slug = preg_replace('/\s+/', '-', mb_strtolower(trim($string), 'UTF-8'));
        $slug = preg_replace('/[^a-z0-9-]/', '', $slug);

        return $slug;
    }
```

The function strip_tags() was removed because, [as the PHP manual says](https://www.php.net/manual/en/function.strip-tags.php#refsect1-function.strip-tags-notes):

> Warning: Because strip_tags() does not actually validate the HTML, partial or broken tags can result in the removal of more text/data than expected.

Whenever we used a title such as `<test`, strip_tags() would remove the entire string in front of it, not just the chevron. An empty slug is then rejected.

So now the slug can only contain a-z, 0-9 and - (hyphen).

Are there any security concerns from removing strip_tags() in this case?

jkufner:
[...]

PR #1019, opened 2019-08-03
Prevent blog_index from breaking because of forbidden characters in the slug

o-alquimista:
This was explained on the issue [1018](https://github.com/symfony/demo/issues/1018).

If someone writes a title `/test`, the resulting slug will be `/test`, which will trigger an error:
```
An exception has been thrown during the rendering of a template ("Parameter "slug" for route "blog_post" must match "[^/]++" ("/test" given) to generate a corresponding URL.").
```
Nothing is currently removing forward slashes for the slug. That is not the case for backward slashes `\`, though.

Now if someone writes a title named `<this is a test`, the resulting slug will be empty, and the same exception will be raised when you access blog_index.

This happens because `strip_tags()` is removing everything after the opening tag. The page on the PHP Manual about this function warns about that.

This solution removes `strip_tags()` from the Slugger and uses a regex to catch forbidden characters. The resulting slugs look good thanks to @jkufner. A title `hello :) world-` will result in the slug `hello-world`, for example.

### DIFF
--- a/src/Utils/Slugger.php
+++ b/src/Utils/Slugger.php
@@ -14,6 +14,7 @@ namespace App\Utils;
 /**
  * @author Ryan Weaver <weaverryan@gmail.com>
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
+ * @author Douglas Silva <doug.hs@protonmail.ch>
  */
 class Slugger
 {
--- a/src/Utils/Slugger.php
+++ b/src/Utils/Slugger.php
@@ -15,14 +15,14 @@ namespace App\Utils;
  * @author Ryan Weaver <weaverryan@gmail.com>
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  * @author Douglas Silva <doug.hs@protonmail.ch>
+ * @author Josef Kufner <???>
  */
 class Slugger
 {
     public static function slugify(string $string): string
     {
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
Tests failed as you can see. IIRC, the original idea was to support not only `a-z0-9` characters

javiereguiluz:
Yes, we'd like to keep supporting non-ASCII characters too because this is a highly multi-lingual app, so we want to allow anything that is a valid character. 

Maybe we can use `\p{L}` and `\p{N}` in the regexp: https://www.php.net/manual/en/regexp.reference.unicode.php

stof:
I suggest writing unit tests for the slugger to show the different cases of the slugger (and to showcase unit tests, as this is a demo)

javiereguiluz:
We already have unit tests for the Slugger: https://github.com/symfony/demo/blob/master/tests/Utils/SluggerTest.php  But you are right that we may consider if we should add more slugs to the data provider.

stof:
Then they should indeed be updated to cover these cases

bocharsky-bw:
Probably not relevant anymore, as `Symfony\Component\String\Slugger\SluggerInterface` was used instead

javiereguiluz:
Yes, you are right! Let's close this one. Thanks!

PR #1024, opened 2019-08-18
Travis ci

mazarini:
1) Suppress a "composer install" in .travis.yml
No more need "composer install" + "composer update" since composer 1.9.

2) Update "composer.lock" to use new doctrine/annotation and solve "console cache:clear"  problem with php 7.4.


javiereguiluz:
@mazarini thanks for improving this!

Issue #1026, opened 2019-09-11
Translation broken, ext-intl or symfony/polyfill-intl-messageformatter should be required in composer


vjandrea:
New installation using composer create-project, when I open the demo in the browser I get:

> An exception has been thrown during the rendering of a template ("Cannot parse message translation: please install the "intl" PHP extension or the "symfony/polyfill-intl-messageformatter" package.").

Please add a requirement for ext-intl or symfony/polyfill-intl-messageformatter in composer.json. If you want I can submit a PR.

stof:
We should add a requirement on the polyfill, so that the demo keeps working without ext-intl

PR #1031, opened 2019-09-20
require symfony/polyfill-intl-messageformatter

dmaicher:
Fixes https://github.com/symfony/demo/issues/1026


javiereguiluz:
Thanks for fixing this bug David.

PR #1035, opened 2019-10-18
update new repo name

cybernet:


javiereguiluz:
@cybernet it's true that we change the repo name in GitHub (it's github.com/symfony/demo instead of the previous github.com/symfony/symfony-demo).

However, the Composer package can't be changed because Packagist doesn't allow to update package names. See https://packagist.org/packages/symfony/symfony-demo  To "rename" a package, you must abandon it and create a new one with the new name. So, we prefer not to do that. Thanks!

PR #1038, opened 2019-10-26
Make paginator strings translatable

davialexandre:
All but the `pt_BR` translations are based on the translations available on the KnpPaginatorBundle. That bundle doesn't have any string equivalent to "current", so this is why this translation is missing in all languages, except for `pt_BR` and `en`.


javiereguiluz:
Thank you @davialexandre.

PR #1046, opened 2019-11-24
Fix demo password autofill on Firefox

bocharsky-bw:
This PR fixes demo password auto-filling on Firefox, steps to reproduce:
1. Open the login page in FF: https://localhost:8000/en/login
2. Reload the page and see empty password - looks like FF clears it due to security reasons but keeps the username


javiereguiluz:
I wasn't aware of this! What Firefox does makes a lot of sense ... but this is a Demo app, so it's very special and I like the trick you used to keep this behavior working. Thanks!

PR #1052, opened 2019-12-02
Update README.md

noniagriconomie:
According to the new composer https://github.com/symfony/demo/blob/master/composer.json#L8

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ to develop applications following the [Symfony Best Practices][1].
 Requirements
 ------------
 
-  * PHP 7.1.3 or higher;
+  * PHP 7.2.9 or higher;
   * PDO-SQLite PHP extension enabled;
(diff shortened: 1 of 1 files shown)


noniagriconomie:
For maintainers, what about the version of symfony main repo and skeleton repo ´s composer? Should we open same PR ?
Thanks

javiereguiluz:
@noniagriconomie thanks for fixing this!

Please note that this is a bit tricky:

1) Symfony 5 requires 7.2.5
2) A few days ago, because of some mistake that it's now fixed, required 7.2.9
3) Twig 3 requires 7.2.9

PHP 7.2.5 is an important version for some LTS Linux distribution, so Symfony 5 will stick to 7.2.5 ... and Twig requirement may be a mistake that could be fixed soon ... or maybe it's OK. Meanwhile, let's merge this PR. Thanks!

PR #1054, opened 2019-12-11
Update README.md, fix command running php server

berezuev:
According to [this](https://github.com/symfony/recipes/pull/353/files) and [this](symfony/symfony#25911) issues if we start php server with -t key, instead of passing exact file, urls with dots wouldn't been recognized by routing

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ $ cd my_project/
 $ symfony serve
 ```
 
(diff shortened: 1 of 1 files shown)


berezuev:
upd: added both -t public/ and public/index.php to serve static files and dots in urls

PR #1055, opened 2019-12-19
Use the String component

javiereguiluz:


src/Twig/SourceCodeExtension.php
```diff
@@ -117,20 +118,20 @@ private function getTemplateSource($template): array
...
     private function unindentCode(string $code): string
     {
-        $formattedCode = $code;
-        $codeLines = explode("\n", $code);
+        $codeLines = u($code)->split("\n");
 
-        $indentedLines = array_filter($codeLines, function ($lineOfCode) {
-            return '' === $lineOfCode || 0 === mb_strpos($lineOfCode, '    ');
+        $indentedOrBlankLines = array_filter($codeLines, function ($lineOfCode) {
+            return u($lineOfCode)->isEmpty() || u($lineOfCode)->startsWith('    ');
```
stof: you should probably instantiate only one string object and reuse it.

javiereguiluz: For an application where performance is critical, I'd agree ... but this is only used when checking the source code explicitly, so it rarely happens. Let's keep the code a bit simpler. Thanks.

.travis.yml
```diff
@@ -10,7 +10,7 @@ cache:
 env:
   global:
     - SYMFONY_PHPUNIT_DIR=./bin/.phpunit
-    - SYMFONY_DEPRECATIONS_HELPER=0
+    - SYMFONY_DEPRECATIONS_HELPER=9
```
COil: Hi, @javiereguiluz  What is 9 ? :thinking:

javiereguiluz: They are the Doctrine deprecations that I couldn't remove and prevented tests from being green. See https://travis-ci.org/symfony/demo/jobs/627625205#L358

PR #1057, opened 2019-12-20
Modal bootstrap reference URL updated

shahariaazam:
Reference URL has been updated


templates/blog/_delete_post_confirmation.html.twig
```diff
@@ -1,4 +1,4 @@
-{# Bootstrap modal, see http://getbootstrap.com/javascript/#modals #}
+{# Bootstrap modal, see https://getbootstrap.com/2.3.2/javascript.html#modals #}
```
noniagriconomie: I do not think forcing the version is a good choice

javiereguiluz: I agree. Let's wait until we merge the PR which updates this application to Bootstrap 4. Thanks!

@shahariaazam even if we didn't merge it for the reasons said above, thanks for creating this pull request.

PR #1058, opened 2019-12-22
Use new functional test specific assertions

voronkovich:
See https://symfony.com/doc/current/testing/functional_tests_assertions.html


javiereguiluz:
This was long overdue! Thanks for taking care of this, Oleg.

PR #1059, opened 2019-12-22
Make tests compatible with PHPUnit 8

voronkovich:
`setUp()` method must have a `void` return type declaration. See: https://phpunit.de/announcements/phpunit-8.html


javiereguiluz:
Thanks Oleg.

PR #1060, opened 2020-01-01
Fix broken links

voronkovich:
The Symfony's documentation was greatly updated and many links in the app's source files are broken. I found all broken links and fixed each one of them where it was possible.

I used this command to extract the links from the source code, so I could miss out some of them:
```sh
grep -ERhoe 'https?://[^[:space:]]*\w' src assets config tests templates | sort -u
```


noniagriconomie:
@voronkovich nice one 👍 

just an idea, can it be usefull to "automate" this? on this repo and all sf repo (I saw https://github.com/symfony/recipes/pull/715, and i am sure on /symfony and /symfony-docs as well)

a small `sh` or `php` script, launched once a week or manually for example
it founds the http(s) endpoints and requests them, then checks if != 200 (also no redirection) => stderr
so that a human can fix/improve those links :)

jkufner:
It is a good idea to incorporate these checks into CI pipeline. I've done that for my static websites and it turned up very useful.

javiereguiluz:
Oleg, this is great! Thanks a lot for fixing all these wrong links. We appreciate it a lot!

voronkovich:
@noniagriconomie, Many links contain fragments (`http://link#fragment`), so your script should check that a requested page has a corresponding element if the fragment presents in URI. To do this you could use the [symfony/browser-kit](https://symfony.com/doc/current/components/browser_kit.html).

noniagriconomie:
@voronkovich yeah just a poc I have done, still i use it in a local project :)
but yeay in sf codebase, need to excluded some urls and thise anchors urls

PR #1062, opened 2020-01-07
Make blog post titles unique

### LINKED ISSUE #1045: More article with the same title
If more than one article is created with the same title, unfortunately the link is only functional for one, even if the user is different

javiereguiluz:
I'm trying to fix #1045. 

We should make the `slug` unique because it's what we use to look for blog posts in the database. However, I can't make this work. If I define `UniqueEntity` on `title`, it works ... but if I define it on `slug`, it doesn't work (I can create any number of posts with the same slug). Anybody sees the error? Thanks!

### DIFF
--- a/src/Entity/Post.php
+++ b/src/Entity/Post.php
@@ -14,11 +14,13 @@ namespace App\Entity;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
+use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
 use Symfony\Component\Validator\Constraints as Assert;
 
 /**
  * @ORM\Entity(repositoryClass="App\Repository\PostRepository")
  * @ORM\Table(name="symfony_demo_post")
+ * @UniqueEntity(fields={"slug"}, errorPath="title")
  *
  * Defines the properties of the Post entity to represent the blog posts.
  *
--- a/src/Entity/Post.php
+++ b/src/Entity/Post.php
@@ -20,7 +20,7 @@ use Symfony\Component\Validator\Constraints as Assert;
 /**
  * @ORM\Entity(repositoryClass="App\Repository\PostRepository")
  * @ORM\Table(name="symfony_demo_post")
- * @UniqueEntity(fields={"slug"}, errorPath="title")
+ * @UniqueEntity(fields={"slug"}, errorPath="title", message="This title was already used in another blog post, but they must be unique.")
  *
  * Defines the properties of the Post entity to represent the blog posts.
  *
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -21,7 +21,6 @@ use Symfony\Component\Form\Extension\Core\Type\SubmitType;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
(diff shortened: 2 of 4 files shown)


src/Entity/Post.php
```diff
@@ -14,11 +14,13 @@
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
+use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
 use Symfony\Component\Validator\Constraints as Assert;
 
 /**
  * @ORM\Entity(repositoryClass="App\Repository\PostRepository")
  * @ORM\Table(name="symfony_demo_post")
+ * @UniqueEntity(fields={"slug"}, errorPath="title")
```
yceruto: Yes, the `slug` value is generated after `$form->isValid()` (when the validation process is fired) in `Admin\BlogController.php`.

To make it work properly we can move this statement `$post->setSlug(Slugger::slugify($post->getTitle()))` before it and after `$form->handleRequest($request)`. What do you think?

javiereguiluz: Yes! That must be the error.

However, if I move that line ... I see this error:

```
Argument 1 passed to Symfony\Component\String\Slugger\AsciiSlugger::slug() must be of the type string, null given.
```

If I change it like this:

```php
        $form->handleRequest($request);

        if ($form->isSubmitted()) {
            $post->setSlug($slugger->slug($post->getTitle())->lower());
        }

            // ...
        }
```

Then I can again create multiple blog posts with the same slug, so the validation is not working.

voronkovich: @javiereguiluz, Calling `isValid()` method doesn't trigger validation. See https://github.com/symfony/form/blob/098f87036aeae1a1f4658d5f9f154b8cca628eb9/Form.php#L739

yceruto: Sorry for the confusion, @voronkovich is right, is the `$form->handleRequest()` who fires the validation process, so we're at the beginning again :)

voronkovich: @javiereguiluz, @yceruto  I would suggest to use a form event listener:
```php
->addEventListener(FormEvents::SUBMIT, function (FormEvent $event) {
    $post = $event->getData();
    $post->setSlug($this->slugger->slug($post->getTitle())->lower());
})
```
This also will decrease code duplicity, because we won't need to generate slug in both [`new`](https://github.com/symfony/demo/blob/7cbada18fdb96b5fdabbd839a05e7740f1b95abb/src/Controller/Admin/BlogController.php#L89) and [`edit`](https://github.com/symfony/demo/blob/7cbada18fdb96b5fdabbd839a05e7740f1b95abb/src/Controller/Admin/BlogController.php#L142) controllers.

yceruto: What about a custom validation by using the `repositoryMethod` option of the `@UniqueEntity()` annotation?

There we might have the chance to build the slug before checking that it's unique.

yceruto: This line `$post->setSlug($this->slugger->slug($post->getTitle())->lower());` is unsafe, you'll need to check before that that the title value is not null. But, that's another option.

javiereguiluz: I liked both solutions a lot. I think they are both equally correct. However, I decided to try the form event solution proposed by @voronkovich because it felt more closely related to Symfony. A repository is a Doctrine thing, not a strictly Symfony thing and that's why I think we should use form events in this case. But I'm open to hear other opinions and change this. Thanks!

voronkovich:
@javiereguiluz, should we add a functional test for this feature?

PR #1063, opened 2020-01-07
Replaced SwiftMailer by Symfony Mailer

### LINKED ISSUE #1036: Replace Swiftmailer with Mailer Component
Currently this project use Swiftmailer to send email, for now only when a comment is created (however, delivery is disabled by default).

As a demo application to demonstrate Symfony capabilities, I think it will make sense to switch to the Mailer Component introduce in 4.3. What do you think ?

javiereguiluz:
This fixes #1036.


tests/Command/AddUserCommandTest.php
```diff
@@ -31,7 +31,7 @@ protected function setUp()
         exec('stty 2>&1', $output, $exitcode);
         $isSttySupported = 0 === $exitcode;
 
-        if ('Windows' === \PHP_OS_FAMILY || !$isSttySupported) {
+        if ('Windows' === PHP_OS_FAMILY || !$isSttySupported) {
```
juanmiguelbesada: just curious. Why this?

javiereguiluz: It's unrelated to this pull request ... but without this change, the php-cs-fixer linter shows an error. I was lazy and did the change here instead of creating a new PR 🙈

PR #1064, opened 2020-01-07
Use short syntax for routes HTTP methods

voronkovich:
BTW, short syntax is used in `bin/console debug:router` command's output.


javiereguiluz:
This looks a bit better. I like it! Thanks Oleg.

PR #1066, opened 2020-01-12
Remove "symfony/polyfill-php72" from the dependencies

rosier:
PHP version 7.2+ is already required by Symfony 5

PR #1067, opened 2020-01-13
Validate composer files while TravisCI build

voronkovich:
This check prevents us from having an out of sync composer's lock-file (For example, when someone forgot to commit composer.lock file: https://github.com/symfony/demo/pull/1066).


.travis.yml
```diff
@@ -25,6 +25,7 @@ matrix:
 before_install:
   - '[[ "$TRAVIS_PHP_VERSION" == "7.4snapshot" ]] || phpenv config-rm xdebug.ini'
   - composer self-update
+  - composer validate
```
stof: please move this to the `script` section, so that it is one of the tests and appear that way.

javiereguiluz:
Thank you Oleg.

Issue #1069, opened 2020-01-21
It works with built-in server but getting 404 errors on apache virtual host


joaopmmartins:
On the Symphony's built-in server, the demo works perfectly well but when I try with Apache virtual host, I can only view the homepage. For every other page, I am getting a 404 error. 

```
<VirtualHost *:80>
	ServerAdmin webmaster@localhost
	ServerName my-local-site
	ServerAlias my-local-site
   	DocumentRoot /home/websites/my-local-site/public
	<Directory "/home/websites/my-local-site/public">
             Options Indexes FollowSymLinks MultiViews
             AllowOverride All
            Require all granted
        </Directory>
	CustomLog "|/usr/bin/rotatelogs -l /var/log/apache2/my-local-site-access.%Y-%m-%d 86400" common
	ErrorLog "|/usr/bin/rotatelogs /var/log/apache2/my-local-site-error.%Y-%m-%d 5M"
</VirtualHost>
```

Can you please point me in the right direction? Many thanks.

bocharsky-bw:
Try to install `symfony/apache-pack` package: https://github.com/symfony/apache-pack -  it containes necessary .htaccess config for public/ dir

javiereguiluz:
Closing as fixed because in the next stable version of this project we'll include the Apache package to make it work out-of-the-box. Thanks!

Issue #1005, opened 2019-06-30
The project not work


adrix71:
Hello,

I tried this project but it doesn't seem intent on working. My Apache web server is configured correctly because I tried the website_skeleton application (and many other project on my server test) and works.

How can I solve it? 

This is a screen.

![Schermata del 2019-06-30 21-37-46](https://user-images.githubusercontent.com/14064365/60401294-66d68f00-9b7f-11e9-98af-8c90e21661dc.png)

bocharsky-bw:
Try to execute:
```bash
composer req apache-pack
```

And try again, please. It's required for running this project on Apache

adrix71:
It's my fault. I apologize.

I read that the project is without additional configuration and I didn't think to read the documentation about how to use symfony with apache.

Now works. Thank you so much and sorry.



> Try to execute:
> 
> ```shell
> composer req apache-pack
> ```
> 
> And try again, please. It's required for running this project on Apache

It's my fault. I apologize.

I read that the project is without additional configuration and I didn't think to read the documentation about how to use symfony with apache.

Now works. Thank you so much and sorry.

javiereguiluz:
@adrix71 Sorry for the confusion! The project requires zero-setup or configuration if you follow the recommended installation/execution process shown in the README. But, if you change the database, or the web server, etc. you'll need to make the needed changes. Luckily the solution proposed by @bocharsky-bw was easy. Closing as fixed then. Thanks!

Issue #711, opened 2017-11-28
No .htaccess


dzuelke:
There is currently no `.htaccess` file in place with rewrites that remove `index.php`, so out of the box, the demo doesn't work beyond the first page. Is that intentional?

jkufner:
Commiting .htaccess is a bad practice, because it complicates deploy. However, it would be nice to have example config files for the most used web servers in the documentation.

piotrgradzinski:
@dzuelke - as per @jkufner comment this is considered as a bad practise. 

Check [Configuring a Web Server](http://symfony.com/doc/current/setup/web_server_configuration.html) documentation section for an explanation and information on how you should configure your web server. 

Nevertheless you can use `apache-pack` recipe for adding `.htaccess` to you `public` folder by executing 

`composer require apache-pack`.

Hope this helps and answers you concern.

bocharsky-bw:
Actually a good point. As far as we have a "Deploy to Heroku" button in README, we should have a proper `.htaccess` for it as well to make it working out of the box

dzuelke:
Yeah I'm mostly interested in having it work out of the box on as many setups as possible, as it's a demo app. I could easily do a PR with a Heroku specific config or something, but would rather not clutter the repo with a vendor specific thing (the demo right now is not working on Heroku, for other reasons as well, such as SensioDistributionBundle being gone which used to write the `Procfile`).

javiereguiluz:
It's true that we don't care about `.htaccess` in this demo app because we recommend to use the Symfony server (`bin/console server:start` command) to browse the app.

However, I agree that this should work out-of-the-box on Heroku (unless that requires us to change our code). We already have an `app.json` with Heroku config and stuff (https://github.com/symfony/demo/blob/master/app.json). Could we make more changes in that file to fix this issue? Thanks!

dzuelke:
For that we'd need a separate `Procfile`, but I think I'll add support for automatic detection of Symfony to Heroku's PHP support. For that I need a definitive answer to https://github.com/symfony/symfony-docs/issues/8766 though.

javiereguiluz:
Let's close this in favor of the automatic detection feature on Heroku proposed by David ... and we'll reopen this if that's not possible to implement for any reason.

PR #1070, opened 2020-01-22
Require symfony/apache-pack package

bocharsky-bw:
What about to add `symfony/apache-pack` to this project? We have had a few related issues about missing it: #1069, #1005, #711 . Basically, it's a really small package that just adds `.htaccess` file into public/ dir and that's it. It would help Apache users and does not do any harm to others I think, so why not have it out of the box?


javiereguiluz:
Thanks Victor!  You explained it well: in other circumstances we wouldn't add this package ... but this app is very special and people have already told us several times "we need this". So, let's add it. Thanks.

PR #1072, opened 2020-01-24
Add serialize-javascript dependency

msalsas:
Add serialize-javascript dependency to avoid regular expressions Cross-Site Scripting (XSS) vulnerability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,5 @@
     "dev": "encore dev",
     "watch": "encore dev --watch",
     "build": "encore production --progress"
-  },
-  "dependencies": {
-    "serialize-javascript": "^2.1.2"
   }
 }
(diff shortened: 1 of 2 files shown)


msalsas:
You're right, @skmedix, sorry. I've just fixed it.

javiereguiluz:
I'm afraid that I don't understand this change. Is this simple change enough? You don't have to do anything in the application code to improve things? Please, explain more in detail which is the problem that this change is solving. Thanks!

javiereguiluz:
We're closing this because of the lack of feedback. Thanks!

stof:
this does not solve anything. Adding an unused devDependency in the project does not impact the version used by a transitive dependency. I guess that this was a misunderstanding of how npm/yarn works (applying the composer knowledge to them).

PR #1074, opened 2020-01-29
Use a more friendly message about hardcoding user details

pierredup:


javiereguiluz:
This is much more friendly :)  Thanks a lot Pierre!

PR #1080, opened 2020-02-12
Update markdown

abdounikarim:
Hello,

I remove the Markdown class which was used to convert markdown to html. 
I used the [markdown_to_html](https://twig.symfony.com/doc/2.x/filters/markdown_to_html.html) filter instead. 

I run the tests after and hey still works.


javiereguiluz:
I like this! Thanks @abdounikarim.

However, we need to revert the changes related to `tgalopin/html-sanitizer-bundle` because we want to keep that bundle and its features. Thanks!

javiereguiluz:
I see now why you removed the HTML sanitizer bundle ... but instead of that ... we may add a new Twig filter to do this -> `... | sanitize_html`   I don't know which could be the best option here ... maybe @tgalopin can suggest us something. Thanks!

javiereguiluz:
After talking with Titouan, he told me that the Sanitizer bundle already provides a `sanitize_html` Twig filter ... so we may try to do the following:

* Reinstall and reenable the HtmlSanitizer bundle as before
* Tweak the needed Twig templates to add the `|sanitize_html` filter after the Markdown filter

abdounikarim:
> After talking with Titouan, he told me that the Sanitizer bundle already provides a `sanitize_html` Twig filter ... so we may try to do the following:
> 
> * Reinstall and reenable the HtmlSanitizer bundle as before
> * Tweak the needed Twig templates to add the `|sanitize_html` filter after the Markdown filter

@javiereguiluz done

src/Twig/AppExtension.php
```diff
@@ -11,15 +11,12 @@
...
-use App\Utils\Markdown;
 use Symfony\Component\Intl\Locales;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFilter;
 use Twig\TwigFunction;
 
 /**
- * This Twig extension adds a new 'md2html' filter to easily transform Markdown
- * contents into HTML contents inside Twig templates.
  *
```
tgalopin: This empty line could probably be removed.

config/packages/html_sanitizer.yaml
```diff
@@ -6,10 +6,12 @@ html_sanitizer:
             # to learn more about which extensions you would like to enable.
             extensions:
                 - 'basic'
-                - 'list'
-                - 'table'
-                - 'image'
-                - 'code'
+                # - 'list'
```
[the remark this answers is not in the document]
abdounikarim: @tgalopin i restore the file like it was initially 😉

src/Twig/AppExtension.php
abdounikarim: done

javiereguiluz:
Thanks for this contribution Karim!

PR #1082, opened 2020-02-14
Update BlogControllerTest.php

DurandSacha:
Adding a verification, does the index send an HTTP 200 code ?

### DIFF
--- a/tests/Controller/BlogControllerTest.php
+++ b/tests/Controller/BlogControllerTest.php
@@ -36,6 +36,7 @@ class BlogControllerTest extends WebTestCase
             $crawler->filter('article.post'),
             'The homepage displays the right number of posts.'
         );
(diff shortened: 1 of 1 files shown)


tests/Controller/BlogControllerTest.php
```diff
@@ -36,6 +36,8 @@ public function testIndex()
             $crawler->filter('article.post'),
             'The homepage displays the right number of posts.'
         );
+        
+        $this->assertEquals(200, $client->getResponse()->getStatusCode());
```
abdounikarim: Hello @DurandSacha, can you use the following code instead : 
`$this->assertResponseIsSuccessful()`

DurandSacha: thank you dear friend

javiereguiluz:
Thanks Sacha.

PR #1087, opened 2020-02-26
use phpunit 8

dmaicher:
PHPUnit 7 is not maintained anymore so we should update to version 8 :blush: 

See https://phpunit.de/supported-versions.html

### DIFF
--- a/tests/Utils/ValidatorTest.php
+++ b/tests/Utils/ValidatorTest.php
@@ -16,89 +16,87 @@ use PHPUnit\Framework\TestCase;
 
 class ValidatorTest extends TestCase
 {
-    private $object;
+    private $validator;
 
-    public function __construct()
+    protected function setUp(): void
     {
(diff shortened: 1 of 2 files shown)


tests/Utils/ValidatorTest.php
```diff
@@ -16,89 +16,87 @@
 
 class ValidatorTest extends TestCase
 {
-    private $object;
+    private $validator;
 
-    public function __construct()
+    protected function setUp(): void
     {
-        parent::__construct();
```
dmaicher: caused

```
Other deprecation notices (1)

  1x: The "PHPUnit\Framework\TestCase::__construct()" method is considered internal This method is not covered by the backward compatibility promise for PHPUnit. It may change without further notice. You should not extend it from "App\Tests\Utils\ValidatorTest".

javiereguiluz:
Thank you David.

PR #1089, opened 2020-03-08
Replace twig and twig-bundle with twig-pack

ker0x:
This PR replace `twig/twig` and `symfony/twig-bundle` with `symfony/twig-pack` which also install `twig/extra-bundle` that is require since #1080 and the addition of `Twig\Extra\TwigExtraBundle\TwigExtraBundle::class` in `bundles.php`

It also bump symfony packages to `v5.0.5`


javiereguiluz:
Thank you Romain.

PR #1091, opened 2020-03-23
Update PHPDoc Comment

cybernet:


javiereguiluz:
@cybernet thanks for your contribution! Sadly, we can't merge it. In Symfony we try to not add PHPDoc comments for parameters which have type-hints and for return types. The reason is that information would be duplicated. Most IDEs can get the needed info from type-hints and return-types, so PHPdoc is not needed in those cases.

PR #1094, opened 2020-04-03
Add minimalist docker integration

ker0x:
This PR add support for Docker with a minimalist configuration. It take advantage offer by the Symfony binary and it's [docker integration](https://symfony.com/doc/current/setup/symfony_server.html#docker-integration)

The application still use the SQLite DB by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,31 @@ $ cd my_project/
 $ ./bin/phpunit
 ```
 
+Docker
+------
+
+To run the "Symfony Demo Application" with Docker, [Download Docker][5] and [Docker Compose][6] and install them.
+
+At the root of the project, create an `.env.local` file and add the following line:
+
+```yaml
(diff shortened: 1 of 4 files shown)


voronkovich:
I would suggest to add MailHog for testing emails: https://github.com/mailhog/MailHog

README.md
```diff
@@ -53,7 +53,31 @@ $ cd my_project/
...
+DATABASE_URL=postgresql://127.0.0.1:5432/db?serverVersion=11&charset=utf8
+```
+
+Finally, run those commands:
+
+```bash
+$ cd my_project/
+$ docker-compose up -d
+$ symfony console doctrine:database:create
+$ symfony console doctrine:migrations:migrate
```
bocharsky-bw, line 73: I'd simplify with `doctrine:schema:create` instead, otherwise on each schema change we would need a new migration, but it's not necessary because we use SQLite here

PR #1100, opened 2020-04-24
Add Symfony Cast's Reset Password Bundle to demo

jrushlow:
Adds the `symfonycasts/reset-password-bundle` to the demo app with the generated `make:reset-password` files. 

The `ResetPasswordController`, `check_email.html.twig`, & `request.html.twig` that are generated by the maker were slightly modified to allow demonstration of the reset password feature.

`symfony/maker-bundle` was updated to `v1.15.1` to introduce the `make:reset-password` command into the demo app.

We set the `throttle_limit` to `5` (seconds) in the `reset_password.yaml` config file. This allows multiple reset password requests for the demo.


stof:
The UI added in this PR does not use translations currently, while the demo app is translated.

javiereguiluz:
@stof yes, some strings need to be converted into translation keys and moved to the https://github.com/symfony/demo/blob/master/translations/messages%2Bintl-icu.en.xlf file ... but apart from that detail, what do you think of the implementation of this feature?

It's a cool feature, so we'd like to add it to the Demo app. Thanks.

jrushlow:
I was thinking about the translations last night, I'll work on getting those converted.

stof:
@weaverryan should MakerBundle provide an option to generate code using the translation component directly ?

translations/messages+intl-icu.en.xlf
```diff
@@ -438,6 +438,59 @@
...
             </trans-unit>
+
+            <trans-unit id="reset_password.button">
+                <source>reset_password.button</source>
+                <target>Reset password</target>
+            </trans-unit>
+            <trans-unit id="reset_password.request">
+                <source>reset_password.request</source>
+                <target>Enter your email address and we we will send you a link to reset your password.</target>
+            </trans-unit>
```
jrushlow, line 449: RE: naming conventions,

Would anyone object to having the translation `id`'s / `source` prefixed with `reset_password.` Personally I feel this makes it easier to maintain the translations for different parts of the app. Seems a bit easier to know where a translation is used. e.g. `reset_password.check_email.title` - translation is used in reset password check email template, and theres a pretty good chance it's the title translation for that page...

jrushlow:
I'm working on getting the formatting for the other 20 translations correct. Do we want to have those translations all in this PR, or add them separately?

javiereguiluz:
@jrushlow given that there are many contents to translate ... it's OK to leave it for a different PR. Let's focus on finishing this PR, which already does a lot of changes.

@weaverryan if you have a few minutes, could you please check this PR? I haven't seen anything wrong, but you know this feature much better than me. Thanks!

jrushlow:
I think we're solid now on having all the translations keyed in the templates / validators

translations/messages+intl-icu.en.xlf
```diff
@@ -438,6 +438,63 @@
...
+                <target>Forgot Password</target>
+            </trans-unit>
+            <trans-unit id="reset_password.title">
+                <source>reset_password.title</source>
+                <target>Reset your password</target>
+            </trans-unit>
+
+            <trans-unit id="reset_password.check_email.p1">
+                <source>reset_password.check_email.p1</source>
+                <target>An email has been sent that contains a link that you can click to reset your password. This link will expire in {token} hour(s).</target>
```
andrewmy, line 453:
```suggestion
                <target>An email has been sent that contains a link that you can click to reset your password. This link will expire in {token} {token, plural, =0 {hours} one {hour} other {hours}}.</target>
```
Am I doing this pluralization thing right?

translations/messages+intl-icu.en.xlf
```diff
@@ -438,6 +438,63 @@
...
+                <target>Password Reset Email Sent</target>
+            </trans-unit>
+
+            <trans-unit id="reset_password.email.title">
+                <source>reset_password.email.title</source>
+                <target>Hi!</target>
+            </trans-unit>
+            <trans-unit id="reset_password.email.body">
+                <source>reset_password.email.body</source>
+                <target><![CDATA[To reset your password, please visit <a href="{email_link}">here</a> This link will expire in {expires} hour(s)..]]></target>
```
andrewmy, line 478:
```suggestion
                <target><![CDATA[To reset your password, please visit <a href="{email_link}">here</a> This link will expire in {expires} {token, plural, =0 {hours} one {hour} other {hours}}.]]></target>
```

translations/messages+intl-icu.en.xlf
stof: no need for `=0`. The pluralization will already know that `0` uses `other` in English.

javiereguiluz:
Jesse, I thank your efforts here trying to add this feature to the Symfony Demo app. However, after thinking about this, I consider that this introduces a non-negligible complexity that it's out of the main goals of this project.

That's why I'm closing this without merging. Also, we can review this decision in the future. Thanks!

PR #1103, opened 2020-05-05
Fix unbound version constraints and update dependencies

rosier:
Fixes the following `composer validate` warnings:

See https://getcomposer.org/doc/04-schema.md for details on the schema
require.symfony/expression-language : unbound version constraints (*) should be avoided
require.symfony/framework-bundle : unbound version constraints (*) should be avoided


javiereguiluz:
Thank you @rosier.

PR #1107, opened 2020-05-14
Updated jQuery to 3.5.1

javiereguiluz:
Replaces https://github.com/symfony/demo/pull/1102


yarn.lock
```diff
@@ -3903,11 +3879,16 @@ isstream@~0.1.2:
   resolved "https://registry.yarnpkg.com/isstream/-/isstream-0.1.2.tgz#47e63f7af55afa6f92e1500e690eb8b8529c099a"
   integrity sha1-R+Y/evVa+m+S4VAOaQ64uFKcCZo=
 
-jquery@>=1.7, "jquery@^1.8.3 || ^2.0 || ^3.0", jquery@^3.3.1:
+jquery@>=1.7, "jquery@^1.8.3 || ^2.0 || ^3.0":
   version "3.4.1"
```
stof, line 3883: du to the way `yarn upgrade` works for partial upgrades, you end up with 2 versions fof jQuery in your build. You should force yarn to upgrade other packages depending on jQuery (use `yarn why jquery` to find them) so that they also use the new version.

javiereguiluz: Thanks for this suggestion. This is the command output:

```
$ yarn why jquery

yarn why v1.22.4
[1/4] 🤔  Why do we have the module "jquery"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...

=> Found "jquery@3.5.1"
info Has been hoisted to "jquery"
info This module exists because it's specified in "devDependencies".
info Disk size without dependencies: "1.7MB"
info Disk size with unique dependencies: "1.7MB"
info Disk size with transitive dependencies: "1.7MB"
info Number of shared dependencies: 0

=> Found "typeahead.js#jquery@3.4.1"
info This module exists because "typeahead.js" depends on it.
info Disk size without dependencies: "1.69MB"
info Disk size with unique dependencies: "1.69MB"
info Disk size with transitive dependencies: "1.69MB"
info Number of shared dependencies: 0

=> Found "eonasdan-bootstrap-datetimepicker#jquery@3.4.1"
info This module exists because "eonasdan-bootstrap-datetimepicker" depends on it.
✨  Done in 1.08s.
```

But I don't know what to do now. `typeahead.js` for example was last released five years ago (see https://www.npmjs.com/package/typeahead.js). How can we update those? Thanks!

stof: run `yarn upgrade typeahead.js`. Even though yarn won't find a new version for that package, that will still whitelist updating its dependencies.

javiereguiluz: It worked! Thanks a lot for your help!

stof: note that this yarn behavior preferring to install 2 versions than to update the dependency of other packages during partial updates is the reason why I **always** review yarn.lock changes. For frontend code, installing 2 versions is actually quite bad as it doubles the size (and in the case of jQuery plugin using `dependency` rather than `peerDependency`, it could even break their usage)

javiereguiluz: Yes, this surprised me a lot! Thanks for the insights and the tips to avoid this in the future!

PR #1108, opened 2020-05-15
Updated PHP dependencies

javiereguiluz:
This updates Symfony to 5.0.8.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -229,16 +229,16 @@
         },
         {
             "name": "doctrine/common",
-            "version": "2.12.0",
+            "version": "2.13.0",
             "source": {
                 "type": "git",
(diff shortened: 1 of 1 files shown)


composer.lock
```diff
@@ -7742,6 +6915,5 @@
     "platform-dev": [],
     "platform-overrides": {
         "php": "7.2.9"
-    },
-    "plugin-api-version": "1.1.0"
```
stof, line 7746: Looks like you use an older composer version than for the previous composer.lock change.

javiereguiluz: Yes! I've upgraded my Composer to `1.10.6`, updated again and this was restored. Thanks for reviewing this!

PR #1109, opened 2020-05-16
Explains why there is no need for fixtures

michaelKaefer:


javiereguiluz:
Thanks Michael. Adding this comment will make things easier to understand when looking at this for the first time. We reworded the message a bit while merging to make it more concise, but your contribution was great! Thanks.

PR #1113, opened 2020-05-19
Fix: Avoid unnecessary imports and aliases

localheinz:
This PR

* [x] proposes an alternative to importing every single class and using unnecessary aliases

💁‍♂️ Also see https://localheinz.com/blog/2020/05/19/avoiding-imports-and-aliases/

### DIFF
--- a/src/Entity/Post.php
+++ b/src/Entity/Post.php
@@ -11,16 +11,15 @@
 
 namespace App\Entity;
 
-use Doctrine\Common\Collections\ArrayCollection;
-use Doctrine\Common\Collections\Collection;
-use Doctrine\ORM\Mapping as ORM;
-use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
-use Symfony\Component\Validator\Constraints as Assert;
+use Doctrine\Common;
+use Doctrine\ORM;
+use Symfony\Bridge\Doctrine;
+use Symfony\Component\Validator;
 
 /**
- * @ORM\Entity(repositoryClass="App\Repository\PostRepository")
- * @ORM\Table(name="symfony_demo_post")
- * @UniqueEntity(fields={"slug"}, errorPath="title", message="post.slug_unique")
+ * @ORM\Mapping\Entity(repositoryClass="App\Repository\PostRepository")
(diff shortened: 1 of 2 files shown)


jkufner:
This really reduces readability. There is nothing wrong with a few more imports.

javiereguiluz:
@localheinz thanks for this contribution. @jkufner I agree with you ... but for others this will improve readability, so it's debatable.

However, in this repository we follow Symfony practices as close as possible ... and that includes source code formatting. Given that this differs a lot from the standard Symfony code syntax, I'm afraid we must close without merging. Thanks for understanding it!

localheinz:
@javiereguiluz 

Everything is debatable - even Symfony best practices!

😉

javiereguiluz:
Sure! That's what I tried to say in my comment. We can change Symfony practices ... but we can't do that from this repository, which only follows what Symfony does. Thanks!

src/Entity/Post.php
```diff
@@ -48,86 +47,86 @@ class Post
      *
      * @ORM\Id
      * @ORM\GeneratedValue
```
jkufner, line 49: I think you are missing `Mapping` here (and other places): `@ORM\Mapping\GeneratedValue`.

jkufner:
What I really don't like on this style is the use of multi-level namespaces in the code, e.g., `@Validator\Constraints\NotBlank`. It should be `@NotBlank` or `@Constraint\NotBlank`, i.e., use the class directly, or use an alias for the namespace.

On the other hand, I like that there are no aliases, everything is 1:1 with actual namespace names. It may be a good idea to place all annotations into an extra namespace to introduce this alias globally, for example, instead of using `use Symfony\Component\Validator\Constraints as Assert` the annotations could be placed into the `Assert` namespace — `use Symfony\Component\Validator\Constraints\Assert`. Then the annotation would look like `@Assert\NotBlank` without the need to use a custom alias everywhere.

src/Entity/Post.php
localheinz: I do - doesn't make any sense to waste my time here, since the pull request has been closed already.

PR #1116, opened 2020-05-22
fix method name

### LINKED ISSUE #1115: Bug: there's no setException method in ExceptionEvent
https://github.com/symfony/demo/blob/7e49b6414ab7d143b9aba9354f2c1c3760b38cd4/src/EventSubscriber/CheckRequirementsSubscriber.php#L88

Uncaught PHP Exception Symfony\Component\ErrorHandler\Error\UndefinedMethodError: "Attempted to call an undefined method named "setException" of class "Symfony\Component\HttpKernel\Event\ExceptionEvent"." at D:\home\site\wwwroot\src\EventSubscriber\CheckRequirementsSubscriber.php line 88

xabbuh:
fixes #1115

### DIFF
--- a/src/EventSubscriber/CheckRequirementsSubscriber.php
+++ b/src/EventSubscriber/CheckRequirementsSubscriber.php
@@ -85,7 +85,7 @@ class CheckRequirementsSubscriber implements EventSubscriberInterface
 
         // Check if SQLite is enabled
         if ($isDriverException && $this->isSQLitePlatform() && !\extension_loaded('sqlite3')) {
-            $event->setException(new \Exception('PHP extension "sqlite3" must be enabled because, by default, the Symfony Demo application uses SQLite to store its information.'));
+            $event->setThrowable(new \Exception('PHP extension "sqlite3" must be enabled because, by default, the Symfony Demo application uses SQLite to store its information.'));
         }
     }
 


src/EventSubscriber/CheckRequirementsSubscriber.php
```diff
@@ -85,7 +85,7 @@ public function handleKernelException(ExceptionEvent $event): void
 
         // Check if SQLite is enabled
         if ($isDriverException && $this->isSQLitePlatform() && !\extension_loaded('sqlite3')) {
```
brentybh, line 87: Should it be `pdo_sqlite` rather than `sqlite3`?
Doctrine DBAL is not using `sqlite3`: https://www.doctrine-project.org/projects/doctrine-dbal/en/2.10/reference/configuration.html#driver

src/EventSubscriber/CheckRequirementsSubscriber.php
```diff
@@ -85,7 +85,7 @@ public function handleKernelException(ExceptionEvent $event): void
...
-            $event->setException(new \Exception('PHP extension "sqlite3" must be enabled because, by default, the Symfony Demo application uses SQLite to store its information.'));
+            $event->setThrowable(new \Exception('PHP extension "sqlite3" must be enabled because, by default, the Symfony Demo application uses SQLite to store its information.'));
```
brentybh, line 88: Also the message

src/EventSubscriber/CheckRequirementsSubscriber.php
xabbuh: There is a difference between how Doctrine refers to a driver (`pdo_sqlite` here) and the name of the PHP extension (`sqlite3` here) that needs to be installed to be able to get here.

brentybh: Hmm... `get_loaded_extensions` do shows `pdo_sqlite`, which refers to https://www.php.net/manual/en/ref.pdo-sqlite.php

If only `sqlite3`, which refers to https://www.php.net/manual/en/book.sqlite3.php
is installed, `Uncaught PHP Exception Doctrine\DBAL\Exception\DriverException: "An exception occurred in driver: could not find driver"` is thrown, which passes the `if` check of our subscriber

xabbuh: Hm indeed, maybe we need to check for both extensions.

brentybh: The `sqlite3` (non-PDO) isn't used at all. Only `pdo_sqlite` is OK

stof: I guess this was done this way due to the confusion created by distribution packages. On Debian and Ubuntu, installing `php7.3-sqlite` (or whather the PHP version is) installs both extensions.

javiereguiluz:
Just asking ... is this Pull Request ready or needs more changes? Thanks!

brentybh:
> Just asking ... is this Pull Request ready or needs more changes? Thanks!

@javiereguiluz The main thing finished, I think. Maybe we can't change more (what discussed above) in another PR

javiereguiluz:
Thanks Christian.

PR #1125, opened 2020-06-05
Move doctrine migrations away from src

rosier:
Related to https://github.com/symfony/recipes/issues/718


javiereguiluz:
Nice change! Thanks @rosier.

Issue #1127, opened 2020-06-17
Can not run composer update with composer-v2


gmponos:
As per the tweet here: https://twitter.com/gmponos/status/1273154848712769536?s=20

On a fresh installation that is installed with composer v1 when I run composer update with v2 I get the following errors:

![image](https://user-images.githubusercontent.com/5675248/84869270-7bf54800-b086-11ea-9251-7e47c462a95a.png)

if this is a problem of composer instead we can close and open on composer instead..

PR #1128, opened 2020-06-17
Fix compat with Composer 2 by using composer/package-versions-deprecated instead of ocramius/package-versions

nicolas-grekas:
As suggested by composer 2 when trying to upgrade deps.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
     "require": {
         "php": "^7.2.9",
         "ext-pdo_sqlite": "*",
+        "composer/package-versions-deprecated": "^1.8",
         "doctrine/doctrine-bundle": "^1.12|^2.0",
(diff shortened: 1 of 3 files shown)


gmponos:
Fixes https://github.com/symfony/demo/issues/1127

nicolas-grekas:
Might be fixed with a simple composer up if these PRs are accepted:
- https://github.com/doctrine/migrations/pull/999
- https://github.com/doctrine/dbal/pull/4084
- https://github.com/doctrine/orm/pull/8184

gmponos:
All the above are merged.. so do we close this and wait for the above packages to be tagged?

javiereguiluz:
Let's merge this now so it's ready for everybody ... and in the next update we'll install the other dependencies with the needed changes. Thanks Nicolas!

PR #1129, opened 2020-06-22
Rename namespace App\Events to App\Event to be in sync with the docs

### LINKED ISSUE #1092: Use singular namespaces
The demo has singular namespaces except for DataFixtures, Events and Migrations.

The documentation uses singular form, for example `App\Event\...` (https://symfony.com/doc/current/event_dispatcher.html), I think it would be good to be in-pair with the doc and the conventions.

What do you think ?

rosier:
Fixes #1092


javiereguiluz:
Thanks @rosier for this contribution!

PR #1133, opened 2020-07-13
Add a test for command ListUsers (test emails)

Nek-:
The command list user was not tested until now, and that's quite sad because it's one of the only parts of the code that actually sends an email. Symfony has some tools for testing emails, let's use them!

I also refactored a little bit of command testing since there was an identical piece of code between test cases. (instantiating the command)

### DIFF
--- a/tests/Command/AbstractCommandTest.php
+++ b/tests/Command/AbstractCommandTest.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Tests\Command;
+
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\Console\Tester\CommandTester;
+
+abstract class AbstractCommandTest extends KernelTestCase
+{
+    protected function setUp(): void
+    {
+        if ('Windows' === \PHP_OS_FAMILY) {
+            $this->markTestSkipped('`stty` is required to test this command.');
+        }
+    }
+
+    /**
+     * This helper method abstracts the boilerplate code needed to test the
+     * execution of a command.
+     *
+     * @param array $arguments All the arguments passed when executing the command
+     * @param array $inputs    The (optional) answers given to the command when it asks for the value of the missing arguments
+     */
+    protected function executeCommand(array $arguments, array $inputs = []): CommandTester
+    {
+        self::bootKernel();
+
+        // this uses a special testing container that allows you to fetch private services
+        $command = self::$container->get($this->getCommandFqcn());
+        $command->setApplication(new Application(self::$kernel));
+
+        $commandTester = new CommandTester($command);
+        $commandTester->setInputs($inputs);
+        $commandTester->execute($arguments);
+
+        return $commandTester;
+    }
+
+    abstract protected function getCommandFqcn(): string;
+}
--- a/tests/Command/ListUsersCommandTest.php
+++ b/tests/Command/ListUsersCommandTest.php
@@ -1,0 +1,57 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Tests\Command;
+
+use App\Command\ListUsersCommand;
+
+class ListUsersCommandTest extends AbstractCommandTest
+{
+    /**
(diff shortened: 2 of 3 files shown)


tests/Command/ListUsersCommandTest.php
```diff
@@ -0,0 +1,64 @@
...
+use App\Command\ListUsersCommand;
+
+class ListUsersCommandTest extends AbstractCommandTest
+{
+    /**
+     * @dataProvider maxResultsProvider
+     *
+     * This test verifies the amount of data is right according to the given parameter max results.
+     */
+    public function testListUsers(int $maxResults)
```
ker0x, line 23:
```suggestion
    public function testListUsers(int $maxResults): void
```

tests/Command/ListUsersCommandTest.php
```diff
@@ -0,0 +1,64 @@
...
+        yield [1];
+        yield [2];
+    }
+
+    /**
+     * @dataProvider sendEmailProvider
+     *
+     * This test ensure that the command actually sends an email when the parameter is specified
+     */
+    public function testItSendsAnEmail(bool $sendAndEmail)
```
ker0x, line 44:
```suggestion
    public function testItSendsAnEmail(bool $sendAndEmail): void
```

Nek-:
Thank you @ker0x for feedbacks. This is fixed.

tests/Command/AbstractCommandTest.php
```diff
@@ -0,0 +1,53 @@
...
+
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\Console\Tester\CommandTester;
+
+abstract class AbstractCommandTest extends KernelTestCase
+{
+    protected function setUp(): void
+    {
+        exec('stty 2>&1', $output, $exitcode);
```
javiereguiluz, line 22: I was checking the `Terminal::hasSttyAvailable()` method:

https://github.com/symfony/symfony/blob/de39dbae8f2425f48e790c4f1b3e70f9b702554f/src/Symfony/Component/Console/Terminal.php#L63

It's public ... but marked as `@internal`, so I guess we can't use it. So, could we please at least borrow the check for the tricky `exec()` function to not fail in sistems that have disabled it? Thanks!

Nek-: This is a refactoring (moving code by copy/pasting) from the already existing test. So it seems out of my scope. But I can add it anyway.

Do you mean adding such code?

```php
<?php
        if (!\function_exists('exec')) {
            $this->skip();
        }
```

javiereguiluz: It's true that it's not strictly related to your changes, so sorry for asking this ... but don't you think it would make the code slightly better? The `exec()` function is so special that many servers disable it. Thanks!

Nek-: You didn't answer my question but well, I'll consider it as a yes ^^ .

tests/Command/ListUsersCommandTest.php
```diff
@@ -0,0 +1,64 @@
...
+        );
+
+        $this->assertEmailCount($sendAndEmail ? 1 : 0);
+    }
+
+    public function sendEmailProvider(): ?\Generator
+    {
+        yield [true];
+        yield [false];
+    }
```
voronkovich, line 58: IMHO, it's better to have 2 different tests for this, instead of using data provider.

```suggestion
    public function testItSendsNoEmailByDefault(): void
    {
        $this->executeCommand([]);

        $this->assertEmailCount(0);
    }

    public function testItSendsAnEmailIfOptionProvided(): void
    {
        $this->executeCommand(['--send-to' => 'john.doe@symfony.com']);

        $this->assertEmailCount(1);
    }
```

Nek-:
@javiereguiluz rebased!

javiereguiluz:
Maxime, thanks for providing this test and for your work in the review phase. Thanks to reviewers too!

tests/Command/AbstractCommandTest.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\Console\Tester\CommandTester;
+
+abstract class AbstractCommandTest extends KernelTestCase
+{
+    protected function setUp(): void
+    {
+        if ('Windows' === \PHP_OS_FAMILY) {
+            $this->markTestSkipped('`stty` is required to test this command.');
```
stof, line 23: Moving this to the abstract class is wrong. Not all command tests require `stty`. And AFAICT, testing the `list:users` command will not require stty as it is not an interactive command.

javiereguiluz: Mmmmm, you are right! Anybody willing to make a Pull Request to fix this? Thanks!

bocharsky-bw: Fixed in #1230

PR #1134, opened 2020-07-19
Remove duplicated rules

michaelKaefer:
Remove duplicated rules which are already included in `@Symfony` rules.

To verify that the `@Symfony` rules include the removed rules see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.16/src/RuleSet.php or run:
```bash 
php php-cs-fixer.phar describe @Symfony | grep -A 3 array_syntax
php php-cs-fixer.phar describe @Symfony | grep -A 3 no_superfluous_phpdoc_tags
// ...
```


javiereguiluz:
Thanks Michael. I like this kind of contributions because even if it looks small, it removes "noise" and simplifies maintenance. Thanks!

Issue #1135, opened 2020-07-21
Pagination does not preserve filter by tag


jkufner:
Start the demo and go to `http://localhost:8000/en/blog/?tag=consectetur`, then click "Next →" button in the paginator.

The next URL should be `http://localhost:8000/en/blog/page/2?tag=consectetur`, but it is `http://localhost:8000/en/blog/page/2` instead — the `?tag=consectetur` is missing.

javiereguiluz:
Thanks for reporting ... and fixing this in #1136.

PR #1136, opened 2020-07-21
bug #1135 Fix pagination with tag filter

jkufner:
See https://github.com/symfony/demo/issues/1135


javiereguiluz:
And another great bug fix! You are on a roll Josef! Thank you.

PR #1138, opened 2020-07-21
Fix search query

jkufner:
When visiting `http://localhost:8000/cs/blog/search?q=de` directly, the search is not triggered automatically. This PR triggers the search on page load and passes the provided query argument to the input field.


javiereguiluz:
Very good catch Josef! I like this contribution a lot. Thanks!

PR #1141, opened 2020-07-31
Updated dependencies

javiereguiluz:
This updates both the back and the front dependencies.

-----

When testing the app with the new dependencies, I see this error in the browser console:

image

This is what I see in the HTML source code of the page:

image

Which is the same as in the `entrypoints.json` file:

image


javiereguiluz:
Closing in favor of #1151 and #1152.

PR #1145, opened 2020-08-30
Update _comment_form.html.twig

Chi-teck:
Hiding the textarea label with CSS makes no sense here. We could simply remove the call `form_label` instead. However I think it makes sense to keep the label for assistive technologies.


bocharsky-bw (approved):

yceruto (approved):

javiereguiluz:
Much better! Thanks Ivan!

PR #1151, opened 2020-09-10
Updated front-end dependencies

javiereguiluz:


stof:
@javiereguiluz this still uses an outdated version of Encore, because you performed only semver-compatible upgrades (and Encore is still in the 0.x range, so upgrading from 0.28 to 0.31 involves 3 semver-breaking upgrades)

javiereguiluz:
@stof you are right! I'm trying to fix this in #1153.

PR #1154, opened 2020-09-30
Added "Close" Translation

### LINKED ISSUE #1144: Translate aria-label attribute in close buttons
Currently it always appears in English.
https://github.com/symfony/demo/blob/v1.6.1/templates/debug/source_code.html.twig#L14
https://github.com/symfony/demo/blob/v1.6.1/templates/default/_flash_messages.html.twig#L23

eltomato89:
Fixed #1144


javiereguiluz:
Thank you Jens.

PR #1156, opened 2020-10-02
replaced external $_SERVER variable with filter_input

francescobianco:
It is just a micro fix, I hope it can be considered a small improvement towards the use of this standard.

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -9,21 +9,21 @@ require dirname(__DIR__).'/vendor/autoload.php';
 
 (new Dotenv())->bootEnv(dirname(__DIR__).'/.env');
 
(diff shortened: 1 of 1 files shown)


stof:
-1 for that. There is no reason to make the front controller diverge from the official recipe.

francescobianco:
@stof sorry for hurting you, is there something bad on usage of filter_input?

stof:
Well, using `filter_input` without any filter provides no value.

And this repo is the demo project. Making its `index.php` diverge from what gets created by default for a Symfony project does not make any sense, and would rather create confusion when newcomers compare the demo to their newly created project..

GrahamCampbell:
This PR looks like spam.

PR #1157, opened 2020-10-04
Add missing spanish translation

DIOHz0r:
There is a missing string translation from the English version.


javiereguiluz:
Thank you Domingo.

PR #1163, opened 2020-10-08
Update to Symfony 5.2

javiereguiluz:


composer.json
```diff
@@ -17,26 +17,26 @@
...
         "doctrine/orm": "^2.5.11",
         "erusev/parsedown": "^1.6",
-        "sensio/framework-extra-bundle": "^5.1",
+        "sensio/framework-extra-bundle": "^5.6",
         "symfony/apache-pack": "^1.0",
-        "symfony/asset": "^5.1",
-        "symfony/console": "^5.1",
-        "symfony/dotenv": "^5.1",
-        "symfony/expression-language": "^5.1",
+        "symfony/asset": "5.2.0-beta1",
```
stof, line 22: use `^5.2.0-beta1` to avoid CI failures on `composer validate --strict`

stof:
have you checked where the new deprecation comes from ?

javiereguiluz:
@stof yes ... it's from Symfony's Twig Loader pass ... but I was going to fix it and couldn't, because apparently it's been fixed since the beta release:

image

composer.json
javiereguiluz: Done. Thanks.

stof:
this might indicate missing tests in Symfony, as the testsuite should forbid deprecations...

javiereguiluz:
@stof @nicolas-grekas after upgrading to Symfony 5.2 beta3 I can still see that deprecation. Is this something to be fixed upstream in Symfony? Thanks! Details:

image

javiereguiluz:
I've just updated to 5.2 RC1. I don't see any issues or deprecations ... and the application feels super lightweight and responsive.

javiereguiluz:
I've updated to 5.2 RC2. These are the reported deprecations:

```
Remaining indirect deprecation notices (7)

  6x: The "Doctrine\Common\Inflector\Inflector::classify" method is deprecated and will be dropped in doctrine/inflector 2.0. Please update to the new Inflector API.
    2x in AddUserCommandTest::testCreateUserNonInteractive from App\Tests\Command
    2x in AddUserCommandTest::testCreateUserInteractive from App\Tests\Command
    1x in BlogControllerTest::testAdminNewPost from App\Tests\Controller\Admin
    1x in UserControllerTest::testEditUser from App\Tests\Controller

  1x: Since symfony/dependency-injection 5.2: The "Symfony\Component\DependencyInjection\Alias::setPrivate()" method is deprecated, use "setPublic()" instead.
```

PR #1165, opened 2020-10-10
Updates symfony recipes to latest

mstrom:
This PR is to update recipes to its latest versions 

- symfony/framework-bundle 
- symfony/security-bundle
- symfony/monolog-bundle
- doctrine/doctrine-fixtures-bundle
- doctrine/doctrine-migrations-bundle

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,7 +75,8 @@
     "scripts": {
         "auto-scripts": {
             "cache:clear": "symfony-cmd",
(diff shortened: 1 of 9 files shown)


javiereguiluz:
Michael, thanks for taking care of updating these recipes. We've reverted some minor changes made in the framework.yaml recipe because those were some custom comments that we added in this application to better explain some features ... so they don't come from the original Symfony recipe. Thanks!

PR #1169, opened 2020-10-15
PHP 8 Prototype

derrabus:
I have updated the demo application to showcase the new possibilities with php 8 and Symfony 5.2. These changes are of course too bleeding edge to merge, but you might get a good first impression about how a Symfony app on php 8 might look like.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         "symfony/polyfill-php73": "*"
     },
     "require": {
-        "php": "^7.3 || ^8.0",
+        "php": ">=8.0.2",
         "ext-pdo_sqlite": "*",
         "composer/package-versions-deprecated": "^1.8",
         "doctrine/dbal": "^3.1",
@@ -57,7 +57,7 @@
     },
     "config": {
         "platform": {
-            "php": "7.3"
+            "php": "8.0.3"
         },
         "preferred-install": {
             "*": "dist"
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -32,12 +32,10 @@ use Symfony\Component\Routing\Annotation\Route;
  *
  * See http://knpbundles.com/keyword/admin
  *
- * @Route("/admin/post")
- * @IsGranted("ROLE_ADMIN")
- *
  * @author Ryan Weaver <weaverryan@gmail.com>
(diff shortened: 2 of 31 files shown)


composer.json
```diff
@@ -3,40 +3,43 @@
...
     "replace": {
         "symfony/polyfill-php70": "*",
-        "symfony/polyfill-php72": "*"
+        "symfony/polyfill-php72": "*",
+        "symfony/polyfill-php73": "*",
+        "symfony/polyfill-php80": "*"
     },
     "require": {
-        "php": "^7.2.9",
+        "php": ">= 7.4",
```
derrabus, line 15: ⚠️ Note: This is a lie. Not all of our dependencies oficially support PHP 8 yet, so we pretend to pull packages for PHP 7.4.

shyim:
Did you get opcache preload running? I am getting segsev on each request

derrabus:
@shyim I haven't tried, tbh.

src/Controller/Admin/BlogController.php
```diff
@@ -64,13 +63,16 @@ public function index(PostRepository $posts): Response
...
      * NOTE: the Method annotation is optional, but it's a recommended practice
      * to constraint the HTTP methods each controller responds to (by default
      * it responds to all methods).
      */
+<<<<<<< HEAD
     public function new(Request $request, EntityManagerInterface $entityManager): Response
+=======
+    #[Route('/new', methods: ['GET', 'POST'], name: 'admin_post_new')]
+    public function new(Request $request): Response
+>>>>>>> 44c4fc1 (PHP 8 Prototype)
```
stof, line 75: This is a broken conflict resolution

symfony.lock
```diff
@@ -44,6 +44,9 @@
     "doctrine/deprecations": {
         "version": "v0.5.3"
     },
+    "doctrine/deprecations": {
+        "version": "v0.5.3"
+    },
```
derrabus, line 49: Should be reverted because the block is redundant now: see the three lines above it.

javiereguiluz: Done in https://github.com/symfony/demo/pull/1277/commits/8c4defe7556f4a08cf95d2a5d1ad977a5c2c47a2. Thanks!

PR #1170, opened 2020-10-16
Add missing type declarations

derrabus:
This PR adds some type declarations where they were missing before.


javiereguiluz:
I like it! Thanks Alexander!

PR #1171, opened 2020-11-03
Replace Travis and Appveyor by GitHub Actions

javiereguiluz:


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,87 @@
...
+              run: |
+                  ./bin/console lint:yaml config --parse-tags
+
+            - name: Lint Twig templates
+              run: |
+                  ./bin/console lint:twig templates --env=prod
+
+            - name: Lint XLIFF translations
+              run: |
+                  ./bin/console lint:xliff translations
```
ogizanagi, line 67: I think there is something non optimal here: as the composer update above will be executed with cache-warmers (due to the `cache:clear` command called after an update), it is likely to fail before reaching these linters steps and we'll get an incomplete run.

A solution to circumvent this could be to run the update without scripts.

javiereguiluz: Good catch! I've added the `--no-scripts` to `composer update`. Thanks!

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+            - name: "Composer update"
+              run: composer update --no-scripts
+
+            - name: Lint YAML files
+              run: |
+                  ./bin/console lint:yaml config --parse-tags
+
+            - name: Lint Twig templates
+              run: |
```
ogizanagi, line 62: If you want to execute all linters steps, no matter there was an error in a previous linter step, you can declare an id `deps` on the composer update step and use:

```suggestion
              if: always() && steps.deps.outcome == 'success'
              run: |
```

on each steps.
So even if the previous step is failing, the next ones are executed and we get all the infos about what is wrong in the PR.

(https://github.com/ogizanagi/symfony-lint-gha-demo/blob/000fcd5b832d6b35a3359d6b07a279445d9d8091/.github/workflows/lint.yaml#L40-L45)

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+jobs:
+    test:
+        name: "${{ matrix.operating-system }} / PHP ${{ matrix.php-version }}"
+        runs-on: ubuntu-latest
+        continue-on-error: ${{ matrix.allow-failures }}
+
+        strategy:
+            matrix:
+                allow-failures: [false]
```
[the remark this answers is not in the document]
OskarStark: This is used in line 19

.github/workflows/lint.yaml
javiereguiluz: Good idea! Let's do that. Thanks!

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+            - name: "Set composer cache directory"
+              id: composer-cache
+              run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+            - name: "Cache composer"
+              uses: actions/cache@v2.1.2
+              with:
+                  path: ${{ steps.composer-cache.outputs.dir }}
+                  key: ${{ runner.os }}-${{ matrix.php-version }}-composer-${{ hashFiles('composer.json') }}-allow-failures-${{ matrix.allow-failures }}
```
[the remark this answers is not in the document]
javiereguiluz: I don't know. I think I copied this from a file created by @OskarStark  Maybe he can share his reasoning. Thanks!

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,69 @@
...
+jobs:
+    test:
+        name: "${{ matrix.operating-system }} / PHP ${{ matrix.php-version }}"
+        runs-on: ubuntu-latest
+        continue-on-error: ${{ matrix.allow-failures }}
+
+        strategy:
+            matrix:
+                allow-failures: [false]
+                operating-system: ['ubuntu-latest', 'windows-latest', 'macos-latest']
```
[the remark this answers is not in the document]
javiereguiluz: GitHub Actions are much faster than Travis ... so I thought it was OK to test this. Which jobs would you propose to remove? Thanks!

javiereguiluz:
Here's a question about GitHub Actions: I guess we don't see the checks added to this PR because it's the PR where we adding them, so I only see the checks in my personal fork: https://github.com/javiereguiluz/demo/actions

That's OK ... but why GitHub only runs the `lint.yaml` workflow and ignores the `ci.yaml` workflow? Thanks!

ogizanagi:
> That's OK ... but why GitHub only runs the `lint.yaml` workflow and ignores the `ci.yaml` workflow? Thanks!

Because the ci workflow does only run on pushing new commits to master:

https://github.com/javiereguiluz/demo/blob/58bd1eb7456e12ee2a5fe8e915c80b0ba0f1ecee/.github/workflows/ci.yaml#L5-L7

while the lint one allows to run on any branch

javiereguiluz:
@ogizanagi Sorry I missed that! Thanks!

.github/workflows/ci.yaml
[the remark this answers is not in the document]
stof: the fact that you have a single value in the matrix makes it useless IMO.

And `continue-on-error: true` will make the exit code of each step be ignored and continue executing further steps, which will likely cause WTF errors if the setup steps fail for instance (as next step will still try to run).

.github/workflows/ci.yaml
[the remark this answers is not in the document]
stof: it is not only about the time it takes. It is also about execution limits. The concurrency limit is [20 jobs for the free plan](https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration#usage-limits) (I don't remember whether the symfony organization has a paid plan or no). And this limit applies **for the account**, not the repository. This means that jobs created in the demo are competing with `symfony/symfony` (and other repos) for the concurrency limit.
Even though the demo repository has low activity, it might make sense to avoid triggering 12 test jobs per build (plus 2 linting jobs).

Here, I would probably test all PHP versions on ubuntu (after all, I think prod servers are much more likely to run on Linux than windows or mac) and limiting the testing  on other  OS to either one version only or to the lowest and highest ones.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,88 @@
...
+              uses: docker://oskarstark/php-cs-fixer-ga
+              with:
+                  args: --diff --dry-run
+
+    linters:
+        name: Linters
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                allow-failures: [false]
```
stof, line 24: this is unused in the matrix.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,88 @@
...
+                operating-system: ['ubuntu-latest']
```
stof, line 25: this is unused in the matrix (`runs-on` does not use the matrix)

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,88 @@
...
+                php-version: ['7.4']
```
stof, line 26: As there is no valid reason to run our linters on several PHP versions (the result does not depend on the PHP version), I would remove the usage of a single-value matrix here and put this `7.4` directly in the config of `shivammathur/setup-php`

stof:
And for a workflow to run on a PR, the workflow file must exist in the repository. *new* workflows are not running for PRs sent from a fork.

.github/workflows/lint.yaml
OskarStark: I know this is not exactly need, but it is also used in the composer cache key so it is some kind of variable which is very well to read via: `${{ matrix.php-version }}`

I am for keeping the matrix for `php-version`

.github/workflows/ci.yaml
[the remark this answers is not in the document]
OskarStark:
```suggestion
                  key: ${{ runner.os }}-${{ matrix.php-version }}-composer-${{ hashFiles('composer.json') }}
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+        steps:
+            - name: "Checkout code"
+              uses: actions/checkout@v2.3.3
+
+            - name: "Install PHP with extensions"
+              uses: shivammathur/setup-php@2.7.0
+              with:
+                  coverage: "none"
+                  extensions: intl, mbstring, pdo_sqlite
```
OskarStark, line 35:
```suggestion
                  extensions: "intl, mbstring, pdo_sqlite"
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,69 @@
...
+                  restore-keys: ${{ runner.os }}-${{ matrix.php-version }}-composer-
+
+            - name: "Require symfony/flex"
+              run: composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-main
+
+            - if: matrix.php-version != '8.0'
+              run: composer update
+
+            - if: matrix.php-version == '8.0'
+              run: composer update --ignore-platform-reqs
```
OskarStark, line 60:
```suggestion
              run: composer update --ignore-platform-req=php
```

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,88 @@
+name: "Lint"
+
+on: [push, pull_request]
```
OskarStark, line 3:
```suggestion
on:
    push:
    pull_request:
```

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+env:
+    fail-fast: true
+
+jobs:
+    php-cs-fixer:
+        name: PHP-CS-Fixer
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
```
OskarStark, line 13:
```suggestion
            - name: "Checkout code"
              uses: actions/checkout@v2
```

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,88 @@
...
+            - name: PHP-CS-Fixer
```
OskarStark, line 14:
```suggestion

            - name: PHP-CS-Fixer
```

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+            - name: "Set composer cache directory"
+              id: composer-cache
+              run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+            - name: "Cache composer"
+              uses: actions/cache@v2.1.2
+              with:
+                  path: ${{ steps.composer-cache.outputs.dir }}
+                  key: ${{ runner.os }}-${{ matrix.php-version }}-composer-${{ hashFiles('composer.json') }}-allow-failures-${{ matrix.allow-failures }}
```
OskarStark, line 48:
```suggestion
                  key: ${{ runner.os }}-${{ matrix.php-version }}-composer-${{ hashFiles('composer.json') }}
```

.github/workflows/lint.yaml
javiereguiluz: Both of you have good arguments ... but let's keep it in the matrix for now.

.github/workflows/lint.yaml
javiereguiluz: I like the compact notation more in this case.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
javiereguiluz: OK, understood. However, the main Symfony repos are going to keep using Travis and Appveyor for now ... so this won't impact in the important Symfony code repository. If this is a problem in the future, we can tweak this or kindly ask for higher limits to the friendly folks at GitHub.

javiereguiluz:
This is now merged ... so the switch is complete 🎉 

Maxime, Christophe and Oskar ... THANKS for your great help (and patience) here. I couldn't make this without your help. Thanks! 🙇

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,69 @@
...
+              with:
+                  path: ${{ steps.composer-cache.outputs.dir }}
+                  key: ${{ runner.os }}-${{ matrix.php-version }}-composer-${{ hashFiles('composer.json') }}-allow-failures-${{ matrix.allow-failures }}
+                  restore-keys: ${{ runner.os }}-${{ matrix.php-version }}-composer-
+
+            - name: "Require symfony/flex"
+              run: composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-main
+
+            - if: matrix.php-version != '8.0'
+              run: composer update
```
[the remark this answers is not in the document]
mstrom: @javiereguiluz maybe `composer install `?

PR #1173, opened 2020-11-05
Updated backend dependencies with Composer 2

### LINKED ISSUE #1172: Install demo - error
Installing twig/markdown-extra (v3.0.5): Extracting archive
    0 [>---------------------------]    0 [>---------------------------]Package zendframework/zend-code is abandoned, you should avoid using it. Use laminas/laminas-code instead.
Package zendframework/zend-eventmanager is abandoned, you should avoid using it. Use laminas/laminas-eventmanager instead.
Generating autoload files
composer/package-versions-deprecated: Generating version class...
composer/package-versions-deprecated: ...done generating version class
86 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
> symfony-cmd
'symfony-cmd' is not recognized as an internal or external command,
operable program or batch file.
Script symfony-cmd handling the auto-scripts event returned with error code 1
Script @auto-scripts was called via post-install-cmd

javiereguiluz:
Fixes #1172.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -81,16 +81,16 @@
         },
         {
             "name": "doctrine/annotations",
-            "version": "1.10.4",
+            "version": "1.11.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/annotations.git",
-                "reference": "bfe91e31984e2ba76df1c1339681770401ec262f"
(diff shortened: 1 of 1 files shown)


stof (approved):
To fully fix the issue, we also need a new release

PR #1176, opened 2020-11-22
Updated JavaScript dependencies

Coderberg:
I have updated js dependencies for two reasons:

1. Webpack Encore requires version ^9.0.1 of sass-loader, but your version (7.3.1) is too old. T
he related feature will probably not work correctly.

2. Vulnerability in "object-path" <= 0.11.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "jquery": "^3.5.1",
     "lato-font": "^3.0.0",
     "node-sass": "^4.9.3",
-    "sass-loader": "^7.1.0",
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Thanks Valery!

Minor comment: because of how this app is distributed, we must recompile all assets for production whenever we change them. So, I run `yarn install && yarn encore production` after merging your PR and then I committed the result in https://github.com/symfony/demo/commit/d8d09068eba6bd2ad68ae6bffd88f65dc263dc21

PR #1177, opened 2020-11-28
Replaced `return 0` with `return Command::SUCCESS`

Coderberg:


src/Command/AddUserCommand.php
```diff
@@ -204,7 +204,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             $this->io->comment(sprintf('New user database id: %d / Elapsed time: %.2f ms / Consumed memory: %.2f MB', $user->getId(), $event->getDuration(), $event->getMemory() / (1024 ** 2)));
         }
 
-        return 0;
+        return Command::SUCCESS;
```
derrabus, line 207:
```suggestion
        return self::SUCCESS;
```

javiereguiluz: Alexander, you are technically correct. However, when using `self:: ...` it looks like it's something from our own app. Seeing `Command:: ...` however, makes it more easy to understand as a Symfony thing. Let's merge the original proposal and we can revisit this decision in the future. Thanks!

javiereguiluz:
Thank you Valery.

PR #1179, opened 2020-12-02
Update setters

seb-jean:
Hi,

I updated setters because there is error :

`Expected argument of type "string", "null" given at property path "title".
`


javiereguiluz:
Thank you @seb-jean.

PR #1030, opened 2019-09-17
Update bootswatch and bootstrap dependencies

abdounikarim:
Hey guys, 

I think i've finished this PR symfony/demo#1029

Here is what I did : 

1. Update dependencies : 
- [bootstrap](https://yarnpkg.com/en/package/bootstrap) (to version 4)
- [bootswatch](https://yarnpkg.com/en/package/bootswatch) (to version 4)

2. Replace dependencies : 
- [bootstrap-sass](https://yarnpkg.com/en/package/bootstrap-sass) by [Bootstrap](https://yarnpkg.com/en/package/bootstrap) updated version
- [eonasdan-bootstrap-datetimepicker](https://yarnpkg.com/en/package/eonasdan-bootstrap-datetimepicker) by [tempusdominus-bootstrap-4](https://yarnpkg.com/en/package/tempusdominus-bootstrap-4)

3. Update assets : 
- replace bootstrap-sass/assets/javascripts/bootstrap/transition.js by bootstrap/scss/transitions

4. Remove unused dependencies : 
- [bootstrap-sass] (https://yarnpkg.com/en/package/bootstrap-sass)
- [eonasdan-bootstrap-datetimepicker](https://yarnpkg.com/en/package/eonasdan-bootstrap-datetimepicker)

5. Update templates in `templates` directory (sorry if i miss something)

6. CSS : 
well class not replaced (not existing in the new [Flatly](https://bootswatch.com/flatly/) theme

7. Generate new files for production with `yarn build`

Let me know if i miss something please, thank you 😉

PR #1183, opened 2020-12-24
Upgrade bootswatch to version 4

bechir:
Hi,

After comparing changes before and after in #1030 I said to myself It's maybe better to create remake the job with new newly updated `Symfony 5.2` (and `bootswtach v5.4.3`).

I know @javiereguiluz wanted to resolve the conflicts himself but you can still see the work I did here

### Before
home

## After
home

### Before
blog

## After
blog

### Before
show-code

### After
show-code

### Before
pagination

### After
pagination

### Before
login

### After
login

### Before
admin

### After
admin

### Before
edit

### After
edit

### Before
show-post

### After
show-post

### Before
post-tags

### After
post-tags

### Before
post-errors

### After
post-errors

## CHANGELOG

**In packages.json**

* Upgraded bootswatch to version 4.5.3
* Upgraded bootstrap to version 4.5.3
* Removed `eonasdan-bootstrap-datetimepicker` dependency (This is not compatible with bootstrap 4)

**In scss files**

* Removed the `$icon-font-path` we no longer need this
* Moved variables import to new [`_variables.scss`](assets/scss/_variables.scss) file containg our custom variables
* Updated imports
* Replaced hardcoded colors value with custom variables name

**In javascript files**

* Replaced deprecated method `submit()` by `trigger('submit')`
* Updated highlightjs imports

**In twig files**

* Replaced `well` class by `jumbotron` with some changes in the default paddings
* Replaced `label label-*` by `badge badge-*`
* Replaced `pull-right` by `float-right` I d'ont kown if it's the same but we have the same result
* Updated classes name
* Added the active route class to `active`

**Tests**

* Beacause the form layout changed, we need to update assertions in `App\Tests\Controller\Admin\BlogControllerTest::testAdminNewPost`

**I didn't added `datetimepicker` yet because the [tempusdominus/bootstrap-4](https://github.com/tempusdominus/bootstrap-4) project is getting rolled back into the orginal repo, but it's still in progress.**

[This lib](https://www.jqueryscript.net/time-clock/Date-Time-Picker-Bootstrap-4.html) is for bootstrap 4 and it's based on the Eonasdan's Bootstrap 3 date/time picker.
But [flatpickr](https://github.com/flatpickr/flatpickr) is more powerful with zero dependencies

I didn't added it yet, I'm juste waiting for feedbacks.

Thanks for your time!


assets/js/app.js
```diff
@@ -1,11 +1,12 @@
 import '../scss/app.scss';
 
 // loads the Bootstrap jQuery plugins
-import 'bootstrap-sass/assets/javascripts/bootstrap/transition.js';
-import 'bootstrap-sass/assets/javascripts/bootstrap/alert.js';
-import 'bootstrap-sass/assets/javascripts/bootstrap/collapse.js';
-import 'bootstrap-sass/assets/javascripts/bootstrap/dropdown.js';
-import 'bootstrap-sass/assets/javascripts/bootstrap/modal.js';
+// import 'bootstrap-sass/assets/javascripts/bootstrap/transition.js';
```
noniagriconomie, line 4: why commented?
also next line is duplicated

templates/form/layout.html.twig
```diff
@@ -1,16 +1,18 @@
...
 {# Errors #}
 
+{#
 {% block form_errors -%}
     {% if errors|length > 0 -%}
         {% if form is not rootform %}<span class="help-block">{% else %}<div class="alert alert-danger">{% endif %}
         <ul class="list-unstyled">
         {%- for error in errors -%}
-            {# use font-awesome icon library #}
+            {# use font-awesome icon library # }
```
noniagriconomie, line 11: to rollback

templates/base.html.twig
```diff
@@ -3,6 +3,9 @@
    common elements and decorates all the other templates.
    See https://symfony.com/doc/current/templates.html#template-inheritance-and-layouts
 #}
+
+{% set route = app.request.get('_route') %}
```
noniagriconomie, line 7:
```suggestion
{% set _route = app.request.get('_route') %}
```

what about this?

noniagriconomie (commented):
nice one :)

templates/form/layout.html.twig
bechir: Ok i rolled it back and i got this
Screen Shot 2020-12-30 at 1 21 09 AM

bechir: And the test `testAdminNewDuplicatedPost` method is failing because we no longer have `form-group.has-error`

bechir:
Hi @noniagriconomie,
Thanks you for your feedbacks

I updated my code but the datetime picker initialization is still missing. 
You can check [flatpickr](https://github.com/flatpickr/flatpickr) the design look good and it's about `13KB`

noniagriconomie:
@bechir i think i am not the person who will decide if or not to replace this lib,
for me it does not matter, as the main point of this projetc is showcasing symfony etc, but lets wait main maintainers 👍

templates/form/layout.html.twig
noniagriconomie: i was commenting to rollback the extra space on the precise line i commented on

from

`{# use font-awesome icon library #}`

to
 `{# use font-awesome icon library # }`

sorry if it was unclear :s

bechir: Ah sorry.
The reason why I added that extra space is when I don't do so the code after that line will not be commented

bechir:
OK thanks you again.

javiereguiluz:
@bechir thanks a lot for contributing this!

The proposed changes look fantastic to me! About flatpicker, yes let's use it. I'd prefer to use the browser native datetime-local input widget, but that doesn't work in Firefox yet, so let's use flatpicker. Thanks!

assets/scss/bootstrap-tagsinput.scss
```diff
@@ -87,11 +87,10 @@
...
 * ---------------------------------------------------------------------------- */
 .bootstrap-tagsinput {
-  display: table-cell;
-  vertical-align: middle;
-  width: 100%;
+  flex: 1 1;
+  width: 1%;
   height: 45px;
-  padding: 0;
+  padding: 00.125rem 0 0 0;
```
stof, line 93: why `00.` ?

package.json
```diff
@@ -23,5 +22,8 @@
     "dev": "encore dev",
     "watch": "encore dev --watch",
     "build": "encore production --progress"
+  },
+  "dependencies": {
```
stof, line 26: why in dependencies when all others are defined in `devDependencies` ?

templates/base.html.twig
```diff
@@ -3,6 +3,9 @@
    common elements and decorates all the other templates.
    See https://symfony.com/doc/current/templates.html#template-inheritance-and-layouts
 #}
+
+{% set _route = app.request.get('_route') %}
```
stof, line 7: you should define this variable only in the `header` block where you need it rather than at the root of the layout (where it would impact the scope of all rendered templates)

templates/form/fields.html.twig
```diff
@@ -10,17 +10,21 @@
 {% block date_time_picker_widget %}
     <div class="input-group date" data-toggle="datetimepicker">
         {{ block('datetime_widget') }}
-        <span class="input-group-addon">
-            <span class="fa fa-calendar" aria-hidden="true"></span>
-        </span>
+        <div class="input-group-append">
+            <button class="btn btn-outline-secondary">
```
stof, line 14: a button without an explicit `type` is a very bad idea: this makes it a submit button. and this is even worse here because this is indeed inside a form...

package.json
```diff
@@ -23,5 +22,8 @@
...
+    "tempusdominus-bootstrap-4": "^5.39.0"
```
stof, line 27: this library looks unused

package.json
bechir: Sorry i missed this

assets/scss/bootstrap-tagsinput.scss
bechir: This is a typo, fixed!

package.json
bechir: I've  missed this

templates/form/fields.html.twig
bechir: So I replaced the `button` by a `span`

package.json
bechir: Removed.

templates/base.html.twig
bechir: Ok, I see that. Thanks.

bechir:
Thanks for your feedbacks @javiereguiluz and @stof!
I definitively need help about integrating `flatpickr`.

The problem is the date is not parsed correctly in `admin.js` 
```js
$('[data-toggle="datetimepicker"]').flatpickr({
        enableTime: true,
        dateFormat: $('#post_publishedAt').data('date-format'),
        allowInput: true,
        parseDate: (datestr, format) => {
            return moment(datestr, format, true).toDate();
        },
        formatDate: (date, format, locale) => {
            return moment(date).format(format);
        }
});
```

The calendar widget is shown but when I select a date nothing happen.

templates/form/fields.html.twig
stof: using a button was the right thing to do from an accessibility point of view. But it needs an explicit `type=button` so that it is not a submit button.

bechir: Ok I understand. Thanks.

atierant:
> 

@see https://github.com/symfony/demo/pull/1412/commits/4a1d48b6967677bee3de30264a51174ed165f8b0#diff-38752f070c2739871db0d206c9862337269e4d635f2e63e37e53c5a248c791bd for a resolution without moment.js which is deprecated

javiereguiluz:
Closing as fixed by #1412. Bechir, I'm really sorry we didn't merge this contribution ... but I'm happy that most of your work was reused as the base of #1412. Thanks!

PR #1184, opened 2021-01-12
Add Bosnian (bs) & Serbian (Latn & Cyrl) translation

tambait:


javiereguiluz:
Ivan, thanks a lot for contributing all these translations. *puno hvala* and *много вам хвала*.

PR #1185, opened 2021-01-28
Allow using PHP 8 to run the project

javiereguiluz:
This project is not compatible with PHP 8 at the moment because of `doctrine/dbal`.

The problem is that to upgrade DBAL we need to upgrade our PHP requirement too. We require PHP `^7.2.9` (to match Symfony's own requirements) but the DBAL version compatible with PHP 8 requires `^7.3|^8.0`.

So, I guess we can make an exception and require PHP 7.3 in this project to fix this issue. @nicolas-grekas do you agree with this solution or do you think we can solve this differently? Thanks!


nicolas-grekas:
Works for me.
Don't miss upgrading the CI.

PR #1189, opened 2021-02-19
starting the new beginning

jromero82:
This is a sample pull request


javiereguiluz:
@jromero82 thanks for proposing this contribution!

However, we don't promote any commercial service in this repository (unless provided by the Symfony company) so we cannot accept these changes. It's true that gitpod provides a limited free tier, but it's mostly a commercial service. Thanks for understanding!

PR #1190, opened 2021-02-22
Update highlight.js

Kolbasyatin:
Fix the error

app.js:133 Uncaught TypeError: highlight_js_lib_highlight__WEBPACK_IMPORTED_MODULE_0___default.a.registerLanguage is not a function
    at Module../assets/js/highlight.js (app.js:133)
    at __webpack_require__ (runtime.js:80)
    at Module../assets/js/app.js (app.js:26)
    at __webpack_require__ (runtime.js:80)
    at checkDeferredModules (runtime.js:46)
    at Array.webpackJsonpCallback [as push] (runtime.js:33)
    at app.js:1

### DIFF
--- a/assets/js/highlight.js
+++ b/assets/js/highlight.js
@@ -1,8 +1,8 @@
-import hljs from 'highlight.js/lib/highlight';
+import hljs from 'highlight.js/lib/core';
 import php from 'highlight.js/lib/languages/php';
 import twig from 'highlight.js/lib/languages/twig';
 
 hljs.registerLanguage('php', php);
 hljs.registerLanguage('twig', twig);
 
-hljs.initHighlightingOnLoad();
+hljs.highlightAll();


javiereguiluz:
Alexandr, thanks for this contribution! I've checked the docs for highlight.js (https://highlightjs.org/usage/) and this is indeed the new recommended way of using the library. Thanks!

javiereguiluz:
Please note that we had to revert a method rename because I saw a "function doesn't exist" error:

https://github.com/symfony/demo/commit/79cf203fbb7787276aadc7c428d80b558da54555#diff-2526483395401c05a58326421dec001903c633fb650b64e8cd66597e6201fbec

Kolbasyatin:
@javiereguiluz  It's weird. Something is in 
high2
high1

versions difference ?

Kolbasyatin:
@javiereguiluz  Sorry, it was  my mistake. I've updated library .

high5

javiereguiluz:
It's good to know that in newer versions this method is renamed. Sooner or later we'll need to update the highlight.js dependency to avoid using legacy versions. Pull Requests are welcome to do that 😄  Thanks!

PR #1191, opened 2021-03-06
Add security checker to auto scripts

ahmed-bhs:


javiereguiluz:
Ahmed, thanks for this contribution.

We agree on checking if dependencies are vulnerable ... but we don't agree with this approach. Adding a 5MB binary to the repository doesn't feel right (and the binary updates continuously, so we should had to keep it updated forever).

Instead of this, we prefer to keep with the current security check (done with the same binary your propose here) but performed in the continuous integration server. See https://github.com/symfony/demo/blob/b199d7892c9fe9831ee9fe9b583697f51d1d8d9f/.github/workflows/lint.yaml#L79-L85

PR #1194, opened 2021-03-12
Allow PHP 8 and update dependencies

### LINKED ISSUE #1175: "composer create-project" gives an older version v1.0.5 when running with PHP 8.0.0
As PHP 8.0.0 is around the corner, it will be interesting to try out Symfony on this new version of PHP.

While as the current PHP requirement in composer.json is "php: ^7.2.9", PHP 8 would not be able to get this latest version. Instead, it gets v1.0.5.

I also tried to create the demo project with v1.0.5, installation was fine but eventually Doctrine/ORM threw an error because of a bug in the old version. Error message is below:

```
> Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::clearCache
PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\ContextErrorException: Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in /home/vagrant/pizza/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:2636
```

Creating the demo project with its latest version v1.6.3 works good with PHP 8.0.0, by appending `--ignore-platform-reqs` in the command:

```
composer create-project symfony/symfony-demo my_project --ignore-platform-reqs
```

rosier:
Fixes #1175 and #1204

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,16 +6,16 @@
     "minimum-stability": "stable",
     "prefer-stable": true,
     "replace": {
-        "symfony/polyfill-php70": "*",
-        "symfony/polyfill-php72": "*"
+        "symfony/polyfill-php72": "*",
+        "symfony/polyfill-php73": "*"
     },
     "require": {
-        "php": "^7.3",
+        "php": "^7.3 || ^8.0",
         "ext-pdo_sqlite": "*",
         "composer/package-versions-deprecated": "^1.8",
         "doctrine/doctrine-bundle": "^2.0",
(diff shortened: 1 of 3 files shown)


ostrolucky (approved):

derrabus:
CI fails because of doctrine/DoctrineMigrationsBundle#413.

derrabus (approved):

javiereguiluz:
Thank you Jan.

PR #1197, opened 2021-03-21
[CI] Use ramsey/composer-install

rosier:
~~Running `composer update` changes composer.lock~~

Use ramsey/composer-install to cache and install the composer dependencies

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -45,15 +45,12 @@ jobs:
               uses: actions/cache@v2.1.2
               with:
                   path: ${{ steps.composer-cache.outputs.dir }}
-                  key: ${{ runner.os }}-${{ matrix.php-version }}-composer-${{ hashFiles('composer.json') }}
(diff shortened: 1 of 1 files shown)


stof:
what should be done instead is running `composer install`. We don't want to run an update in that CI job, as we are committing the lock file.

jkufner:
`composer install && npm ci` is the correct way to do things in CI. However, running `composer validate` can be considered as running a test, which is a good idea too.

stof:
but that's the whole point. With your change moving it higher, other linters won't run anymore if the composer.json reports any warning. The validation should stay in the part of the steps being about running tests rather than going into the setup part.

rosier:
I moved `composer validate` up because I don't know why `composer update` was used instead of `composer install`. 

I agree with using `composer install` and like to propose using the action ramsey/composer-install for that.

javiereguiluz:
I'd prefer to minimize the third-party actions we use ... if `composer update` is wrong in our CI update, could we just replace it by `composer install`? Thanks

javiereguiluz:
Jan, thanks for improving this ... and thanks to your patience during the review process

rosier:
No worries

PR #1200, opened 2021-04-02
Avoid creating a session when the user is not logged in

gliderShip:
Calling app.user (app. getUser()) immediately starts a session if a session does not exists. This may create problems for caching public pages.

### DIFF
--- a/templates/base.html.twig
+++ b/templates/base.html.twig
@@ -60,7 +60,7 @@
                                     <a href="{{ path('blog_search') }}"> <i class="fa fa-search"></i> {{ 'menu.search'|trans }}</a>
                                 </li>
 
-                                {% if app.user %}
+                                {% if app.request.hasPreviousSession and app.user %}
                                     <li class="dropdown">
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
Hm, I just made a quick test:
```
{{ dump(app.request.hasPreviousSession) }} // false
{{ dump(app.request.hasSession) }} // true
{{ dump(app.request.session.isStarted) }} // false
{{ dump(app.user) }} // null 
{{ dump(app.request.hasPreviousSession) }} // false
{{ dump(app.request.hasSession) }} // true
{{ dump(app.request.session.isStarted) }} // false
```

I don't see the session is started after `app.user` check. Could you explain it a bit further, probably with steps to reproduce the problem?

bocharsky-bw:
@gliderShip please, could you give any real-world example where your change may help? Or at least how to reproduce the problem?

javiereguiluz:
Let's close as "can't reproduce". We'll reopen if we receive more details. Thanks!

PR #1202, opened 2021-04-05
Remove PHP 8 workarounds from CI

derrabus:
We should not need to treat PHP 8.0 differently anymore in CI.


javiereguiluz:
Nice changes! Thanks Alexander.

PR #1206, opened 2021-04-29
[CI] Lint the container loaded from source instead of xml

rosier:
Workaround for https://github.com/doctrine/DoctrineMigrationsBundle/issues/413 and https://github.com/symfony/symfony/issues/39259


javiereguiluz:
Thank you Jan.

PR #1214, opened 2021-05-17
Update workflow triggers

rosier:
To prevent dependabot from triggering the lint workflow twice


javiereguiluz:
Jan, thanks for taking care of this maintenance!

PR #1215, opened 2021-05-17
Bump dependencies to Symfony 5.2.9

rosier:


rosier:
Tests are broken because the setup of PHP fails on the Mac OS X job with the following error:

dyld: Library not loaded: /usr/local/opt/openldap/lib/libldap-2.5.0.dylib

javiereguiluz:
Thank you Jan.

PR #1220, opened 2021-05-26
fix entrypoints error

Alex-trem:
I suggest a bug fix  about my issue


bocharsky-bw:
`entrypoints.json` was changed and now we have merge conflicts here, I think we can close this PR now. Please, try the new changes first

bocharsky-bw:
I think we can close this as it's not relevant anymore, the file was already updated

javiereguiluz:
OK then. Let's close. Thanks!

PR #1230, opened 2021-07-12
Do not skip user list command test because of stty

bocharsky-bw:
Fixes an issue in this discussion: https://github.com/symfony/demo/pull/1133/files#r665459095


javiereguiluz:
Thank you Victor.

PR #1231, opened 2021-07-21
Add EasyAdmin to the project

javiereguiluz:


javiereguiluz:
Sorry, the PR was sent to the wrong repository!

bocharsky-bw:
Just for the cross-reference: https://github.com/EasyCorp/easyadmin-demo - in case someone would be looking for EasyAdminBundle demo and got here :)

PR #1261, opened 2021-08-27
Remove useless exclusion for autowiring

ScullWM:
Hello,

Reading the configuration of the project, I was curious to know why this folder was excluded but I couldn't find any trace of it in the current version of the project. So we can safely remove it.

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -24,7 +24,6 @@ services:
     App\:
         resource: '../src/'
         exclude:
-            - '../src/DependencyInjection/'
             - '../src/Entity/'
             - '../src/Kernel.php'
             - '../src/Tests/'


config/services.yaml
```diff
@@ -24,7 +24,6 @@ services:
     App\:
         resource: '../src/'
         exclude:
-            - '../src/DependencyInjection/'
```
bocharsky-bw, line 27: This comes from the recipe: `https://github.com/symfony/recipes/blob/5e74696062f73b2e8292781ba85404fcd1a11119/symfony/framework-bundle/5.4/config/services.yaml#L19` - probably better to keep it to be in sync with the recipe

ScullWM: Well, recipes take care of the setup of a component for a project. So I don't see any constraint of owning our own `service.yaml` file of our symfony/demo project after the execution of the recipe. Recipe are a part of "default configuration" that could be changed.

Issue #1260, opened 2021-08-15
Add phpstan 


BackEndTea:
A lot has changed since https://github.com/symfony/demo/pull/733, we now have baselines and a good way to ignore errors.

Is there any interest in adding it to the project?

javiereguiluz:
Yes ... this demo project follows Symfony practices as close as possible ... so, given that Symfony code already uses PHPStan, we can try to add it here too. Thanks!

burned42:
I gave this a try and created https://github.com/symfony/demo/pull/1265 I hope that's ok :)

PR #1265, opened 2021-10-03
Add phpstan to github actions workflow

burned42:
Fixes https://github.com/symfony/demo/issues/1260

I tried checking different repos under https://github.com/symfony to see how those solve running phpstan and tried to apply that here. I saw that all the repos seem to install phpstan in the actions workflows instead of adding them to composer.json,so I did that, too. Also it seems like all of the projects I checked out are using the phpstan level 5, so that's what I was using. What I also found out is that most of them configured phpstan to use some phpunit autoload file, that's also why phpunit has to be installed in the workflow before running phpstan.

Most of the errors reported with this configuration are probably not that hard to solve, but I didn't want to include any unrelated code changes here, so I added a baseline file.

I think adding phpstan to the lint workflow sounds right, but I wasn't too sure if phpstan should be added to the 'linters' job or if it should get its own job. I decided to add it to the existing one to avoid having to duplicate all the composer/php setup steps into another job.

I hope that's somehow what you expected to see regarding running phpstan for this project :)


phpstan.neon.dist
```diff
@@ -0,0 +1,10 @@
+includes:
+    - phpstan-baseline.neon
+
+parameters:
+    level: 5
```
stof, line 5: Well, I suggest running at the highest possible level, as the higher the level, the better phpstan is at catching mistakes. The level configuration is a balance between catching mistakes (the higher the better for that) and not being too annoying (when introducing phpstan in a new project, higher levels can be hard to reach)

.github/workflows/lint.yaml
```diff
@@ -87,3 +87,11 @@ jobs:
             - name: Check if any dependencies are compromised
               if: always() && steps.install.outcome == 'success'
               run: /home/runner/.symfony/bin/symfony check:security
+
+            - name: Install PHPUnit
```
stof, line 91: If phpunit is installed because one of the linting steps depends on it, it should be part of the installation section (which is checked with `steps.install.outcome == 'success'` in linting steps to ensure that the install worked before running linting steps).

Note that `id: install` must be set on the last step belonging to the installation.

phpstan-baseline.neon
```diff
@@ -0,0 +1,22 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Strict comparison using \\=\\=\\= between null and App\\\\Entity\\\\User will always evaluate to false\\.$#"
```
stof, line 4: Instead of putting that in the baseline, the right solution is to fix the invalid phpdoc that says `/** @var User $user */` in DeleteUserCommand when it should say `/** @var User|null $user */`

phpstan.neon.dist
burned42: @stof I totally agree :+1: 

As written in the description, I checked multiple repos below github.com/symfony/ and all that I checked used level 5, so I thought it was some kind of convention for symfony repos :D 

Updated to max level in 761acfdd07b56feb2eb8de0803acc8c8c8c3f456 (which also results in a quite a bit larger baseline file)

.github/workflows/lint.yaml
burned42: @stof Thanks, I didn't notice there was kind of a 'installation section'. I moved the phpunit installation step there in 05fce162699163f5ccb64fbab2ae3cf4273d883f.

Shouldn't the 'Download Symfony CLI' step also be part of that 'installation section'?

phpstan-baseline.neon
burned42: @stof Sure, as written in the description, I wasn't sure if fixing the reported errors should be part of this PR. I fixed the type hint in 39c3d07e9ee90c8a49dcff96f949a4ab36241124.

Should I try fixing some more of the remaining issues as well in this PR?

seb-jean:
Why not use this repository?
https://github.com/OskarStark/phpstan-ga

As it is done here:
https://github.com/symfony/demo/blob/main/.github/workflows/lint.yaml#L21

burned42:
> Why not use this repository? https://github.com/OskarStark/phpstan-ga
> 
> As it is done here for PHP CS Fixer: https://github.com/symfony/demo/blob/main/.github/workflows/lint.yaml#L21

@seb-jean sure, we could do that. What I did was searching through the symfony organization to see how it's done in other repos here, and I found e.g. these three which all install phpstan through `shivammathur/setup-php@v2`:
- https://github.com/symfony/webpack-encore-bundle/blob/eeb893dec94ce5c471592588c945e364fe3ead09/.github/workflows/static.yml#L13
- https://github.com/symfony/mercure/blob/aac27352a8f17ce867a6b63c4f64906afb787dc4/.github/workflows/static-analysis.yml#L15
- https://github.com/symfony/panther/blob/ec2ff073e80d3efb986022a2937a9cd94310237d/.github/workflows/ci.yml#L31

So I thought solving it the same way it's already done in other symfony repos is probably a good idea, isn't it?

seb-jean:
> > Why not use this repository? https://github.com/OskarStark/phpstan-ga
> > As it is done here for PHP CS Fixer: https://github.com/symfony/demo/blob/main/.github/workflows/lint.yaml#L21
> 
> @seb-jean sure, we could do that. What I did was searching through the symfony organization to see how it's done in other repos here, and I found e.g. these three which all install phpstan through `shivammathur/setup-php@v2`:
> 
> * https://github.com/symfony/webpack-encore-bundle/blob/eeb893dec94ce5c471592588c945e364fe3ead09/.github/workflows/static.yml#L13
> * https://github.com/symfony/mercure/blob/aac27352a8f17ce867a6b63c4f64906afb787dc4/.github/workflows/static-analysis.yml#L15
> * https://github.com/symfony/panther/blob/ec2ff073e80d3efb986022a2937a9cd94310237d/.github/workflows/ci.yml#L31
> 
> So I thought solving it the same way it's already done in other symfony repos is probably a good idea, isn't it?

I prefer that more experienced people answer because I don't know what is the best solution.

javiereguiluz:
@burned42 thanks for this contribution. Do you consider that it's "mergeable" or is there anything pending? Thanks!

burned42:
> @burned42 thanks for this contribution. Do you consider that it's "mergeable" or is there anything pending? Thanks!

@javiereguiluz if you don't expect me to try and fix the reported phpstan errors within this PR, then yes, I consider this mergeable as is :)

PR #1266, opened 2021-10-08
Update PHP to 8.0

igornast:
Hi,

As PHP 8.0 was release almost a year ago, I thought that it would be nice to start using it's features for example attributes. Before I would be able to prepare attributes PR for it, demo's composer should be updated to support php:^8.0. 

Thanks for feedback and thoughts


javiereguiluz:
Hi again @igornast! After checking this PR and the older PR #1169, I've decided to merge the older one because it makes more changes to more files.

In any case, after it's merged, we can do new PRs to update other pending things to PHP 8. Thanks for understanding!

PR #1267, opened 2021-10-30
Create Procfile

herve1964:
web: vendor/bin/heroku-php-apache2 public/

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: vendor/bin/heroku-php-apache2 public/


javiereguiluz:
@herve1964 thanks for this contribution. However, this seems exclusively related to a particular cloud provider called Heroku. I've asked internally to the Core Team and there was a consensus about not merging PRs like these that are related to specific cloud providers. That's why we're closing it. Sorry!

PR #1268, opened 2021-11-05
Upgrade to Symfony 5.4

javiereguiluz:


composer.json
```diff
@@ -19,39 +19,41 @@
         "erusev/parsedown": "^1.6",
         "sensio/framework-extra-bundle": "^6.1",
         "symfony/apache-pack": "^1.0",
-        "symfony/asset": "^5.3",
-        "symfony/console": "^5.3",
-        "symfony/dotenv": "^5.3",
-        "symfony/expression-language": "^5.3",
+        "symfony/asset": "5.4.0-beta1",
```
stof, line 22: should probably use `^5.4.0-beta1` so that it can update to newer beta releases on next `composer update`, and then to stable releases.

composer.json
```diff
@@ -19,39 +19,41 @@
...
+        "symfony/console": "5.4.0-beta1",
+        "symfony/dotenv": "5.4.0-beta1",
+        "symfony/expression-language": "5.4.0-beta1",
         "symfony/flex": "^1.1",
```
stof, line 26: should probably be updated to `^1.17.1` due to https://symfony.com/blog/upgrade-flex-on-your-symfony-projects

composer.json
```diff
@@ -19,39 +19,41 @@
...
+        "symfony/validator": "5.4.0-beta1",
         "symfony/webpack-encore-bundle": "^1.4",
-        "symfony/yaml": "^5.3",
+        "symfony/yaml": "5.4.0-beta1",
         "tgalopin/html-sanitizer-bundle": "^1.2",
+        "twig/extra-bundle": "^2.12|^3.0",
         "twig/intl-extra": "^3.0",
-        "twig/markdown-extra": "^3.0"
+        "twig/markdown-extra": "^3.0",
+        "twig/twig": "^2.12|^3.0"
```
stof, line 45: allowing 2.12 here does not make sense when you extra packages are requiring the 3.0 versions which depend on Twig 3.

composer.json
```diff
@@ -75,7 +77,8 @@
     "scripts": {
         "auto-scripts": {
             "cache:clear": "symfony-cmd",
-            "assets:install --symlink --relative %PUBLIC_DIR%": "symfony-cmd"
+            "assets:install --symlink --relative %PUBLIC_DIR%": "symfony-cmd",
+            "assets:install %PUBLIC_DIR%": "symfony-cmd"
```
stof, line 81: having both commands does not make sense. The second one will override the first installation that used symlinks.

config/packages/security.yaml
```diff
@@ -25,12 +25,8 @@ security:
             security: false
 
         main:
-            # this firewall applies to all URLs
-            pattern: ^/
-
+            # this firewall doesn not have a 'pattern' option because it applies to all URLs
```
stof, line 28: typo here (extra `n`

src/Controller/BlogController.php
```diff
@@ -92,7 +93,7 @@ public function postShow(Post $post): Response
      * (postSlug) doesn't match any of the Doctrine entity properties (slug).
      * See https://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/converters.html#doctrine-converter
      */
-    public function commentNew(Request $request, Post $post, EventDispatcherInterface $eventDispatcher): Response
+    public function commentNew(Request $request, Post $post, EventDispatcherInterface $eventDispatcher, ManagerRegistry $doctrine): Response
```
stof, line 96: You could even inject the `EntityManagerInterface` directly.

stof (changes requested):

stof:
you should regenerate the phpstan baseline as you upgraded it to 1.0 and messages have changed.

composer.json
javiereguiluz: Yes. Done!

composer.json
javiereguiluz: Yes, this is very important. Thanks!

composer.json
javiereguiluz: Correct! This leftover has been removed.

composer.json
javiereguiluz: Yes, this came from here: https://github.com/symfony/recipes/blob/cb1cb50ace8fe990b60a7ea9e8f840cd6c30cbfb/symfony/framework-bundle/5.4/manifest.json#L12

I'm going to remove our customized command based on symlinks, to better align with Symfony's defaults.

src/Controller/BlogController.php
javiereguiluz: Much better. Done!

javiereguiluz:
@stof thanks for your nice and detailed review! I've done all the requested changes.

I'm only missing the "regenerate the phpstan baseline". I don't use PhpStan so could anyone please tell me the exact command to run for this case? Thanks a lot!

composer.json
```diff
@@ -89,7 +91,6 @@
     },
     "extra": {
         "symfony": {
-            "require": "5.3.*",
```
derrabus, line 92: Why not `"require": "5.4.*"`?

composer.json
```diff
@@ -3,7 +3,7 @@
     "license": "MIT",
     "type": "project",
     "description": "Symfony Demo Application",
-    "minimum-stability": "stable",
+    "minimum-stability": "beta",
     "prefer-stable": true,
```
stof, line 7: due to this `prefer-stable` and the removal of `extra.symfony.require`, many transitive symfony dependencies are staying on 5.3

stof:
@javiereguiluz `vendor/bin/phpstan analyse --generate-baseline`

composer.json
javiereguiluz: You are right! This was an important mistake of mine. I've just reverted the removal of `extra.symfony.require` and now everything is on 5.4 beta1 and I see more deprecation messages. Thanks!

javiereguiluz:
After properly upgrading dependencies, I'm seeing these indirect deprecations:

```
Testing
...................................................               51 / 51 (100%)
OK (51 tests, 116 assertions)

Remaining indirect deprecation notices (5)

  1x: Since symfony/doctrine-bridge 5.4: The "Symfony\Bridge\Doctrine\SchemaListener\PdoCacheAdapterDoctrineSchemaSubscriber" class is deprecated, use "Symfony\Bridge\Doctrine\SchemaListener\DoctrineDbalCacheAdapterSchemaSubscriber" instead.

  1x: Method "Symfony\Component\Config\Definition\ConfigurationInterface::getConfigTreeBuilder()" might add "TreeBuilder" as a native return type declaration in the future. Do the same in implementation "Twig\Extra\TwigExtraBundle\DependencyInjection\Configuration" now to avoid errors or add an explicit @return annotation to suppress this message.

  1x: Method "Symfony\Component\Config\Definition\ConfigurationInterface::getConfigTreeBuilder()" might add "TreeBuilder" as a native return type declaration in the future. Do the same in implementation "HtmlSanitizer\Bundle\DependencyInjection\Configuration" now to avoid errors or add an explicit @return annotation to suppress this message.

  1x: Method "Symfony\Component\Console\Command\Command::execute()" might add "int" as a native return type declaration in the future. Do the same in child class "Doctrine\DBAL\Tools\Console\Command\RunSqlCommand" now to avoid errors or add an explicit @return annotation to suppress this message.

  1x: Method "Doctrine\DBAL\Driver\Connection::quote()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "DAMA\DoctrineTestBundle\Doctrine\DBAL\AbstractStaticConnection" now to avoid errors or add an explicit @return annotation to suppress this message.
```

GromNaN:
You can upgrade to 5.4-rc1, all deprecation notices have been solved.

composer.json
```diff
@@ -3,7 +3,7 @@
     "license": "MIT",
     "type": "project",
     "description": "Symfony Demo Application",
-    "minimum-stability": "stable",
+    "minimum-stability": "beta",
```
derrabus, line 6:
```suggestion
    "minimum-stability": "rc",
```

composer.json
```diff
@@ -19,39 +19,41 @@
...
-        "symfony/asset": "^5.3",
-        "symfony/console": "^5.3",
-        "symfony/dotenv": "^5.3",
-        "symfony/expression-language": "^5.3",
-        "symfony/flex": "^1.1",
-        "symfony/form": "^5.3",
-        "symfony/framework-bundle": "^5.3",
-        "symfony/intl": "^5.3",
-        "symfony/mailer": "^5.3",
+        "symfony/asset": "^5.4.0-beta1",
```
derrabus, line 22: I think, you can change all of those to…

```suggestion
        "symfony/asset": "^5.4",
```

composer.json
```diff
@@ -89,8 +91,8 @@
     },
     "extra": {
         "symfony": {
-            "require": "5.3.*",
-            "allow-contrib": true
+            "allow-contrib": true,
+            "require": "^5.4.0-rc1"
```
derrabus, line 95:
```suggestion
            "require": "5.4.*"
```

javiereguiluz:
We're running deprecation-free. I had to set `tgalopin/html-sanitizer-bundle` to `dev-master`, but we'll have a new stable version soon, so this is temporary.

composer.json
```diff
@@ -3,7 +3,7 @@
     "license": "MIT",
     "type": "project",
     "description": "Symfony Demo Application",
-    "minimum-stability": "stable",
+    "minimum-stability": "rc",
```
derrabus, line 6:
```suggestion
    "minimum-stability": "RC",
```

Apparently, the composer linter is picky about this. 🙈

composer.json
stof: @javiereguiluz this does not seem removed

composer.json
GromNaN: Fortunately, it will be lowercase `stable` soon.

javiereguiluz: Fixed. Thanks.

derrabus (approved):
`phpstan --generate-baseline` and we're good to go, I guess. 🎉

PR #1269, opened 2021-11-17
Update SecurityController.php

seb-jean:
I propose this PR to be in agreement with maker-bundle:
https://github.com/symfony/maker-bundle/blob/main/src/Security/SecurityControllerBuilder.php#L38


javiereguiluz:
Thank you @seb-jean.

src/Controller/SecurityController.php
```diff
@@ -36,7 +36,7 @@ class SecurityController extends AbstractController
     public function login(Request $request, Security $security, AuthenticationUtils $helper): Response
```
yceruto, line 36: after this change the `Security $security` can be removed

PR #1270, opened 2021-11-19
Update German translation

ihmels:
- Sort entries so that they match the order in the English translation
- Improve some translations
- Add missing translations


javiereguiluz:
Yannick, thanks for improving and completing this translation! *Danke schön*

PR #1273, opened 2021-11-25
Update PHPStan baseline

derrabus:
The PHPUnit baseline apparently has been created with an older version of PHPStan. I have recreated it with the latest version 1.2.0 which is also currently used by the CI.


javiereguiluz:
Thank you

PR #1275, opened 2021-11-27
WIP: Feature/dockerize

mehdibo:
I am working on Dockerizing the demo, this is a starter but still need to improve some stuff.

* Is this supposed to be used for dev? meaning is the code going to change while the container is running?
* About the frontend part, should it be built when running the containers or just use the pushed build?
* What's the decision about the database? should I move it to Postgres or leave SQLite?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ ARG CADDY_VERSION=2
 
 ARG APP_ENV=dev
 
-FROM php:${PHP_VERSION}-fpm-alpine AS  symfony_php
+FROM php:${PHP_VERSION}-fpm-alpine AS symfony_php
 
 WORKDIR /srv/symfony
 VOLUME /var/run/php
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,69 @@
+# This Dockerfile contains different images
+
+ARG PHP_VERSION=8.0
+ARG CADDY_VERSION=2
+
+# Build PHP FPM
+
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,69 @@
+# This Dockerfile contains different images
+
+ARG PHP_VERSION=8.0
+ARG CADDY_VERSION=2
+
+# Build PHP FPM
+
+ARG APP_ENV=dev
+
+FROM php:${PHP_VERSION}-fpm-alpine AS  symfony_php
```
creocoder, line 10:
```suggestion
FROM php:${PHP_VERSION}-fpm-alpine AS symfony_php
```

COil:
Hello @creocoder, you will find related answers from Javier here : https://github.com/symfony/demo/pull/931#pullrequestreview-192161914 

>  Is this supposed to be used for dev? meaning is the code going to change while the container is running?

I think yes.

> About the frontend part, should it be built when running the containers or just use the pushed build?

Pushed build. But the containers should be able to regenerate the assets IMHO.

>  What's the decision about the database? should I move it to Postgres or leave SQLite?

Leave SQLite.

COil:
Regarding the other PR, I'm not sure it worths working on this.

mehdibo:
Hello @COil 

That's my understanding after @javiereguiluz explained to me the debate,
If however the community accepts this as a Good enough to Dockerize the demo I'm ready to improve it.

PR #1277, opened 2021-12-02
Upgrade to Symfony 6.0

javiereguiluz:


composer.json
```diff
@@ -7,10 +7,11 @@
     "prefer-stable": true,
     "replace": {
         "symfony/polyfill-php72": "*",
-        "symfony/polyfill-php73": "*"
+        "symfony/polyfill-php73": "*",
+        "symfony/polyfill-php74": "*"
```
stof, line 11: you can also add `symfony/polyfill-php80` here

javiereguiluz: Yes! Done. Thanks.

config/packages/assets.yaml
```diff
@@ -1,4 +1,4 @@
 framework:
     assets:
-        json_manifest_path: '%kernel.project_dir%/public/build/manifest.json'
+        base_path: '/'
         strict_mode: '%kernel.debug%'
```
stof, line 4: strict_mode is now dead code, as it is used only for the JsonManifestStrategy

javiereguiluz: OK. I've removed it. Thanks.

However, I don't understand why `json_manifest_path` worked before but it's failing now.

stof: it is still working now in non-strict mode, by being a passthrough.

Before, there was no strict mode.

composer.json
```diff
@@ -20,25 +22,25 @@
         "erusev/parsedown": "^1.6",
         "sensio/framework-extra-bundle": "^6.1",
```
rosier, line 23: Bump other dependencies to versions that allow Symfony 6
```suggestion
        "sensio/framework-extra-bundle": "^6.2",
```

composer.json
```diff
@@ -20,25 +22,25 @@
...
         "symfony/flex": "^1.17.1",
-        "symfony/form": "^5.4",
-        "symfony/framework-bundle": "^5.4",
-        "symfony/intl": "^5.4",
-        "symfony/mailer": "^5.4",
+        "symfony/form": "^6.0",
+        "symfony/framework-bundle": "^6.0",
+        "symfony/intl": "^6.0",
+        "symfony/mailer": "^6.0",
         "symfony/monolog-bundle": "^3.1",
```
rosier, line 34:
```suggestion
        "symfony/monolog-bundle": "^3.7.1",
```

composer.json
```diff
@@ -20,25 +22,25 @@
...
-        "symfony/translation": "^5.4",
-        "symfony/twig-bundle": "^5.4",
-        "symfony/validator": "^5.4",
+        "symfony/runtime": "^6.0",
+        "symfony/security-bundle": "^6.0",
+        "symfony/string": "^6.0",
+        "symfony/translation": "^6.0",
+        "symfony/twig-bundle": "^6.0",
+        "symfony/validator": "^6.0",
         "symfony/webpack-encore-bundle": "^1.4",
```
rosier, line 42:
```suggestion
        "symfony/webpack-encore-bundle": "^1.13",
```

composer.json
```diff
@@ -47,17 +49,17 @@
     "require-dev": {
         "dama/doctrine-test-bundle": "^6.7",
         "doctrine/doctrine-fixtures-bundle": "^3.0",
-        "symfony/browser-kit": "^5.4",
-        "symfony/css-selector": "^5.4",
-        "symfony/debug-bundle": "^5.4",
+        "symfony/browser-kit": "^6.0",
+        "symfony/css-selector": "^6.0",
+        "symfony/debug-bundle": "^6.0",
         "symfony/maker-bundle": "^1.11",
```
rosier, line 55:
```suggestion
        "symfony/maker-bundle": "^1.35",
```

composer.json
```diff
@@ -20,25 +22,25 @@
...
+        "symfony/string": "^6.0",
+        "symfony/translation": "^6.0",
+        "symfony/twig-bundle": "^6.0",
+        "symfony/validator": "^6.0",
         "symfony/webpack-encore-bundle": "^1.4",
-        "symfony/yaml": "^5.4",
+        "symfony/yaml": "^6.0",
         "tgalopin/html-sanitizer-bundle": "^1.4",
         "twig/extra-bundle": "^3.0",
         "twig/intl-extra": "^3.0",
```
rosier, line 46:
```suggestion
        "twig/intl-extra": "^3.3.4",
```

composer.json
```diff
@@ -20,25 +22,25 @@
...
+        "symfony/security-bundle": "^6.0",
+        "symfony/string": "^6.0",
+        "symfony/translation": "^6.0",
+        "symfony/twig-bundle": "^6.0",
+        "symfony/validator": "^6.0",
         "symfony/webpack-encore-bundle": "^1.4",
-        "symfony/yaml": "^5.4",
+        "symfony/yaml": "^6.0",
         "tgalopin/html-sanitizer-bundle": "^1.4",
         "twig/extra-bundle": "^3.0",
```
rosier, line 45:
```suggestion
        "twig/extra-bundle": "^3.3.4",
```

composer.json
```diff
@@ -47,17 +49,17 @@
     "require-dev": {
         "dama/doctrine-test-bundle": "^6.7",
         "doctrine/doctrine-fixtures-bundle": "^3.0",
```
rosier, line 51:
```suggestion
        "doctrine/doctrine-fixtures-bundle": "^3.4.1",
```

javiereguiluz: Thanks! I've updated this and all the other dependency versions.

94noni:
Great PR :)
Also it could be great to add in the readme that the demo uses latest symfony version, and also upgrade the php required version mentionned
:)

javiereguiluz:
@stof what do you propose to fix the bugs reported by the "Lint / Linters (8.0)" GitHub action?

* Ignore them in this PR, but fix them in a separate PR before the next stable release (the first one compatible with 6.0)
* Ignore them and release the next stable release without fixing them (fix them in the future)
* Fix them in this PR
* Tweak the PHPStan config to remove most of these "bugs"
* other?

Thanks!

stof:
Most of the errors have 2 sources:

- Symfony 6 provides more precise return types, which means that some baseline errors have disappeared entirely
- the CI now runs phpstan 1, which changed some error message (talking about `type` rather than `typehint`) meaning the baseline does not match anymore

In both cases, the fix for this PR is the same: running `phpstan analyze --generate-baseline` to regenerate the baseline and commit the new phpstan-baseline.neon file.

stof:
btw, phpstan should be installed through require-dev instead of relying on a global install in github actions, so that the second point (changes due to changing the version of phpstan) only happens when actually updating the phpstan dependency.

javiereguiluz:
Christophe, I've done what you proposed ... and it worked like a charm 😍  Thanks a lot! I think this PR is now ready.

PR #1279, opened 2021-12-03
Fix $passwordHasher

seb-jean:


javiereguiluz:
Thanks for fixing this bug @seb-jean.

PR #1282, opened 2021-12-10
doc: v2 use PHP 8.0

bchatard:
Update README, according to v2 change (see `composer.json`)


94noni (approved):

javiereguiluz:
Good catch, thanks Brice.

PR #1288, opened 2021-12-15
Add Faker to Fixtures

nasrbechir:
I added Faker to avoid putting the data manually in Fixtures

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
         "doctrine/doctrine-migrations-bundle": "^3.0",
         "doctrine/orm": "^2.10",
         "erusev/parsedown": "^1.7",
+        "fzaninotto/faker": "^1.5",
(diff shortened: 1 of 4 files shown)


javiereguiluz:
@nasrbechir I thank you this contribution effort, but I'm afraid we need to close this without merging for these reasons:

* This library is archived/abandoned ... so we'd need to use one of its forks
* I've downloaded the latest version as a ZIP file and it takes 3.5 MB. It's way too much for a library that we'll barely use.
* People like @weaverryan and others are advocating for a different way of creating fixtures. Instead of this, they recommend https://github.com/zenstruck/foundry We're not using it in this project yet because it's not a standard practice in Symfony too, but we might to.

So, that's why we're keeping the current fixtures, but I hope you understand it. Thanks!

PR #1289, opened 2021-12-16
Leverage PHP 8 attributes to configure Command

94noni:
I wanted to test PHP 8 install on my new machine, thus tried some features on this project as it now requires v8
I followed https://symfony.com/blog/new-in-symfony-5-3-lazy-command-description
Feel free to close, but I think it could be useful to showcase new way on this Symfony demo project as it is kind off an entry point, also Controllers uses this feature


javiereguiluz:
I like this proposal. IT might look "edgy", but it's the latest recommended Symfony practice, so let's follow it in this project too. Thanks!

PR #1291, opened 2021-12-16
Add various comments in the codebase

94noni:
Add some comments while reading some source code
Ref https://symfony.com/doc/current/setup.html#the-symfony-demo-application

### DIFF
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -30,7 +30,7 @@ use Symfony\Component\Routing\Annotation\Route;
  * purposes. However, in your real Symfony application you should use any of the
  * existing bundles that let you generate ready-to-use backends without effort.
  *
- * See http://knpbundles.com/keyword/admin
+ * See https://symfony.com/bundles/EasyAdminBundle/current/index.html
  *
  * @author Ryan Weaver <weaverryan@gmail.com>
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
--- a/src/Controller/BlogController.php
+++ b/src/Controller/BlogController.php
@@ -81,6 +81,13 @@ class BlogController extends AbstractController
         // have enabled the DebugBundle. Uncomment the following line to see it in action:
         //
         // dump($post, $this->getUser(), new \DateTime());
+        //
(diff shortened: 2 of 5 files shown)


src/Controller/Admin/BlogController.php
```diff
@@ -30,6 +30,7 @@
  * purposes. However, in your real Symfony application you should use any of the
  * existing bundles that let you generate ready-to-use backends without effort.
  *
+ * See https://symfony.com/bundles/EasyAdminBundle/current/index.html
  * See http://knpbundles.com/keyword/admin
```
stof, line 34: I suggest removing that knpbundles.com link as knpbundles.com does not seem functional anymore (it has 0 bundles referenced in it)

src/Controller/BlogController.php
```diff
@@ -75,12 +75,17 @@ public function index(Request $request, int $page, string $_format, PostReposito
     #[Route('/posts/{slug}', methods: ['GET'], name: 'blog_post')]
     public function postShow(Post $post): Response
     {
-        // Symfony's 'dump()' function is an improved version of PHP's 'var_dump()' but
+        // Symfony's 'dump()' and 'dd()' functions are an improved version of PHP's 'var_dump()' but
```
stof, line 78: ``dd` is not just an improved `var_dump`. It is dump + die, which is quite different that just dumping.

94noni: you are right, now the phrasing is weird
do you think I revert and adding the `dd()` is not worth it? or keep it but rephrase?
thank you @stof

javiereguiluz:
Thanks Antoine!

src/Controller/Admin/BlogController.php
```diff
@@ -30,7 +30,7 @@
  * purposes. However, in your real Symfony application you should use any of the
  * existing bundles that let you generate ready-to-use backends without effort.
  *
- * See http://knpbundles.com/keyword/admin
+ * See https://symfony.com/bundles/EasyAdminBundle/current/index.html
```
javiereguiluz, line 33: I've replaced this link by `https://symfony.com/bundle` There are several admin bundles, so it's not fair to link specifically to one of them 👍

PR #1292, opened 2021-12-16
Move JavaScript entry function in the HTML head

94noni:
Ref https://symfony.com/doc/current/frontend/encore/simple-example.html#configuring-encore-webpack


javiereguiluz:
This is important to follow latest Symfony recommended practices. Thanks!

PR #1293, opened 2021-12-16
Add documentation on frontend controllers as application entrypoints

94noni:
Some nice to have comments (for new comers), perhaps the wording is to be improved :s 
Ref https://symfony.com/doc/current/setup.html#the-symfony-demo-application

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -10,6 +10,15 @@ if (!is_file(dirname(__DIR__).'/vendor/autoload_runtime.php')) {
 
 require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
 
+/**
+ * This file is a Symfony Front Controller (entrypoint)
+ * for the command-line from a terminal.
+ *
+ * It boots Symfony in a CLI context.
+ * 
+ * See https://symfony.com/doc/current/console.html
+ */
+
 return function (array $context) {
     $kernel = new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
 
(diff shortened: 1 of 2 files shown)


bin/console
```diff
@@ -10,6 +10,15 @@ if (!is_file(dirname(__DIR__).'/vendor/autoload_runtime.php')) {
 
 require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
 
+/**
+ * This file is a Symfony Front Controller (entrypoint)
```
javiereguiluz, line 14: Just asking: is it correct to use the term "front controller" when we are not in the HTTP/web context? Thanks.

94noni: Well this is a good question I can not answer
Entrypoint is ok, perhaps we could just write this

the aim or this PR is to show the 2 way or booting symfony

94noni: @javiereguiluz hello

in this page: https://symfony.com/doc/current/the-fast-track/en/3-zero.html#initializing-the-project

`The bin/ directory contains the main CLI entry point: console. You will use it all the time.`

so what do you think I rewrite: `This file is a Symfony CLI entrypoint`

bin/console
```diff
@@ -10,6 +10,14 @@ if (!is_file(dirname(__DIR__).'/vendor/autoload_runtime.php')) {
 
 require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
 
+/**
+ * This file is a Symfony command-line entrypoint.
```
rosier, line 14: How about also adding the word interface and shouldn't entrypoint be two words?

```suggestion
 * This file is a Symfony command-line interface (CLI) entry point.
```

94noni: indeed, makes more sens to really document with full names thx

PR #1294, opened 2021-12-17
Swap markdown package from erusev to league

94noni:
Ref https://packagist.org/packages/league/commonmark/stats VS https://packagist.org/packages/erusev/parsedown/stats


javiereguiluz:
I like this change. `erusev` parser is faster, but `commonmark` is the reference implementation, so let's use it.

Note that no code changes are necessary because Twig extension automatically detects several Markdown libraries and picks one of them.

PR #1297, opened 2021-12-17
Add example and documentation of logging

94noni:
I found no example of log inside the demo app, I think it could be wise to showcase it
Ref https://symfony.com/doc/current/index.html in the `The Basics` toc


javiereguiluz:
I like it! Thanks Antoine

PR #1298, opened 2021-12-17
Fix translator parameters

94noni:
Even if it works with`%var%` I think without is better, wdyt?
Ref https://symfony.com/doc/current/translation/message_format.html
Also added a direct link to doc

### DIFF
--- a/src/EventSubscriber/CommentNotificationSubscriber.php
+++ b/src/EventSubscriber/CommentNotificationSubscriber.php
@@ -52,8 +52,8 @@ class CommentNotificationSubscriber implements EventSubscriberInterface
 
         $subject = $this->translator->trans('notification.comment_created');
         $body = $this->translator->trans('notification.comment_created.description', [
-            '%title%' => $post->getTitle(),
-            '%link%' => $linkToPost,
+            'title' => $post->getTitle(),
+            'link' => $linkToPost,
         ]);
 
         // See https://symfony.com/doc/current/mailer.html
--- a/config/packages/translation.yaml
+++ b/config/packages/translation.yaml
@@ -1,6 +1,9 @@
+# See https://symfony.com/doc/current/translation.html
 framework:
     default_locale: '%locale%'
     translator:
+        # Translations are defined using the ICU Message Format
+        # See https://symfony.com/doc/current/translation/message_format.html
         default_path: '%kernel.project_dir%/translations'
         fallbacks:
             - '%locale%'
--- a/templates/admin/blog/edit.html.twig
+++ b/templates/admin/blog/edit.html.twig
@@ -3,7 +3,7 @@
 {% block body_id 'admin_post_edit' %}
 
 {% block main %}
-    <h1>{{ 'title.edit_post'|trans({'%id%': post.id}) }}</h1>
+    <h1>{{ 'title.edit_post'|trans({'id': post.id}) }}</h1>
 
     {{ include('admin/blog/_form.html.twig', {
         form: form,
--- a/templates/bundles/TwigBundle/Exception/error.html.twig
+++ b/templates/bundles/TwigBundle/Exception/error.html.twig
@@ -13,13 +13,13 @@
 {% block body_id 'error' %}
 
 {% block main %}
-    <h1 class="text-danger">{{ 'http_error.name'|trans({ '%status_code%': status_code }) }}</h1>
+    <h1 class="text-danger">{{ 'http_error.name'|trans({ 'status_code': status_code }) }}</h1>
 
     <p class="lead">
-        {{ 'http_error.description'|trans({ '%status_code%': status_code }) }}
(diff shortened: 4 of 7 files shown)


src/EventSubscriber/CommentNotificationSubscriber.php
```diff
@@ -50,10 +50,11 @@ public function onCommentCreated(CommentCreatedEvent $event): void
             '_fragment' => 'comment_'.$comment->getId(),
         ], UrlGeneratorInterface::ABSOLUTE_URL);
 
+        // See https://symfony.com/doc/current/translation.html
         $subject = $this->translator->trans('notification.comment_created');
         $body = $this->translator->trans('notification.comment_created.description', [
-            '%title%' => $post->getTitle(),
-            '%link%' => $linkToPost,
+            'title' => $post->getTitle(),
+            'link' => $linkToPost,
```
bocharsky-bw, line 57: IIRC it's still best practice to use `%%` for translation params, at least the link you provided uses `%link%` instead of `link`, so why should we remove it?

94noni: I am reading in the link:

// prints "Hello Fabien!"
echo $translator->trans('say_hello', ['name' => 'Fabien']);

// prints "Hello Symfony!"
echo $translator->trans('say_hello', ['name' => 'Symfony']);

bocharsky-bw: Hm, are you sure you're on the "current"? i.e. https://symfony.com/doc/current/translation.html

I don't see "say_hello" at all, but see:
Screenshot 2021-12-20 at 15 37 36

94noni: @bocharsky-bw now I got it ...
I am on the https://symfony.com/doc/current/translation/message_format.html (as in PR description) not the one you answered me (as in PR code for documentation of translations)

And the subpart at https://symfony.com/doc/current/translation/message_format.html#message-placeholders

for ICU translations messages => **as it is what this demo app uses**
Capture d’écran 2021-12-20 à 14 45 53

perhaps I should put both links?
thank you

bocharsky-bw: Aha, now I see. Changes look good then, but I'm not sure about the links. We use translations all over the project code, and having the link only in this spot sounds incomplete to me, but on the other hand having the link in all the spots where we use translations is overkill probably. Maybe we don't need that link at all?

Also, if we want to get rid of `%%` in this spot, probably we should do the same on other spots too?

javiereguiluz:
But if I'm right, and @wouterj could confirm this because he wrote that doc, the MessageFormatter format without `%` is recommended when you use those MessageFormatter files, which we are not using (we still use traditional Symfony translation files).

94noni:
@javiereguiluz this demo app uses `messages+intl-icu.en.xlf` [doc](https://symfony.com/doc/current/translation/message_format.html)

according to this [doc](https://symfony.com/doc/current/translation/message_format.html#message-placeholders), the placeholder defined [here](https://github.com/symfony/demo/blob/d863b68c6088c8faca4daa0e37e6eb93d7975b9c/translations/messages%2Bintl-icu.en.xlf#L359) should be replaced like the [doc screened here](https://github.com/symfony/demo/pull/1298#discussion_r772378558)

perhaps I am missing something but I really think this is legit

wouterj:
Yes, using `%...%` is only recommended without the ICU formatter. However, this package is using the ICU formatter (indicated by the `+intl-icu` suffix in the filename). I would also say that using the ICU formatter is the recommended method for Symfony 5.0+ applications, so it makes sense for the demo application to use these.

However, let's remove `%` from all translation parameter keys in that case, I'm almost certain translator is used more than once? (given the amount of messages in the catalogues).

94noni:
@wouterj 

> I'm almost certain translator is used more than once? 

in fact as [written here](https://github.com/symfony/demo/pull/1298#discussion_r772425612), in this demo app php code, this is the only code all others are in twig templates

wouterj:
> in fact as written here, in this demo app php code, this is the only code all others are in twig templates

Let's remove them from the twig templates as well :)

94noni:
but they are needed in [twig tplts](https://symfony.com/doc/current/translation.html#using-twig-filters) @wouterj

wouterj:
They aren't, the Twig filter indirectly uses `TranslatorInterface::trans()`. Inside this, `trim($key, '%{ }')` is used to trim the remaining percent signs (inside the `IntlFormatter` to be really honest).

So that's mostly a piece of docs that we forgot to update :)

94noni:
oki I got it, lets remove then :)

javiereguiluz:
Antoine, thanks a lot for updating all this! And thanks to our lovely reviewers too for their help here 🙏

PR #1303, opened 2021-12-23
Also run tests on php 8.1

rosier:


javiereguiluz:
Yes, this is important! Thanks Jan

PR #1304, opened 2021-12-24
Update code to use the latest Node.js LTS version (16)

fabpot:


javiereguiluz:
Thanks Fabien!

Please note that I've removed the `Node.js 16+` from tech requirements, because this project distributes the web assets already compiled. This is done precisely to reduce friction by not forcing developers to install any Node version or any node packages and to not compile any assets before using this app if they don't want to.

PR #1308, opened 2022-01-03
Disable password manager for the login form

javiereguiluz:
Today I received a new security notification from Google saying that some passwords of mine were breached ... they were the passwords of this Symfony Demo application. It's not the first time, so I looked for possible solutions to prevent storing this password.

Apparently it's impossible to do it with the autocomplete attribute (https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) because modern browsers don't respect it when set to `off` in a password field.

So, I used the solution explained here:
https://stackoverflow.com/questions/41945535/html-disable-password-manager/41946532

I've tested this solution successfully with both Chrome and Firefox. What do you think?

### DIFF
--- a/templates/security/login.html.twig
+++ b/templates/security/login.html.twig
@@ -34,6 +34,16 @@
                             <i class="fa fa-sign-in" aria-hidden="true"></i> {{ 'action.sign_in'|trans }}
                         </button>
                     </fieldset>
+
+                    <!-- by default, modern browsers ask users to store passwords. This behavior is correct in
+                         real sites, but it's nonsensical in this demo project, where storing this password
+                         will result in unneeded security notifications about 'breached passwords'.
+                         Sadly it's not possible to change or disable this browser behavior (even 'autocomplete=off'
+                         doesn't work). That's why we use the trick of adding many password fields to the form
+                         (only one of them is real and the ones below are unused by the application). When a form
+                         includes many password fields, browsers no longer ask to store the password.  -->
+                    <input type="password" id="disable-password-manager-1" style="display: none;" value="stop-password-manager-1"/>
+                    <input type="password" id="disable-password-manager-2" style="display: none;" value="stop-password-manager-2"/>
                 </form>
             </div>
         </div>


jkufner:
Bad idea.

Storing passwords in a password manager is completely valid and in fact more secure than remembering the passwords.

Google's false positives should be reported to Google and solved there. (Probably won't happen.) It would be better to use a better password, that is not in a dictionary and considered breached to avoid these false positives.

Having a piece of code that breaks password manager is wrong, having that in an app that should represent best practices is exceptionally wrong.

And most importantly, some idiots ~may~ will leave this hack in the real applications when they use the demo as a starting point. That is too dangerous to risk.

javiereguiluz:
@jkufner please, read the comment with the explanation that added this PR:

```
by default, modern browsers ask users to store passwords. This behavior is correct in
real sites, but it's nonsensical in this demo project, where storing this password
will result in unneeded security notifications about 'breached passwords'.

[...]
```

You are saying what we already said: storing passwords is a good practice. The only difference is that this is a local non-real app and storing these passwords is bad and creates a bad user experience. That's why we propose this change.

I don't think the other idea that you propose is viable (to ask Google to add these passwords to the allow list of bad passwords)

jkufner:
This demo showcases the best practices; disabling password manager is not a best practice. No explanation can change this conflict. The problem is, that the solution works in real apps based on this demo and it will stay in many of them. A few false positives from Google is not really a problem.

pbowyer:
@javiereguiluz I've read the comment you quote above in https://github.com/symfony/demo/pull/1308#issuecomment-1003995569 and disagree. I store passwords for demo projects & more: it means I don't have to think at all if the password is "testing", "testing123", "password", "admin" or any other of the passwords used by demo projects.

javiereguiluz:
@pbowyer that's indeed something important ... that's why in this project, the password is hardcoded in the `<input type="password">`. This is again something that you wouldn't do in a real app, but we do it because this is a demo app and we want to provide an excellent user experience.

javiereguiluz:
I still think that this is a good idea to (slightly) improve UX ... but it's clear that I won't be able to convince anyone about this ... so let's close it. Thanks!

PR #1310, opened 2022-01-12
Remove the dependency on composer/package-versions-deprecated

stof:
This package was added in the past to force keeping support for composer 1 when using doctrine/orm. In the meantime, doctrine/orm has switched to using that package directly, making this root requirement useless. And in the next releases of doctrine/orm and doctrine/dbal, they are migrating to use composer-runtime-api (dropping support for composer 1) instead of using this plugin, as part of the effort to fade out that compat plugin.

Refs https://github.com/symfony/symfony/issues/44726

The package is still present for now (and so still needed in the list of allowed plugins too) because doctrine packages require it. But this will change in their next releases for DBAL, ORM and Migrations (the PRs are already merged)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,6 @@
     "require": {
         "php": ">=8.0.2",
         "ext-pdo_sqlite": "*",
-        "composer/package-versions-deprecated": "^1.8",
         "doctrine/dbal": "^3.1",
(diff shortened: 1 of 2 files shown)


GromNaN (approved):

javiereguiluz:
Thank you Christophe.

PR #1312, opened 2022-02-05
Use logout_path() helper + add CSRF protection to logout

wouterj:
The login form already had CSRF protection, but logout didn't yet in this demo application. Generally, I think it's a good idea to add CSRF protection to logout. At the very least, it avoids annoying situations where a website can force users to logout from your service on each visit. But depending on the type of application, things can also get more serious and cause actual security issues when CSRF on logout isn't enabled.

Fortunately, CSRF protection on logout is quite easy using the `logout_path()` helper: it automatically knows the logout URL of the current firewall and it automatically adds the correct CSRF token to the URL. I think the logout path/url helpers are little known gems in Symfony, so let's showcase them :)

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -50,6 +50,10 @@ security:
                 # The name of the route to redirect to after logging out
                 target: homepage
 
+                # Secure the logout against CSRF
+                csrf_parameter: logout
+                csrf_token_generator: security.csrf.token_manager
+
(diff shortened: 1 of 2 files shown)


seb-jean:
I didn't know about this `logout_path()` helper.
So thank you @wouterj.

javiereguiluz:
Thank you Wouter.

PR #1315, opened 2022-02-08
Update PHP doc about annotations IsGranted

adpauly:
The doc now suggests to use PHP attributes instead of "old" annotation for "IsGranted" instruction.
Sorry if my PR format doesn't match with your requirements. I just wanna help. :)

### DIFF
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -112,7 +112,7 @@ class BlogController extends AbstractController
     public function show(Post $post): Response
     {
         // This security check can also be performed
-        // using an annotation: @IsGranted("show", subject="post", message="Posts can only be shown to their authors.")
+        // using a PHP attribute: #[IsGranted('show', subject: 'post', message: 'Posts can only be shown to their authors.')]
         $this->denyAccessUnlessGranted(PostVoter::SHOW, $post, 'Posts can only be shown to their authors.');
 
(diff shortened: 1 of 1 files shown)


94noni (approved):

src/Controller/Admin/BlogController.php
```diff
@@ -112,7 +112,7 @@ public function new(Request $request, EntityManagerInterface $entityManager): Re
     public function show(Post $post): Response
     {
         // This security check can also be performed
-        // using an annotation: @IsGranted("show", subject="post", message="Posts can only be shown to their authors.")
+        // using a PHP attribute: #[IsGranted("show", subject="post", message="Posts can only be shown to their authors.")]
```
bocharsky-bw, line 115: I think we should write params with `:` to match the same style in other spots, e.g. in https://github.com/symfony/demo/blob/f1ee7c6b8a70f8fed2ca032052a7760d58299cef/src/Controller/Admin/BlogController.php#L127

```suggestion
        // using a PHP attribute: #[IsGranted('show', subject: 'post', message: 'Posts can only be shown to their authors.')]
```

adpauly: For sure! Sorry for the oversight.

javiereguiluz:
Nice catch @adpauly ! Thanks for fixing this.

PR #1318, opened 2022-02-20
Propose to add a reference link to Symfony book

94noni:
I think it is a nice addition here to add such reference
Ref https://symfony.com/blog/a-week-of-symfony-790-14-20-february-2022


javiereguiluz:
This is a great idea! Thanks Antoine for proposing this improvement.

PR #1324, opened 2022-03-29
Update twig to fix a security issue & Add unit tests

nasrbechir:


94noni:
I think 2 separated PRs should be better, composer upgrade and tests additions :)

PR #1327, opened 2022-04-19
bump dama/doctrine-test-bundle to ^7.0

dmaicher:
See https://github.com/dmaicher/doctrine-test-bundle/releases/tag/v7.0.0


javiereguiluz:
Thanks {{ name }} and thanks for maintaining that bundle!

PR #1330, opened 2022-05-05
Display source code using dark mode

javiereguiluz:
Creating a "dark mode" style for this application is outside of the scope of the project, so we're not going to do it.

However, I propose to use a "dark mode" style to display the source code in the modal windows. In addition to being the most common style for developers, showing it in "dark mode" further emphasizes the difference between the application interface and its code:

**Before / After**

code-highlight-comparison


bocharsky-bw:
Looks  better 👍

EdwinLove:
👍

PR #1331, opened 2022-05-05
Fix issue reported by PHPStan

javiereguiluz:
The original author of this fix is @dkarlovi, who kindly provided the solution in https://github.com/javiereguiluz/demo/pull/1/files

Thanks!


dkarlovi:
The new error should be fixed by updating the PHPStan baseline, an error previously reported is no longer reported.

PR #1332, opened 2022-05-05
Update dependencies

javiereguiluz:


phpstan-baseline.neon
```diff
@@ -80,6 +85,11 @@ parameters:
 			count: 1
 			path: src/Controller/Admin/BlogController.php
 
+		-
+			message: "#^Call to an undefined method object\\:\\:getName\\(\\)\\.$#"
```
stof, line 89: This would probably be fixed if you add the `@extends ServiceEntityRepository<User>` tag on the UserRepository to specify the generic type (and same for other repositories adapting the entity class)

javiereguiluz: Thanks for the insight.

However, to be honest, I'm not sure about adding that 😐  From the point of view of Symfony, we already tell that the repository class is about `User::class`. Repeating that information in a comment/annotation to help a specific tool looks like too much to me.

Given that this is the official Symfony Demo, it's all about showcasing Symfony ... so tools like static analyzers should be a distant secondary issue. In any case, I'm open to hearing other opinions. Thanks!

stof: Well, it also helps IDEs that support generic classes (PHPStorm has partial support for them, improving it in each release).
Generics are actually a de-facto standard, being supported with the same syntax by phpstan, psalm, PHPStorm (and maybe VSCode but I haven't checked)

And MakerBundle generates that `@extends` tag now. So users generating a User entity today will have it.

PR #1336, opened 2022-05-05
Use attributes to define ORM metadata

javiereguiluz:


src/Entity/Post.php
```diff
@@ -33,77 +30,57 @@
...
     private Collection $comments;
 
     /**
      * @var Tag[]|Collection
-     *
-     * @ORM\ManyToMany(targetEntity="App\Entity\Tag", cascade={"persist"})
-     * @ORM\JoinTable(name="symfony_demo_post_tag")
-     * @ORM\OrderBy({"name": "ASC"})
      */
+    #[ORM\OneToMany(targetEntity: Tag::class, mappedBy: 'post', cascade: ['persist'])]
```
stof, line 81: the bug is that you added `mappedBy: 'post'` here, which is wrong (and was not there in the annotation

stof: btw, the linter job gives you the error directly in the "lint doctrine entities" failure.

javiereguiluz: I added the `mappedBy` property because otherwise I see this error:

doctrine-error

voronkovich: @javiereguiluz, You changed the relation type from ManyToMany to OneToMany.

```suggestion
    #[ORM\ManyToMany(targetEntity: Tag::class, cascade: ['persist'])]
```

javiereguiluz: Yes!!! Super nice catch! That was it. Thanks a lot Oleg 🙇🙇  Fixed!

PR #1337, opened 2022-05-05
Tweak source code modal for large screens

voronkovich:
I suggest to make source code modal wider on large screens to avoid word wrapping.

Screenshot 2022-05-05 at 23-40-14 Symfony Demo application

Bootstrap 5 has [.modal-xl](https://getbootstrap.com/docs/5.1/components/modal/#optional-sizes) CSS class for this. But we use Bootstrap 3, so I just created a "polyfill" for that feature.


javiereguiluz:
This is a nice improvement! Thanks Oleg.

PR #1339, opened 2022-05-12
Upgrade to Symfony 6.1

javiereguiluz:
Don't mind much about the `sanitize_html` Twig filter. I added it quickly to make the application work. We're discussing about how adding again this Twig filter in the Symfony HtmlSanitizer component.

-----

These are the remaining deprecations:

```
Remaining indirect deprecation notices (6)

  1x: The "DAMA\DoctrineTestBundle\Doctrine\DBAL\StaticDriver" class implements "Doctrine\DBAL\VersionAwarePlatformDriver" that is deprecated All drivers will have to be aware of the server version in the next major release.
    1x in PHPUnitExtension::executeBeforeFirstTest from DAMA\DoctrineTestBundle\PHPUnit

  1x: The "Symfony\Bridge\Doctrine\Logger\DbalLogger" class implements "Doctrine\DBAL\Logging\SQLLogger" that is deprecated Use {@see \Doctrine\DBAL\Logging\Middleware} or implement {@see \Doctrine\DBAL\Driver\Middleware} instead.

  1x: Method "ArrayAccess::offsetExists()" might add "bool" as a native return type declaration in the future. Do the same in implementation "Dflydev\DotAccessData\Data" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in BlogControllerTest::testAdminShowPost from App\Tests\Controller\Admin

  1x: Method "ArrayAccess::offsetGet()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Dflydev\DotAccessData\Data" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in BlogControllerTest::testAdminShowPost from App\Tests\Controller\Admin

  1x: Method "ArrayAccess::offsetSet()" might add "void" as a native return type declaration in the future. Do the same in implementation "Dflydev\DotAccessData\Data" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in BlogControllerTest::testAdminShowPost from App\Tests\Controller\Admin

  1x: Method "ArrayAccess::offsetUnset()" might add "void" as a native return type declaration in the future. Do the same in implementation "Dflydev\DotAccessData\Data" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in BlogControllerTest::testAdminShowPost from App\Tests\Controller\Admin
```

@dmaicher do you know how can we remove the deprecation notice about DoctrineTestBundle? Thanks!

@derrabus do you know how can we fix the deprecation about `Symfony\Bridge\Doctrine\Logger\DbalLogger`? Thanks!


stof:
For the DbalLogger, the work is in progress at https://github.com/doctrine/DoctrineBundle/pull/1517

src/Twig/AppExtension.php
```diff
@@ -64,4 +71,11 @@ public function getLocales(): array
 
         return $this->locales;
     }
+
+    public function sanitize(string $unsafeHtmlContents): string
+    {
+        $htmlSanitizer = new HtmlSanitizer((new HtmlSanitizerConfig())->allowSafeElements());
```
stof, line 77: you should rather showcase using the service configured by FrameworkBundle (by injecting the HtmlSanitizerInterface)

dmaicher:
@javiereguiluz regarding

```
  1x: The "DAMA\DoctrineTestBundle\Doctrine\DBAL\StaticDriver" class implements "Doctrine\DBAL\VersionAwarePlatformDriver" that is deprecated All drivers will have to be aware of the server version in the next major release.
    1x in PHPUnitExtension::executeBeforeFirstTest from DAMA\DoctrineTestBundle\PHPUnit
``` 

unfortunately there is no way to fix this deprecation :confused: Also see https://github.com/dmaicher/doctrine-test-bundle/issues/188#issuecomment-1002179428

javiereguiluz:
@dmaicher thanks a lot for your reply. However, I'm afraid I don't understand this well:

> unfortunately there is no way to fix this deprecation

You mean that this can't be fixed in DoctrineTestBundle and need to be fixed somewhere else (e.g. Doctrine DBAL) or you really mean that it cannot be fixed in any way. Thanks!

dmaicher:
There is no way currently to get rid of this deprecation because of DBAL. The `VersionAwarePlatformDriver` interface is deprecated but still needs to be implemented. There is no forward compatibility layer on DBAL so that it works without implementing the interface.

So I don't see any way of fixing this unless maybe there is a forward compatibility path on DBAL itself that would allow not implementing the interface.

javiereguiluz:
@dmaicher I see. Thanks for explaining this in detail. Maybe @stof knows people in the Doctrine organization or, specifically, the DBAL project that could take a look about this. Thanks!

javiereguiluz:
I've deleted the custom `sanitize_html` Twig filter to use the filter with the same name provided by HtmlSanitizer component (it was added here: https://github.com/symfony/symfony/pull/46335)

But, I'm seeing this error:

error-2

Any clue? Thanks!

stof:
@javiereguiluz there is a bug in the PR you linked: @nicolas-grekas added the new Twig extension in the twig bridge, but forgot to update TwigBundle to register it.

composer.json
```diff
@@ -9,10 +9,11 @@
         "symfony/polyfill-php72": "*",
         "symfony/polyfill-php73": "*",
         "symfony/polyfill-php74": "*",
-        "symfony/polyfill-php80": "*"
+        "symfony/polyfill-php80": "*",
+        "symfony/polyfill-php81": "*"
     },
     "require": {
-        "php": ">=8.0.2",
+        "php": ">=8.1",
```
94noni, line 16: do not forget the readme file's requirements :)

94noni (approved):

composer.json
```diff
@@ -9,10 +9,11 @@
         "symfony/polyfill-php72": "*",
         "symfony/polyfill-php73": "*",
         "symfony/polyfill-php74": "*",
-        "symfony/polyfill-php80": "*"
+        "symfony/polyfill-php80": "*",
```
94noni, line 12: both can be removed as now the php version is forced via the require

javiereguiluz: I'm not sure ... all these are in the `replace` section of `composer.json`. If I'm right, listing them here means that Composer won't download those packages even if some dependency requires them. So, the idea is that you add new polyfills when you upgrade your minimum required PHP version.

94noni: ho right! I did not pay attention where the location were at first :s my bad

PR #1343, opened 2022-07-03
Replaced deprecated `node-sass` with `sass`

### LINKED ISSUE #1342: Upgrade yarn packages to latest
Since this is the demo program that shows best practices, is there any reason not to update to the latest packages?  In particular, Webpack and stimulus.

```
@symfony/stimulus-bridge      2.1.0   -> 3.2.1                                                     
@symfony/webpack-encore       1.4.0 ->  2.1.0                                 
stimulus                      2.0.0  ->  3.0.1  
 ``` 

I'm happy to do a PR, but wanted to see if there was any particular reason to keep these older versions in the branch that works with 6.1.

skmedix:
Since Bootstrap 5, recommended is to use [`sass`](https://www.npmjs.com/package/sass) library, as [`node-sass` is deprecated](https://sass-lang.com/blog/libsass-is-deprecated), even if we are still using Bootstrap 3 in the demo, this change should be considered.

@javiereguiluz (Sorry for the ping 🙏) Should I also resolve #1342 here?


javiereguiluz:
As you said, we're using a legacy version of Bootstrap (there's a pending PR to update it ... but it stalled and I don't know if it's mergeable). Do you know if updating `node-sass` to `sass` is safe for such a legacy version? Thanks!

Yes, if you prefer, we can do all the updates in a single PR. Cheers!

skmedix:
The only concern I would see here is the deprecation of [Slash as Division](https://sass-lang.com/documentation/breaking-changes/slash-div), which is still used in Bootswatch, but I don't think this deprecation will be removed anytime soon. (We will probably upgrade the demo to a newer version of Bootstrap sooner.)

javiereguiluz:
Thank you @skmedix.

PR #1354, opened 2022-08-16
Update lint.yaml to fix security check

Coderberg:
Changed directory name from .symfony to .symfony5
See details here https://github.com/symfony-cli/symfony-cli/commit/5301ebfa8f5918cae7fa0bf63c86f84dbd70f599

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -89,7 +89,7 @@ jobs:
 
             - name: Check if any dependencies are compromised
               if: always() && steps.install.outcome == 'success'
-              run: /home/runner/.symfony/bin/symfony check:security
+              run: /home/runner/.symfony5/bin/symfony check:security
 
             - name: Run PHPStan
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yaml
```diff
@@ -89,7 +89,7 @@ jobs:
 
             - name: Check if any dependencies are compromised
               if: always() && steps.install.outcome == 'success'
-              run: /home/runner/.symfony/bin/symfony check:security
+              run: /home/runner/.symfony5/bin/symfony check:security
```
bocharsky-bw, line 92: Hm, but Symfony Demo is on Symfony 5 already 🤔

stof: this is not about the version of Symfony, but the installation path of the CLI tool

javiereguiluz:
Thanks Valery. I'm sorry it took me so long to merge this.

PR #1357, opened 2022-08-27
Remove Tag::setName

chr-hertel:
In this case we can move it to constructor


src/Entity/Tag.php
```diff
@@ -12,6 +12,7 @@
 namespace App\Entity;
 
 use Doctrine\ORM\Mapping as ORM;
+use JsonSerializable;
```
stof, line 15: in the Symfony coding standards, we don't add use statements for classes in the global namespace

stof (changes requested):

TheDevick:
Can we use a `readonly` proprierty in the Entity and keep the `setName` function?

> Why?

Just because sometimes it makes sense to call the constructor with no parameters - `new Tag();` and then call the `setName()` function, so making the code more readable in some cases

chr-hertel:
I would not agree on having public methods for "just in case..." and within this demo application it is not needed.
And I usually try to avoid those nullable properties if possible.

And no, you cannot combine readonly with setters. That's basically the feature :)

javiereguiluz:
Christopher, thanks for this nice contribution and sorry it took me so long to merge it.

Please note that while merging I tweaked the new `Tag` entity constructor a bit. See https://github.com/symfony/demo/commit/fbc76bbdf65c61fda8f5ae8478c282f127ead29a Your code is of course perfectly valid, but I think it could look too advanced/unconventional for some folks using this demo application. In the future we might change it again and use your version, but for now I'd prefer to use this more traditional constructor. Thanks!

chr-hertel:
@javiereguiluz Totally get your point! :) Thanks for coming back to this!

PR #1359, opened 2022-10-04
Use 'composer audit' instead of 'symfony check:security'

rosier:
Since Composer 2.4, there is a new command called audit, that outputs a list of reported security vulnerabilities for the list of packages versions currently installed.

This removes the need to download and install Symfony CLI


stof (approved):

javiereguiluz:
Thank you Jan.

PR #1362, opened 2022-10-28
Update to Symfony 6.2

javiereguiluz:
Comments:

* I've removed the SensioFrameworkExtraBundle because Symfony 6.2 now includes everything we need
* The `dev` dependency of DoctrineBundle is needed because we need the unreleased `2.7.1` to have the equivalent of the old ParamConverter
* The following indirect deprecations remain:

```
Remaining indirect deprecation notices (3)

1x: The "Symfony\Bridge\Doctrine\Logger\DbalLogger" class implements
"Doctrine\DBAL\Logging\SQLLogger" that is deprecated
Use {@see \Doctrine\DBAL\Logging\Middleware}
or implement {@see \Doctrine\DBAL\Driver\Middleware} instead.

1x: The "Monolog\Logger" class is considered final. It may change without
further notice as of its next major version. You should not extend it
from "Symfony\Bridge\Monolog\Logger".

1x: Since symfony/translation 6.2: "Symfony\Component\Translation\Extractor\PhpExtractor"
is deprecated, use "Symfony\Component\Translation\Extractor\PhpAstExtractor" instead.
```


stof:
For the first one, we need to debug why the DbalLogger still gets autoloaded while DoctrineBundle properly supports using middlewares when using DBAL 3.2+ and Symfony versions containing the middleware for the profiler integration.

for the `Symfony\Bridge\Monolog\Logger`, this should be reported to Symfony to find a way to stop extending that class (or to Monolog to revert that decision if we cannot find another way).

For the extractor, either a config setting needs to be added to switch to the new extractor, or something needs to be fixed in Symfony to avoid triggering the deprecation in all projects.

config/packages/doctrine.yaml
```diff
@@ -27,6 +27,10 @@ when@prod:
             result_cache_driver:
                 type: pool
                 pool: doctrine.result_cache_pool
+            controller_resolver:
+                enabled: true
+                auto_mapping: true
```
stof, line 32: I would actually disable auto_mapping here. For cases where the URL contains the id, it is not necessary. And for the case where the URL contains a slug, you are using an explicit mapping because of the different name. So I don't think auto_mapping is used in this project.
Btw, in my experience, this auto_mapping behavior can have surprising results (especially when using optional arguments), which is precisely why I advocated to make it configurable (in the DoctrineParamConverter of SFEB, there was no way to turn it off and so explicit `@ParamConverter` annotation was necessary to override it in case of buggy behavior introduced by it)

stof (commented):
As the DataTransformerInterface is now generic in Symfony 6.2+, you should add the `@template-implements DataTransformerInterface<Tag[], string>` on the TagArrayToStringTransformer to make phpstan happy about it (as it wants to know the type for that class)

config/packages/doctrine.yaml
javiereguiluz: Thanks. I've just removed it. I read the docs about this and I thought it was necessary.

My local app didn't work, I added this and it worked ... but now I remove it and the app still works. So I guess it's not needed.

seb-jean:
Instead of using `$this->getUser()`, we could use `#[CurrentUser]`:
https://symfony.com/blog/new-in-symfony-5-2-controller-argument-attributes

javiereguiluz:
I've just updated it until Symfony 6.2 RC1 and also updated all recipes. We still have these indirect deprecations:

```
Remaining indirect deprecation notices (3)

  1x: The "Symfony\Bridge\Doctrine\Logger\DbalLogger" class implements
      "Doctrine\DBAL\Logging\SQLLogger" that is deprecated
      Use {@see \Doctrine\DBAL\Logging\Middleware}
      or implement {@see \Doctrine\DBAL\Driver\Middleware} instead.

  1x: The "Monolog\Logger" class is considered final. It may change without
      further notice as of its next major version. You should not extend
      it from "Symfony\Bridge\Monolog\Logger".

  1x: Since symfony/translation 6.2: "Symfony\Component\Translation\Extractor\PhpExtractor"
      is deprecated, use "Symfony\Component\Translation\Extractor\PhpAstExtractor" instead.

```

src/Controller/BlogController.php
```diff
@@ -98,11 +100,10 @@ public function postShow(Post $post): Response
      */
     #[Route('/comment/{postSlug}/new', methods: ['POST'], name: 'comment_new')]
     #[IsGranted('IS_AUTHENTICATED')]
-    #[ParamConverter('post', options: ['mapping' => ['postSlug' => 'slug']])]
-    public function commentNew(Request $request, Post $post, EventDispatcherInterface $eventDispatcher, EntityManagerInterface $entityManager): Response
+    public function commentNew(#[CurrentUser] User $user, Request $request, #[MapEntity(mapping: ['postSlug' => 'slug'])] Post $post, EventDispatcherInterface $eventDispatcher, EntityManagerInterface $entityManager): Response
```
fabpot, line 103: When using attributes, what about having one argument per line?

javiereguiluz: Yes, good idea 👍  Done. Thanks!

seb-jean:
@javiereguiluz , It is good practice to put each controller argument on its own line?

javiereguiluz:
@seb-jean the traditional Symfony practice was to put all arguments on the same line, even if it's too long.

However, now that we're applying PHP attributes to some arguments, we think it's better in that case to put each argument on its own line, to not miss those PHP attributes.

seb-jean:
We no longer need to call `createView()`
https://symfony.com/blog/new-in-symfony-6-2-dx-improvements#simpler-form-rendering

stof:
For the remaining deprecations, here is the status:

1. work is in progress in DoctrineBundle to get rid of it (we probably need to mark the SQL logger service as deprecated, unless we mis-identify the root cause)
2. this is discussed in https://github.com/symfony/symfony/issues/47096 but it requires adding a new feature in FrameworkBundle to change the way the LoggerDataCollector is wired which might require waiting for Symfony 6.3 to solve the Monolog 3.2 deprecation (or doing weird things in a compiler pass in MonologBundle)
3. Does this deprecation go away if you install `nikic/php-parser` to enable the new AST-based extractor ? If yes, we might need to find a solution to avoid reporting that deprecation in a different way for projects that don't want to use the translation extractor.

.github/workflows/tests.yaml
```diff
@@ -10,7 +10,7 @@ env:
     fail-fast: true
     PHPUNIT_FLAGS: "-v"
     SYMFONY_PHPUNIT_DIR: "$HOME/symfony-bridge/.phpunit"
-    SYMFONY_REQUIRE: ">=6.0"
+    SYMFONY_REQUIRE: ">=6.2"
```
stof, line 13: why having this env variable instead of letting the CI rely on the setting defined in the composer.json ?

javiereguiluz:
@stof thanks for the update about the deprecations! After the latest updates, the deprecation related to translation is gone. I can only see these:

```
  1x: The "Symfony\Bridge\Doctrine\Logger\DbalLogger" class implements "Doctrine\DBAL\Logging\SQLLogger" that is deprecated Use {@see \Doctrine\DBAL\Logging\Middleware} or implement {@see \Doctrine\DBAL\Driver\Middleware} instead.

  1x: The "Monolog\Logger" class is considered final. It may change without further notice as of its next major version. You should not extend it from "Symfony\Bridge\Monolog\Logger".
```

.github/workflows/tests.yaml
javiereguiluz: I removed it. Thanks.

javiereguiluz:
@seb-jean I've merged this PR without doing other changes because it was taking too long already ... but I'd love to review a PR that updates the code to use some of the nice features introduced in Symfony 6.2, like the ones you mentioned here. Thanks!

lxregistry:
there are also other deprecation:
User Deprecated: Method "Countable::count()" might add "int" as a native return type declaration in the future. Do the same in implementation "Doctrine\Common\Collections\AbstractLazyCollection" now to avoid errors or add an explicit @return annotation to suppress this message.

Issue #1314, opened 2022-02-08
Use logout_path() helper + add CSRF protection to logout


lxregistry:
after these changes Use logout_path() helper + add CSRF protection to logout

change password is not working , Invalid CSRF token.

in UserController 
return $this->redirectToRoute('security_logout');

bocharsky-bw:
I suppose we need to leverage `LogoutUrlGenerator::getLogoutPath()` in that spot too, which should fix the problem. Could you try this and probably create a PR?

alexis78-sym:
@lxregistry Do you still have the problem?

PR #1364, opened 2022-10-29
Add CSRF protection to logout url in the user controller

rosier:
Fixes: #1314

Related: #1312


tests/Controller/UserControllerTest.php
```diff
@@ -91,7 +91,7 @@ public function testChangePassword(): void
         ]);
 
         $this->assertResponseRedirects(
-            '/en/logout',
+            null,
```
javiereguiluz, line 94: Is this change OK? The previous value looked better to me. Thanks.

javiereguiluz:
Thank you Jan.

PR #1369, opened 2022-11-13
Create devcontainer.json

shyim:
Adds an example how to setup it on [GitHub Codespaces](https://github.com/features/codespaces)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,20 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
+  "features": {
+    "ghcr.io/shyim/devcontainers-features/symfony-cli:0": {},
+    "ghcr.io/shyim/devcontainers-features/php:0": {
+      "version": "8.1"
+    },
+    "ghcr.io/devcontainers/features/node:1": {
+      "version": "lts"
+    }
+  },
+  "updateContentCommand": {
+    "composer install": ["composer", "install"],
+    "yarn": ["yarn"]
+  },
+  "postAttachCommand": {
+    "server": "symfony server:start",
+    "yarn": ["yarn", "watch"]
+  }
+}


GromNaN (approved):
This is fantastic!

Why we should merge this?
- This is the fastest way to run the demo
- This is the easiest way to setup a development environment
- We are on GitHub and this is a great feature of GitHub

Also, we can try to promote Symfony by having a Symfony template on https://github.com/codespaces/templates

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,20 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
+  "features": {
+    "ghcr.io/shyim/devcontainers-features/symfony-cli:0": {},
```
alexislefebvre, line 4: Can the Symfony organisation host these images, so that it doesn't depend on an external contributor? (no offence to shyim, but this account may be closed, hacked, etc. and people would end up with no image or a counterfeit image).

shyim: I don't know if Symfony wants to host some stuff. I am open :) 
You should also be pinning to a specific sha if you prefer

alexislefebvre: @javiereguiluz Would it be possible to add a project to the Symfony organisation in order to build some Docker images? Where (what repo) should we request this?

stof: For the Symfony CLI, this should probably be discussed on https://github.com/symfony-cli/symfony-cli/issues (and then, the repo needed for that would also live in that organization, not in the Symfony one)

alexislefebvre: Actually it looks like I was confused by the `ghcr.io` name, `ghcr.io/shyim/devcontainers-features/symfony-cli:0` is not a Docker image as I expected but it's a text recipe. But it would be nice to host it on Symfony's side though, I'll open an issue.

PR #1373, opened 2022-12-02
CI: Remove "Require symfony/flex" step from tests

rosier:
I think this step can be remove since "symfony/flex" is already part of the project dependencies.

https://github.com/symfony/demo/blob/v2.2.0/composer.json#L28


rosier:
Looking at the following line I would say that the flex plugin is still active

https://github.com/symfony/demo/actions/runs/3603191289/jobs/6071084273#step:7:133

Is this maybe something that changed with a newer version of composer?

stof:
Ah, I forgot that recent Flex versions are automatically restarting the composer update when they are installed as part of an update that did not have the filter applied yet.

javiereguiluz:
Thanks

Issue #1334, opened 2022-05-05
Fix issues reported by PHPStan


javiereguiluz:
When we introduced PHPStan, we added lots of errors to a "baseline file" to ignore them instead of fixing them. So, let's fix them.

In my opinion there are two kinds of issues:

(1) Issues that are PHP errors or improvements. Such as https://github.com/symfony/demo/pull/1331/files We need to fix those.

(2) Issues that are solved by adding metadata for PHPStan. I'm not sure 100% about fixing these. This is a Symfony Demo app, so only Symfony is important. If these issues require lots of changes and future maintenance, I wouldn't fix them.

What do you think?

javiereguiluz:
About (2) @stof added the following in https://github.com/symfony/demo/pull/1332#discussion_r866011366

-----

Well, it also helps IDEs that support generic classes (PHPStorm has partial support for them, improving it in each release).
Generics are actually a de-facto standard, being supported with the same syntax by phpstan, psalm, PHPStorm (and maybe VSCode but I haven't checked)

And MakerBundle generates that `@extends` tag now. So users generating a User entity today will have it.

kniziol:
No errors found. PHPStan said `[OK] No errors`. Details below:
```shell
$ ./vendor/bin/phpstan --memory-limit=-1 -vvv
Note: Using configuration file phpstan.neon.dist.
 44/44 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100% < 1 sec/< 1 sec 40.0 MiB

Result cache is saved.

                                                                                                                        
 [OK] No errors                                                                                                         
                                                                                                                        

Used memory: 40 MB
```

PR #1374, opened 2022-12-03
Update PHPDoc and remove unnecessary PHPDoc applied to fix PHP-CS-Fixer errors #1334

kniziol:


javiereguiluz:
A very nice continuation of #1370. Thanks @kniziol and thanks @stof for your excellent review of the previous pull request.

PR #1376, opened 2022-12-03
Remove composer/package-versions-deprecated from allow-plugins

rosier:
The plugin composer/package-versions-deprecated is no longer in use.

Followup of #1310


javiereguiluz:
Thank you

Issue #1347, opened 2022-07-17
Global `symfony/flex` is not needed


asispts:
`symfony/flex` is already added to `composer.json`
https://github.com/symfony/demo/blob/0ff22b52b6ec3e0217d33f0a12a66020ec277ed8/composer.json#L29

So, I think this action is not required.
https://github.com/symfony/demo/blob/0ff22b52b6ec3e0217d33f0a12a66020ec277ed8/.github/workflows/tests.yaml#L58-L59

The action will also produce an unnecessary error
![000](https://user-images.githubusercontent.com/79239132/179424810-d5f0110d-9905-4542-9540-aa3d18af5bc9.png)

Btw, why run `composer update` instead of `composer install`?
https://github.com/symfony/demo/blob/0ff22b52b6ec3e0217d33f0a12a66020ec277ed8/.github/workflows/tests.yaml#L61

javiereguiluz:
The global `symfony/flex` dependency was removed in #1373.

About using `composer update` instead of `composer install` ... since we only test one specific Symfony version, I guess we could use `install`. If anyone wants to send a Pull Request for this change, please do it. Thanks.

PR #1382, opened 2022-12-19
Use 'composer install' instead of 'composer update'

rosier:
Fixes _https://github.com/symfony/demo/issues/1347#issuecomment-1348751078_


javiereguiluz:
Thank you Jan.

PR #1390, opened 2023-01-09
Replace AJAX search with Live component

ker0x:
This pull request proposes to replace the AJAX search with a live component example


94noni:
its a nice idea, following stimulus intro 👍🏻 
but it could be wise to also add some code doc to link to documentation etc, as just browsing the demo code without explanation can be disturbing imho

templates/blog/search.html.twig
```diff
@@ -2,24 +2,14 @@
...
-                   class="form-control search-field"
-                   placeholder="{{ 'post.search_for'|trans }}"
-                   autocomplete="off"
-                   value="{{ query }}"
-                   autofocus
-                   data-no-results-message="{{ 'post.search_no_results'|trans }}"
-            >
-        </div>
-    </form>
+    {{ component('blog_search', {
```
94noni, line 10:
```suggestion
    {# Lear more on https://symfony.com/bundles/ux-live-component/current/index.html #}
    {{ component('blog_search', {
```

templates/components/blog_search.html.twig
```diff
@@ -0,0 +1,20 @@
+<div {{ attributes }}>
```
94noni, line 1: same as above, adding link to `{# Lear more on https://symfony.com/bundles/ux-live-component/current/index.html #}`

src/Twig/Components/BlogSearchComponent.php
```diff
@@ -0,0 +1,45 @@
...
+namespace App\Twig\Components;
+
+use App\Entity\Post;
+use App\Repository\PostRepository;
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\LiveComponent\DefaultActionTrait;
+
+/**
+ * Live component to display instant search for Posts.
```
94noni, line 21: the file https://github.com/symfony/demo/blob/main/src/Command/AddUserCommand.php describe some usage etc
could be wise to do same here, wdyt?

for ex to https://ux.symfony.com/live-component or/and https://symfony.com/bundles/ux-live-component/current/index.html

ker0x: Good idea indeed. I added some comments in my code!

templates/blog/search.html.twig
```diff
@@ -2,24 +2,18 @@
 
 {% block javascripts %}
     {{ parent() }}
-    {{ encore_entry_script_tags('search') }}
 {% endblock %}
```
rosier, line 5: This javascripts block can be remove now

ker0x: Indeed, nice catch thanks!

94noni (approved):

javiereguiluz:
Romain, thanks a lot for contributing this nice feature. I feared that introducing an UX example would complicate things too much ... but making this change was simply perfect because the result is very nice and it allows to compare traditional JS features vs modern UX-based JS-like features.

Thanks to reviewers too 🙏

COil:
Excellent, that's the kind of "real use case" that really shows the power of UX. 👏

PR #1399, opened 2023-02-12
doc: note about GitHub clone case

COil:


.gitignore
```diff
@@ -2,6 +2,7 @@
 /public/build/images/glyphicons-*
 
 .php-version
+.idea
```
OskarStark, line 5: This should be part of your global .gitignore file

OskarStark (approved):

README.md
```diff
@@ -47,6 +47,13 @@ $ cd my_project/
 $ symfony serve
 ```
 
+If you have cloned the [GitHub repository](https://github.com/symfony/demo), you
+must install the Composer dependencies.
+
+```bash
+$ composer install
```
bocharsky-bw, line 54: What about using Symfony CLI in the command? It will use the correct PHP version
```suggestion
$ symfony composer install
```

.gitignore
```diff
@@ -17,6 +17,7 @@
 .phpunit.result.cache
 /phpunit.xml
 ###< symfony/phpunit-bridge ###
+
```
stof, line 20: this should be reverted

javiereguiluz: Done while merging. Thanks.

javiereguiluz:
Thank you Loïc!

All of you had nice ideas for this ... so I merged them all while merging this: https://github.com/symfony/demo/commit/de31ae90564fb1594509acfb1e626765ab138023

README.md
```diff
@@ -47,6 +47,13 @@ $ cd my_project/
 $ symfony serve
 ```
 
+If you have cloned the [GitHub repository](https://github.com/symfony/demo), you
```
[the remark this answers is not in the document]
COil: Yes.

README.md
COil: Yes, seems better in this case.

javiereguiluz: I'd prefer to not mix `symfony` binary with Composer ... because people who will use the Composer installation method is people who don't want or can't download the Symfony CLI, so they won't have access to the `symfony` binary. Thanks.

PR #1401, opened 2023-02-13
Update the README contents

javiereguiluz:
After merging #1399, I thought about tweaking `README` contents to make it more clear the different alternative ways of installing and using this project.


bocharsky-bw (approved):
It's cleaner now, I like it 👍

PR #1403, opened 2023-02-14
php8: fixed named arguments order

COil:
This is to fix:

Capture d’écran 2023-02-14 à 22 25 44

### DIFF
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -52,8 +52,8 @@ class BlogController extends AbstractController
      *     could move this annotation to any other controller while maintaining
      *     the route name and therefore, without breaking any existing link.
      */
-    #[Route('/', methods: ['GET'], name: 'admin_index')]
-    #[Route('/', methods: ['GET'], name: 'admin_post_index')]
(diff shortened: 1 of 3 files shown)


derrabus:
There's nothing to fix here. That PhpStorm inspection is pointless and disabling it gives me peace of mind in every codebase I've worked on so far. 😌

OskarStark (approved):

COil:
> There's nothing to fix here. That PhpStorm inspection is pointless and disabling it gives me peace of mind in every codebase I've worked on so far. 😌

Yes, but it's mainly for consistency to avoid having different orders everywhere. It should/could be a php-cs-fixer rule.

OskarStark:
You are right from a technical POV @derrabus, but I would argue that most of the devs write it in the from @COil proposed order anyway. That said I am +1 to merge this change.

This repo ist mostly for newcomers 🤷🏼‍♂️

jderusse (approved):
For enforcing consistency

javiereguiluz:
I agree with all of you ... but there seems to be more people in favor of merging this, so let's do it. Thanks Loïc!

PR #1404, opened 2023-02-16
cleanup: use roles constants

COil:
Before we had to hard-code roles strings: 

```
use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;

class PostController extends Controller
{
    /**
     * @IsGranted("ROLE_ADMIN")
     *
     * or use @Security for more flexibility:
     *
     * @Security("is_granted('ROLE_ADMIN') and is_granted('ROLE_FRIENDLY_USER')")
     */
    public function index()
    {
        // ...
    }
}
```

But with attributes, can use constants. I find this cleaner. I have already used this on several projects and I didn't find drawbacks.

### DIFF
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -32,6 +32,12 @@ use Symfony\Component\Validator\Constraints as Assert;
 #[ORM\Table(name: 'symfony_demo_user')]
 class User implements UserInterface, PasswordAuthenticatedUserInterface
 {
+    // We can use constants for roles to find usages all over the application rather
+    // than doing a full-text search on the "ROLE_" string.
+    // It also prevents from making typo errors.
+    final public const ROLE_USER = 'ROLE_USER';
+    final public const ROLE_ADMIN = 'ROLE_ADMIN';
+
     #[ORM\Id]
     #[ORM\GeneratedValue]
     #[ORM\Column(type: Types::INTEGER)]
@@ -118,7 +124,7 @@ class User implements UserInterface, PasswordAuthenticatedUserInterface
 
         // guarantees that a user always has at least one role for security
         if (empty($roles)) {
-            $roles[] = 'ROLE_USER';
+            $roles[] = self::ROLE_USER;
         }
 
         return array_unique($roles);
--- a/composer.lock
+++ b/composer.lock
@@ -8153,16 +8153,16 @@
         },
         {
             "name": "phpstan/phpstan-doctrine",
-            "version": "1.3.35",
+            "version": "1.3.36",
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpstan/phpstan-doctrine.git",
-                "reference": "3dbb999bd56cc21d6733bab60898ebdf388245b5"
+                "reference": "8c22f2499444897941f9162c25f8538217c16f15"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpstan/phpstan-doctrine/zipball/3dbb999bd56cc21d6733bab60898ebdf388245b5",
-                "reference": "3dbb999bd56cc21d6733bab60898ebdf388245b5",
+                "url": "https://api.github.com/repos/phpstan/phpstan-doctrine/zipball/8c22f2499444897941f9162c25f8538217c16f15",
+                "reference": "8c22f2499444897941f9162c25f8538217c16f15",
                 "shasum": ""
             },
             "require": {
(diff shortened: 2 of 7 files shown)


src/Entity/User.php
```diff
@@ -32,6 +32,12 @@
 #[ORM\Table(name: 'symfony_demo_user')]
 class User implements UserInterface, PasswordAuthenticatedUserInterface
 {
+    // we can use constants for roles so we get all usages all over the application
+    // more easyly than doing a search on the "ROLE_" string. It also prevents from
```
xabbuh, line 36:
```suggestion
    // more easily than doing a search on the "ROLE_" string. It also prevents from
```

config/packages/twig.yaml
```diff
@@ -4,6 +4,12 @@ twig:
         - 'form/layout.html.twig'
         - 'form/fields.html.twig'
 
+    # These variables are now available in all Twig templates
+    globals:
+        # We can use PHP constants from the application thanks to the !php/const sintaxis.
```
xabbuh, line 9:
```suggestion
        # We can use PHP constants from the application thanks to the !php/const syntax.
```

src/Entity/User.php
OskarStark: We should avoid using "easily"

javiereguiluz:
Could we use an Enum instead of constants? Would Enums work everywhere? Is it worth it or is it overengineering?

stof:
No we cannot. Roles are strings in Symfony, not an enum object.

src/Entity/User.php
COil: I have remove "easily".

config/packages/twig.yaml
COil: Done.

wouterj:
You can use e.g. `RoleEnum::USER->value` (since PHP 8.1) in attributes and php code, bit IIRC not in Yaml.

COil:
> You can use e.g. `RoleEnum::USER->value` (since PHP 8.1) in attributes and php code, bit IIRC not in Yaml.

With PHP 8.2 it will be OK, but we can't yet with 8.1:

Capture d’écran 2023-02-18 à 11 03 57

COil:
Hello, is it OK for this PR? Or should we close it?

javiereguiluz:
My opinion:

* 🟢 I like a lot using these constants in PHP files
* 🟡 I'm not entirely sure about injecting these constants in all Twig templates as global constants
* 🔴 I don't like how complicated this looks in the YAML file, especially in the `security.yaml` file

What do others think about this? Thanks.

derrabus:
I fully agree with @javiereguiluz' assessment.

> 🟢 I like a lot using these constants in PHP files

Me too, I wouldn't make those constants members of the `User` class though. In my projects, I usually have a `final class Roles` that holds those constants.

> 🟡 I'm not entirely sure about injecting these constants in all Twig templates as global constants

Me neither and I wouldn't do it tbh. We don't gain much here.

> 🔴 I don't like how complicated this looks in the YAML file, especially in the `security.yaml` file

Same. I would revert this as well.

stof:
Sure you can workaround the bad architecture by using `RoleEnum::USER->value` everywhere. But I would not recommend doing that in the demo as this would give the impression that using a enum for roles is a good idea, while this is an abuse of enums based on the fact that you need role strings anyway.

javiereguiluz:
@stof but in this case we're not talking about enums. It's using PHP constants in YAML that looks too complicated to us. E.g. this change:

```diff
     role_hierarchy:
-        ROLE_ADMIN: ROLE_USER
+        !php/const App\Entity\User::ROLE_ADMIN: !php/const App\Entity\User::ROLE_USER
```

But in PHP files these constants look very nice.

COil:
> Me too, I wouldn't make those constants members of the User class though. In my projects, I usually have a final class Roles that holds those constants.

But that looks weird to put this in a separated class. In this case we should use an enum (but we don't want too here). Having these constants in the User class is the more pragmatic way imho.

> 🟡 I'm not entirely sure about injecting these constants in all Twig templates as global constants
> Me neither and I wouldn't do it tbh. We don't gain much here.

Actually, if we don't want to use a constant here, we could use a twig extension doing the job and making use of the PHP constant. But that's not as simple as using `is_granted` indeed.

> 🔴 I don't like how complicated this looks in the YAML file, especially in the security.yaml file

I agree it makes the YAML unreadable. I didn't even knew it was possible to do this, so let's revert this.

derrabus:
> But that looks weird to put this in a separated class.

🤨 🤷🏻‍♂️

src/Entity/User.php
```diff
@@ -32,6 +32,12 @@
 #[ORM\Table(name: 'symfony_demo_user')]
 class User implements UserInterface, PasswordAuthenticatedUserInterface
 {
+    // We can use constants for roles to find usages all over the application rather
+    // than doing a full-text search on the "ROLE_" string.
+    // It also prevents from making typo errors.
+    public const ROLE_USER = 'ROLE_USER';
+    public const ROLE_ADMIN = 'ROLE_ADMIN';
```
derrabus, line 39:
```suggestion
    final public const ROLE_USER = 'ROLE_USER';
    final public const ROLE_ADMIN = 'ROLE_ADMIN';
```

COil: done

javiereguiluz (approved):
I like this proposal a lot. Thanks Loïc!

javiereguiluz:
It's merged now. Thanks!

Issue #1272, opened 2021-11-23
The style are not showing in browser


alexbotnaru:
I installed the application using the command : symfony new --demo my_project. It seems to work, but the styles and maybe the scripts are not working in browser, so the app looks like this: 
![image](https://user-images.githubusercontent.com/71169169/143044204-e7ac275e-d11c-450f-8306-81d77eaa9f98.png)
I tried to run some commands like : composer install, composer require symfony/webpack-encore-bundle, npm install, npm run build but the problem persists.
Any help?

stof:
Are there any errors reported in your browser console ?

bocharsky-bw:
Hm, I can't reproduce this... my steps were:

1. symfony new --demo my_project/
2. cd my_project/
3. symfony serve
4. Open https://127.0.0.1:8000/

Styles are OK to me after this. Do you use the latest version of Symfony CLI? You can check by executing `symfony self:update`. If yes, what browser and OS do you use? What browser version? It may help I think

alexbotnaru:
yeah, so I opened the project in Microsoft Edge and everythink works properly. I dont know why in Google Chrome i get those errors, as I have the last version of it (96.0.4664.45 ). Anyway, thanks a lot!

seb-jean:
@alexbotnaru, Is it resolved?

valentin705:
Hello, I just installed the Symfony demo and I have the same problem no matter what I open it with: Google Chrome, Microsoft Edge, or even Mozilla. I have the same error message as you. Do you have a solution? Thank you.

acsig:
This is still an issue. I just installed it now for the third time in a clean environment on a virtual machine. 

- downloaded the newest version today
- Win 10 prof
- xampp (8.2.0)
- composer installed (2.5.5)
- demo installed with "`composer create-project symfony/symfony-demo`"
- git installed by composer during the process
- result in edge AND chrome: css missing
- "`symfony serve`" doesn't work either
- used xampps apache server

It's hard to tell my students, that they need to install a node.js Webserver(!!!) and paket Manager npm/yarn to fix this issue. :-/ Any idea how to solve it?

javiereguiluz:
Try to run only these commands:

```
composer create-project symfony/symfony-demo
cd symfony-demo
symfony serve
```

Then browse any page, and tell us the error logs that you see in the terminal that is running `symfony serve`. Thanks.

acsig:
Looks like I found the problem: 

[...]

Issue #1263, opened 2021-09-11
CSS doesn't work


hawkesburygeek:
I have run "yarn" and then "symfony serve". The demo site works but the style seems not to work.
Is there anyone who can help me?
Jesus Watts

bocharsky-bw:
Try to run these commands:
```
$ bin/console assets:install
$ yarn install
$ yarn dev
```

If still not work, I'd recommend you to remove the repo and follow the instruction from readme: https://github.com/symfony/demo#installation

Guikingone:
Hi 👋🏻 

I can confirm the issue of @hawkesburygeek, in my case, I'm facing the issue when using Docker, if I decide to use the Symfony CLI, no issues but with Docker, the CSS is not used. 

Could be the Nginx configuration but the files are not under a 404 error 🤔

bocharsky-bw:
Hm, but this project does not provide any Docker configuration out of the box IIRC. Did you write that Docker setup yourself? If so, then in that case you're responsible for configuring it properly and it has nothing with this project to be reported

Guikingone:
Yes, sorry for the context that I gave, when I first tried the `symfony/demo`, it was with the Symfony CLI then I've faced the same issue as the one described by @hawkesburygeek, same with Docker (in the end, the Docker issue was a configuration one). 

The Docker configuration is fixed but with the Symfony CLI, I needed to update it before it works, strange situation here, maybe just a server configuration that trigger this kind of behaviour?

stof:
> I needed to update it before it works

updating what ? The symfony CLI ? Well, if you are not on the latest version of a tool, updating it is a prerequisite before reporting issues (as they might be fixed already).

Guikingone:
> Well, if you are not on the latest version of a tool, updating it is a prerequisite before reporting issues (as they might be fixed already).

As it wasn't a major or minor version that was "skipped", that what's not the first thing that came in my mind regarding a server issue 🙂

seb-jean:
@hawkesburygee, is it resolved?

marksmanpilotcurler:
Hi all --
    I also ran into this issue.   I had used git clone to pull down the symfony/demo source from github.   I now believe git was the issue.   Apologies -- the below is from memory, I don't have all the of the filenames and locations, but I do remember what happened.

[...]

PR #1407, opened 2023-03-01
Add yarn instructions

Irtoson:
Proposition to update README in order to add yarn instructions (needed if you want some CSS on your page ;) )

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ $ cd my_project/
 $ php -S localhost:8000 -t public/
 ```
 
-Don't forget to compile CSS / JS components :
+If you encounter CSS issues at first glance, try to re-compile CSS / JS components :
 
 ```bash
 $ yarn install
--- a/README.md
+++ b/README.md
@@ -70,6 +70,13 @@ $ cd my_project/
 $ php -S localhost:8000 -t public/
 ```
 
+Don't forget to compile CSS / JS components :
+
+```bash
+$ yarn install
+$ yarn build
+```
+
 Tests
 -----
 


bocharsky-bw:
Symfony Demo has committed public/build/ dir which mean you don't have to run those commands locally, it will just work out of the box for you

Irtoson:
I agree, it should work, but unfortunately, did not work for me. I encountered some errors using Chrome 110.0.5481.104 on Windows 11 :
image
Issues https://github.com/symfony/demo/issues/1263 and https://github.com/symfony/demo/issues/1272 mention it, I thought we could add this quick "fix" to the instructions ?

COil:
> Symfony Demo has committed public/build/ dir which mean you don't have to run those commands locally, it will just work out of the box for you

Yes, but the project car be forked and used to create an actual blog. So, knowing how to regenerate the assets is useful (doc is always useful).

wouterj:
> I agree, it should work, but unfortunately, did not work for me. I encountered some errors using Chrome 110.0.5481.104 on Windows 11 :

Content-Security-Policy errors aren't fixed by regenerating the assets. Also, the Symfony demo application doesn't come with a Content-Security-Policy header so you must have configured something (on your local web server or in the application) that causes that error.

> Yes, but the project car be forked and used to create an actual blog. So, knowing how to regenerate the assets is useful (doc is always useful).

I would say that this is not the purpose of the README of this project. Adding more details might only confuse newcomers for what is necessary to run this project. The README already refers to the official docs and the Fast Track book, both of which document using yarn/npm when working with CSS/JS.

javiereguiluz:
@Irtoson thanks for this proposal. However, we're going to close this as "won't fix" for the reasons explained in the above comments.

It's important that this application works fully without having to install or configure anything. Adding those instructions in the READM could be confusing about the need to run those commands.

In addition to being used as a self-learning tool, this app is used in workshops, to benchmark Symfony, by the core PHP team to test new features, etc. That's why we need to be careful about adding things like this. I hope you understand it.

PR #1409, opened 2023-03-18
cleanup: use final classes

COil:
Hello, these classes could be `final` right? (except the entities)

(just done one example, will do the other if it is OK)

### DIFF
--- a/src/Command/AddUserCommand.php
+++ b/src/Command/AddUserCommand.php
@@ -51,7 +51,7 @@ use function Symfony\Component\String\u;
     name: 'app:add-user',
     description: 'Creates users and stores them in the database'
 )]
-class AddUserCommand extends Command
+final class AddUserCommand extends Command
 {
     private SymfonyStyle $io;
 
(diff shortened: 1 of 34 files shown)


derrabus:
In my apps, I usually declare all classes `final` unless I have a good reason not to. I wouldn't mind if the demo app followed this rule as well. 🙂

So 👍🏻  from my side.

COil:
The only problem I see is that the maker bundle doesn't put the final keyword on controllers.

OskarStark (approved):
> The only problem I see is that the maker bundle doesn't put the final keyword on controllers.

@weaverryan should we use final for controllers in maker bundle?

GromNaN:
There are [good reasons](https://matthiasnoback.nl/2018/09/final-classes-by-default-why/) to use final by default. Can we explain with a comment somewhere is this project?

javiereguiluz:
This practice is very common in Symfony apps, so let's merge this. Thanks Loïc!

jrushlow:
> > The only problem I see is that the maker bundle doesn't put the final keyword on controllers.
> 
> @weaverryan should we use final for controllers in maker bundle?

@OskarStark https://github.com/symfony/maker-bundle/pull/1539

PR #1412, opened 2023-04-05
Rebase and flatpickr

atierant:
Rebase @bechir work to comply with 6.2 branch

Resolve #1183 

Fix flatpickr as mandated by @bechir in https://github.com/symfony/demo/pull/1183#issuecomment-773693644


atierant:
About Intl js date management, moment is not recommanded anymore, but they advice to use their new Intl Date wrapper lib [Luxon](https://moment.github.io/luxon/) or [Temporal](https://momentjs.com/docs/#/-project-status/future/)

javiereguiluz:
Thanks for this contribution. Let's see if this time we can finish and merge this 🙏 

I tested it locally and found some issues.

**(1)** The datepicker doesn't work for me:

Captura de pantalla 2023-04-06 a las 11 25 20

And the error details:

Captura de pantalla 2023-04-06 a las 11 25 56

**(2)** Some spacing and font sizes look too different. The new design doesn't have to be identical to before, but it should be closer.

See before/after comparisons:

before-after-1

In the above screenshot, let's tweak the top bar and the form fields to look like before, but let's keep the new bi buttons because they look better than before.

before-after-2

atierant:
The language spotter in flatpicker.js is now fixed, just an oversight. At first I added the language as a data-attribute of the body, but I finally said to myself that it was already present in the html tag, so I had corrected too quickly...

You can now test the feature, I'll have a look for design adaptations.

atierant:
These differences come from the change to Bootstrap 4 and conforms to the standard Bootstrap 4 interface standards.

Units are basically changed to rem rather than pixels.
They changed the size of the inputs, the functioning of the navbar...
They used a scaffolding script (_scaffolding.scss) which differs from the new reboot system (_reboot.scss) in many points.

I'm not sure I've mastered these concepts enough to act properly on the interface, but if someone reading this in the assembly is an UX/UI master, show up with pleasure!

atierant:
It seems better ! Can you give a new try please ?

image

image

image

image

javiereguiluz:
This looks fantastic now! Thanks a lot Arnaud.

However, there's two minor issues related to the datepicker. Please, tell me if I'm doing something wrong:

(1) There's a design issue with the form field icon:

datepicker-icon

(2) The datepicker works for me in all languages:

|     |     |
| --- | --- |
| datepicker-es | datepicker-fr |

But in English it doesn't work 🤔 

datepicker-en

atierant:
@javiereguiluz Fixed :)

javiereguiluz (approved):
I tested this locally and everything worked as expected.

Thanks @atierant for this brilliant contribution!

bechir:
Good job @atierant

atierant:
> Good job @atierant 

Good job to you @bechir , the most is yours ! Thanks a lot for this !

javiereguiluz:
This is now merged! Thanks Arnaud.

PR #1415, opened 2023-05-02
Update the project to Symfony 6.3

javiereguiluz:
This PR will be WIP until 6.3 is released as stable.

### DIFF
--- a/assets/controllers.json
+++ b/assets/controllers.json
@@ -5,7 +5,7 @@
                 "enabled": true,
                 "fetch": "eager",
                 "autoimport": {
-                    "@symfony/ux-live-component/styles/live.css": true
+                    "@symfony/ux-live-component/dist/live.min.css": true
                 }
             }
         }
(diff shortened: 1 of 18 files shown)


javiereguiluz:
I've updated to Symfony 6.3 beta 2 and I still see this direct deprecation:

deprecation

The problem is that a Symfony internal `Logger` class extends a Monolog class marked as final:

logger-class

I have two questions:

* Why is it marked as "direct deprecation" instead of "indirect deprecation"?
* Are we aware of this in the Symfony code repository so we can fix it before the final release?

wouterj:
Direct vs indirect may be a bug (the logic around this is a bit fragile).

The deprecation itself is known and unrelated to a Symfony upgrade, but rather a Monolog update: https://github.com/symfony/symfony/issues/47096

stof:
Direct vs indirect deprecation is fragile for deprecations triggered by the DebugClassLoader right now

PR #1417, opened 2023-05-03
Update the design of the language selector

javiereguiluz:
The current header design looks a bit crowded to me, with the links so close to each other:

This PR changes this by making the links a bit smaller and adding more space between them:

-----

The other issue is the language selector. We're very lucky to have many translations contributed by the community. But that big number is what makes the current selector a bit hard to use (the list is very long). It also has other design issues:

This PR proposes a completely different design. When you click on the link to select the language, you no longer see a dropdown menu. Instead, you see a full page modal window listing all available languages.

For example, if you are browsing the application in Ukrainian and click on the language selector, you'll see this:

The new language selector was inspired by the one used in https://europa.eu pages. In addition to displaying all languages more clearly, we now respect the features of each language, like the RTL direction of Arabic.


templates/base.html.twig
```diff
@@ -80,26 +80,48 @@
...
+                                {# display this title always in English to make the language selector easier to use #}
+                                Choose your language
+                            </h5>
+                            <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                                <span aria-hidden="true">&times;</span>
+                            </button>
+                        </div>
+
+                        <div class="modal-body">
+                            <ul class="locales" role="menu" aria-labelledby="locales">
```
stof, line 109: `role="menu"` (and the associated `role="menuitem"` and `aria-checked` on the list items) should be removed as you don't implement the interaction patterns expected by an ARIA menu. Setting broken ARIA roles is actually making accessibility worse.

templates/base.html.twig
```diff
@@ -80,26 +80,48 @@
...
+                                {% for locale in locales() %}
+                                    {% set is_active = app.request.locale == locale.code %}
+                                    {% set is_rtl = locale.code in ['ar', 'fa', 'he'] %}
+                                    <li aria-checked="{{ is_active ? 'true' : 'false' }}" class="{{ is_active ? 'active' }} {{ is_rtl ? 'rtl' }}" role="menuitem" translate="no">
+                                        <a class="stretched-link" href="{{ path(app.request.get('_route', 'blog_index'), app.request.get('_route_params', [])|merge({_locale: locale.code})) }}">
+                                            <small>{{ locale.code[0:2] }}</small>
```
stof, line 115: should we really truncate it ? This would display `PT` for the `pt_BR`, while the Brazilian variant of Portuguese is totally different. To me, we should either show the non-truncated locale code or nothing at all.

Note that the europa.eu language selector does not attempt to show this.

templates/base.html.twig
```diff
@@ -80,26 +80,48 @@
...
                     </div>
                 </nav>
             </header>
+
+            <div id="locale-selector-modal" class="modal" tabindex="-1" aria-hidden="true">
+                <div class="modal-lg modal-dialog modal-dialog-centered">
+                    <div class="modal-content">
+                        <div class="modal-header">
+                            <h5 class="modal-title">
+                                {# display this title always in English to make the language selector easier to use #}
```
stof, line 100: you should put `lang="en"` on that element then, as it is not in the language defined in the `<html>` tag

templates/base.html.twig
```diff
@@ -80,26 +80,48 @@
...
+
+                        <div class="modal-body">
+                            <ul class="locales" role="menu" aria-labelledby="locales">
+                                {% for locale in locales() %}
+                                    {% set is_active = app.request.locale == locale.code %}
+                                    {% set is_rtl = locale.code in ['ar', 'fa', 'he'] %}
+                                    <li aria-checked="{{ is_active ? 'true' : 'false' }}" class="{{ is_active ? 'active' }} {{ is_rtl ? 'rtl' }}" role="menuitem" translate="no">
+                                        <a class="stretched-link" href="{{ path(app.request.get('_route', 'blog_index'), app.request.get('_route_params', [])|merge({_locale: locale.code})) }}">
+                                            <small>{{ locale.code[0:2] }}</small>
+                                            {{ locale.name|capitalize }}
```
stof, line 116: As this displays each locale name in its own language, we might need to add the appropriate `lang` attribute.

Looking at the europa.eu selector, they do have this `lang` attribute, and then also use `hreflang` on the link going to a different language.

javiereguiluz:
@stof thanks a lot for your great review 😍  I've made all the changes that you suggested. Thanks!

PR #1420, opened 2023-05-05
Update the design of the source code modal

javiereguiluz:
The modal window that displays the source code of controller/template is nice. But the current design is troublesome. It displays the controller code on top of the template code. It doesn't look good many times because controller/templates are too long:

source-code-modal-before

This PR proposes to use tabs to display controller and template side-by-side:

source-code-modal-after


assets/app.js
```diff
@@ -4,6 +4,8 @@ import './styles/app.scss';
 import 'bootstrap/js/dist/alert';
 import 'bootstrap/js/dist/collapse';
 import 'bootstrap/js/dist/dropdown';
+import 'bootstrap/js/dist/tab';
+import 'bootstrap/js/dist/util';
```
stof, line 8: there is no reason to import that yourselves IMO. Other Bootstrap JS files are importing the util file if they need it, and you don't use it directly by yourselves.

javiereguiluz: You are right! I've removed the `util.js` import. Thanks.

PR #1421, opened 2023-05-06
Bump deps

talk2karthik:
@aetxebeste


javiereguiluz:
Let's close this PR because it looks like it contains commits from other people. Also, we're updating this project to Symfony 6.3 (see #1415) so we're going to update all dependencies very soon.

PR #1426, opened 2023-07-24
Switch partial templates to anonymous Twig components

ker0x:
This PR proposed to switch to the HTML syntax for components. It also introduce a new `PostComponent` component


javiereguiluz:
Thanks for this proposal. Before continuing working on this PR, let's ask JS/UX experts like @weaverryan, @kbond and @stof  about this.

My non-expert opinion is divided. I think it's very obvious that making a UX component for search is correct ... but making an "empty" UX component just to display the post contents looks a bit "unnatural" for a Symfony app. It could confuse users about what's the traditional/common way of working in Symfony apps to display entity contents.

Let's wait for more opinions. Thanks!

kbond:
FYI, once https://github.com/symfony/ux/pull/802 is available, the empty component class will no longer be required.

ker0x:
@javiereguiluz, @kbond  I've updated my PR to convert all partial templates into anonymous Twig components!

templates/blog/index.html.twig
```diff
@@ -4,7 +4,7 @@
 
 {% block main %}
     {% for post in paginator.results %}
-        {{ include('blog/_post.html.twig') }}
+        <twig:blog:post :post="post"/>
```
stof, line 7: why using a UX-only way of including a Twig template instead of using the native Twig syntax for that, for which we get IDE support and proper documentation when looking at the Twig website ?

even myself being member of the core team, I was not aware of that syntax until 3 minutes ago. Diverging from the native Twig features means that you actually split the community away from the community of Twig, requiring specific Symfony UX knowledge to be able to understand what happens.

I don't recommend that.
And I would even suggest that the Symfony UX reconsiders that feature /cc @weaverryan

weaverryan: I agree. Actually, I have a broader problem. In this case, we have an anonymous component - so just a template. But what if we want to upgrade this to a class-backed component? Right now, it requires explicitly naming the component on the `AsTwigComponent`, which is doable, but not ideal. Also, I upper-case my component names, though there is no clear best-practice on that. It's time to figure this stuff out.

weaverryan: I've got a PR on TwigComponents to fix & settle this: https://github.com/symfony/ux/pull/1140

One key thing is that when if say `<twig:Blog:Post>`, the `Blog:Post` name is a "component name", not a template path. Sure, if it's anonymous, it makes to `templates/components/Blog/Post.html.twig`. But the main idea is that each component has a *name*, and we're using `:` as a separator instead of something like `\` .

For example, if you have a component whose namespace is `App\Twig\Components\Blog\Post`, under the PR, it's name would be `Blog:Post`. I'd love to know if this and that PR help clear things up.

javiereguiluz: I agree with @stof. I'm not sure about showing this syntax in the Symfony Demo app. It's just too new and too "un-Symfony" at the moment. This could change in the future if his syntax becomes popular ... and then we could add it to Symfony Demo.

weaverryan: That's ok for me :). We could revisit near the end of the year, or later. Also, it would be best to introduce the new syntax and twig components in just one or two places instead of everywhere. That would "show it off", but not make it such an overwhelming "we always use twig components" type of experience.

Issue #1344, opened 2022-07-11
API


artyuum:
Hi,

I'm not sure if this has already been asked before as I didn't find anything relevant with the term "api" when I searched in this repo.

This demo is good enough to help the beginners to understand how Symfony works and how to do "things". 

I was just wondering if an API is planned for this application? This would help the beginners who are trying to create more than a simple website: an API.

My suggestion is to create few CRUD routes to interact with the application and show the user how to map & validate the incoming request, why not implement a (basic?) stateless authentication, how to return a consistent reponse across the different routes, etc.

What do you think?

PR #1435, opened 2023-10-15
Use `#[MapRequestPayload]` for admin blog controller

### LINKED ISSUE #1434: Use DTOs to map requests (using MapRequestPayload & MapQueryString)
The demo is a good place to show the best practices. Now that we have the new attributes to map requests to DTOs (`MapRequestPayload` & `MapQueryString`) I think it would be great to use them in the demo. It would also help to see some potential issues we can hit using those (there are some open issues related to that).

If you think we should not, feel free to close this issue.

BafS:
Fix #1434

Unfortunately there is no Rest endpoints in the demo, I think it would have been more interesting to use it then.

### DIFF
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -11,6 +11,7 @@
 
 namespace App\Controller\Admin;
 
+use App\Dto\DeletePostPayload;
 use App\Entity\Post;
 use App\Entity\User;
 use App\Form\PostType;
@@ -22,6 +23,7 @@ use Symfony\Component\Form\Extension\Core\Type\SubmitType;
 use Symfony\Component\Form\SubmitButton;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Attribute\MapRequestPayload;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Security\Http\Attribute\CurrentUser;
 use Symfony\Component\Security\Http\Attribute\IsGranted;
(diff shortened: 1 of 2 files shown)


src/Controller/Admin/BlogController.php
```diff
@@ -159,12 +161,9 @@ public function edit(Request $request, Post $post, EntityManagerInterface $entit
      */
     #[Route('/{id}/delete', name: 'admin_post_delete', methods: ['POST'])]
     #[IsGranted('delete', subject: 'post')]
-    public function delete(Request $request, Post $post, EntityManagerInterface $entityManager): Response
+    public function delete(#[MapRequestPayload] DeletePostPayload $deletePostPayload, Post $post, EntityManagerInterface $entityManager): Response
```
OskarStark, line 164:
```suggestion
    public function delete(
    #[MapRequestPayload] DeletePostPayload $deletePostPayload,
    Post $post,
    EntityManagerInterface $entityManager,
    ): Response
```

BafS:
@94noni I didn't do it because of https://github.com/symfony/demo/issues/1434#issuecomment-1761098715 "Maybe we could do it only in one controller". Unfortunately there is not much to do in any controller. The best would be to introduce a restful api in the demo.

94noni:
oki got it
for the api part, there were already a [discuss on that topic](https://github.com/symfony/demo/issues/1344)

src/Dto/DeletePostPayload.php
```diff
@@ -0,0 +1,23 @@
...
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Dto;
```
94noni, line 12: I personally am used to `\Model` namespace
`\Dto`, for a demo, would make sens only if the dev reading know it (what we can imagine it is ok, but not sure :s)
wdyt?

BafS: Model namespace is often used for database models (in Laravel for example), I think having Dto is less confusing

stof: In Doctrine, the database models are in the `Entity` namespace.

javiereguiluz:
Closing as "won't merge" for the reasons explained in https://github.com/symfony/demo/issues/1434#issuecomment-1987942195

Sorry!

PR #1443, opened 2023-10-22
Fix comment stating isSubmitted is optional

smnandre:
The isSubmitted() _must_ be called before

[src/Symfony/Component/Form/Form.php](https://github.com/symfony/symfony/blob/6637b78a4fcba7cef818f2a022e096d503690ac5/src/Symfony/Component/Form/Form.php#L646-L657)
```php

    public function isValid(): bool
    {
        if (!$this->submitted) {
            throw new LogicException('Cannot check if an unsubmitted form is valid. Call Form::isSubmitted() and ensure that it\'s true before calling Form::isValid().');
        }

        if ($this->isDisabled()) {
            return true;
        }

        return 0 === \count($this->getErrors(true));
    }
 ```

### DIFF
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -86,9 +86,8 @@ final class BlogController extends AbstractController
 
         $form->handleRequest($request);
 
-        // the isSubmitted() method is completely optional because the other
-        // isValid() method already checks whether the form is submitted.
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

yceruto (approved):

javiereguiluz:
Another good catch! Thanks Simon 🙌

PR #1441, opened 2023-10-22
Update documentation links

smnandre:
* update doclinks.js (and remove jquery usage)
* update some links in BlogController
* remove references to SensioFrameworkExtraBundle

PR #1445, opened 2023-10-23
Tweak JS code and rebuild assets

javiereguiluz:
A minor tweak in the JS code updated in #1441.

Also, rebuild front-end assets because in this project we embed those. In real apps, the best practice is of course to not do this ... but this Demo project is special and needs to be fully working without running any command or installing anything else.


smnandre:
Oops, sorry 😶‍🌫️  @javiereguiluz

WebMamba (approved):

PR #1449, opened 2023-11-07
[WIP] Upgrading to AssetMapper 6.4

weaverryan:
Hi!

This converts from Encore to AssetMapper 6.4 (and also upgrading to Symfony 6.4 in general). This was the final product of a live stream walking through this process - https://www.youtube.com/watch?v=pyj1hvhcxGc

There are still some TODOs to finish. The biggest is the first, as Sass is quite embedded in there.

- [X] ~~A) convert Sass->CSS (not a requirement for AssetMapper, but Javier & I talked about this). Then remove SassBundle & bootstrap in `composer.json`. Probably many Sass variables should be converted to CSS variables. See #1450 and #1452~~. Much of this was done, Sass kept just for Bootstrap.
- [X] ~~B) In `assets/app.js`, I should just be able to import from the individual `bootstrap` package instead of requiring each part~~ It's fine how it is. When we upgrade, perhaps we could look at it then.
- [X] C) commit `public/assets/` or `assets/vendor/` so the code works after clone with ZERO other commands. Else, the user will need to run `importmap:install`.
- [x] D) update `devcontainer.json`
- [x] E) fix the `admin` entrypoint & layout

It's a busy month! If anyone is interested in trying to fix any of the above points, just let me know here or on Slack. I'd be more than happy to coach someone through them.

Cheers!


templates/base.html.twig
```diff
@@ -15,13 +15,10 @@
             Those two blocks defines frontend entrypoint for CSS and JavaScript assets
             See https://symfony.com/doc/current/frontend.html
         #}
-        {% block stylesheets %}
-            {{ encore_entry_link_tags('app') }}
-        {% endblock %}
+        {% block stylesheets %}{% endblock %}
         {% block javascripts %}
-            {{ encore_entry_script_tags('app') }}
+            {{ importmap('app') }}
```
94noni, line 20: I like the PR, as well as asset mapper that add a real value in sf world (your live was extremely valueable)
I am just challenging the full removal of « webpack encore »
Could it be wise, in the « migration » phase to showcase both ? (Dont know the faisability in demo repo)
To let devs choose their best case scenario / what  is best for their needs
Asset mapper is a new card added in the deck of devs and encore should remains discoverable in the meantime 
Wdyt ?

javiereguiluz: If I understood Ryan well (@weaverryan can confirm) AssetMapper is the only future of Symfony asset handling and Webpack Encore is the past. Sooner than later, Encore will be marked as deprecated and AssetMapper will be the only option.

weaverryan: That's exactly correct. I'm certainly not in a rush to deprecate Encore & I want to keep supporting the many apps using it. But AssetMapper is the future and that's what people should discover.

94noni: oki copy that, thx for clarification both of you 👍🏻

tacman:
> commit public/assets/ or assets/vendor/ so the code works after clone with ZERO other commands. Else, the user will need to run importmap:install.

I think committing those is going to create a worse DX experience.  Developers download the demo and start tweaking it as a way of learning Symfony's best practices.  I've been burned with accidentally having the public/assets and wondering why stimulus or css wasn't working.

```
bash git clone && composer install && assets:install && start the server
```

That's a pretty amazing installation.   And it makes sense -- download the PHP code, download the front-end assets, start the server.  We shouldn't commit the front-end assets (js/css/images) for the same reason we don't commit the /vendor directory.

tacman:
I'm happy to help, maybe with the admin layout?

javiereguiluz:
@tacman I'm sorry but we don't plant to change the current way of working and installing with this demo.

One of its best features is that it doesn't require anything except cloning and `composer install`. No extra config. No extra asset installing or generation. Nothing.

This feature has made the Symfony Demo to be usable in many different scenarios. For example, the PHP source code uses Symfony Demo in heir continuous benchmarks. That's only possible because of how easy is to install and use it.

Thanks for understanding!

tacman:
Brainstorming: could composer install have an auto-script that installs the
assets, where it clears the cache?

tacman:
@RubenKruiswijk Indeed, my applications all have

```json
        "auto-scripts": {
            "cache:clear": "symfony-cmd",
            "assets:install %PUBLIC_DIR%": "symfony-cmd",
            "fos:js-routing:dump --format=js --target=public/js/fos_js_routes.js --callback=\"export default  \"": "symfony-cmd"
        },
```

Any for deployment, I have the same in either Procfile or app.json (to deploy to heroku / dokku)

```
web:  vendor/bin/heroku-php-nginx -C nginx.conf  -F fpm_custom.conf public/
release: bin/console importmap:install && bin/console asset-map:compile
```

```json
    "scripts": {
        "dokku": {
            "predeploy": "bin\/console importmap:install && bin\/console asset-map:compile",
            "postdeploy": ""
        }
    },
```

I still think putting the public/assets in the repo is going to be a pain point for many developers.

stof:
For heroku, I would recommend to do that in the `heroku-postdeploy` composer script (which is run during deployment) instead of doing that in the `release` step (which will run in a dedicated dyno and so won't impact the web dynos if your files are served by your app and not a CDN).

stof:
And the `importmap:install` should even be in `auto-scripts` so that any `composer install` runs it automatically (maybe it could be added in the recipe)

templates/base.html.twig
94noni: for context https://github.com/symfony/webpack-encore/pull/1239

javiereguiluz:
Ryan, thanks for keep working on this 🙌 

I recently upgraded a full Symfony app from Webpack Encore to AssetMapper. I have some comments and questions about AssetMapper and shared them at https://github.com/symfony/symfony/issues/52939

When this PR is finished, I'll look in detail at Ryan's work to add or remove more comments/questions. Thanks!

weaverryan:
This is now ready! Thanks Javier for the other PR's to smooth my path :).

In some future PR's, I think it's time to modernize a few things also (I'm sure you're already aware) - like going to Bootstrap 5, removing jQuery, etc. Some of the complexity of the PR is due to some older things being used, but I'm happy with the result! I don't know the demo super well, so I did my best to check things.

javiereguiluz:
Just asking: to keep the previous behavior where assets are fully compiled and ready to use out-of-the-box, shouldn't we commit the `assets/vendor/` dir to the repo too?

weaverryan:
> Just asking: to keep the previous behavior where assets are fully compiled and ready to use out-of-the-box, shouldn't we commit the assets/vendor/ dir to the repo too?

Very fair question. I added `importmap:install` to the composer scripts. So if you run `composer install`, everything is there. So if you somehow run `composer install` which you must do somehow even right now, then you should have everything you need. But if you experience something different, please tell me!

javiereguiluz:
Ryan, thanks a lot for working so hard on this contribution 🙇 

Let's merge it now so folks can play with it before tagging a new release. Thanks!

PR #1452, opened 2023-11-09
Import the entire Bootstrap styles

javiereguiluz:
Here's another proposal to improve @weaverryan's work on #1449.

We already import 90% of Bootstrap so, why not import 100% of it?


OskarStark (approved):

weaverryan:
I think it makes sense. As a demo, we should show realistic things, but "splitting hairs" to this level reduces readability. So yes, I like it :).

PR #1459, opened 2023-11-23
Simplify logout

javiereguiluz:
I checked the list of Symfony 6.4 features, and this is the only one that I think it fits this application. So, let's use it.

### DIFF
--- a/src/Controller/SecurityController.php
+++ b/src/Controller/SecurityController.php
@@ -59,16 +59,4 @@ final class SecurityController extends AbstractController
             'error' => $helper->getLastAuthenticationError(),
         ]);
     }
-
-    /**
-     * This is the route the user can use to logout.
-     *
-     * But, this will never be executed. Symfony will intercept this first
-     * and handle the logout automatically. See logout in config/packages/security.yaml
-     */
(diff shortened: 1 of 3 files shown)


src/Controller/SecurityController.php
```diff
@@ -59,16 +59,4 @@ public function login(
...
-     * This is the route the user can use to logout.
-     *
-     * But, this will never be executed. Symfony will intercept this first
-     * and handle the logout automatically. See logout in config/packages/security.yaml
-     */
-    #[Route('/logout', name: 'security_logout')]
-    public function logout(): void
-    {
-        throw new \Exception('This should never be reached!');
-    }
```
bocharsky-bw, line 73: Still might be good for newcomers? It explains how the feature works internally

javiereguiluz: The problem is that this "fake" logout route always felt weird for newcomers and created friction. I think it's better to just forget about this. If someone needs to do something advanced related to logout, they'll just find plenty of information about this in the Symfony Docs. Thanks!

bocharsky-bw: Yeah, probably makes sense 👍

PR #1463, opened 2023-11-30
Update the app to Symfony 6.4.0 stable

javiereguiluz:


composer.json
```diff
@@ -21,29 +21,29 @@
...
-        "symfony/translation": "6.4.*",
-        "symfony/twig-bundle": "6.4.*",
+        "symfony/string": "^6.4",
+        "symfony/translation": "^6.4",
+        "symfony/twig-bundle": "^6.4",
         "symfony/ux-live-component": "^2.6",
-        "symfony/validator": "6.4.*",
+        "symfony/validator": "^6.4",
         "symfony/webpack-encore-bundle": "^2.1",
-        "symfony/yaml": "6.4.*",
```
stof, line 46: As this demo wants to make minor version bumps explicit, should we really switch to the `.*` range to a semver range (which is only equivalent in this case because we know that we won't ever release 6.5) ?

javiereguiluz: You are right. Let's revert to using `6.4.*` as before. Thanks.

PR #1466, opened 2023-12-04
Updated Dutch translations

evertharmeling:
As asked on [Slack](https://symfony-devs.slack.com/archives/C8WHX21K7/p1701681153225289)

- Ran `bin/console translation:extract --force nl` command, this caused the major update in the file. But I guess it should be the recommended way to extract the translations.
  - Running the command also added separate `security` and `validators` files. ~~If everybody uses the command above it may be logical to add them and have control over them in the demo application.~~ ~~Otherwise we could explicitly not add them as they seemed to be correctly translated already?~~ For now not added the extra generated files, this could be done later on for every language if we decide to.
- Added missing `label.remember_me` (for `en` and `nl`)

### DIFF
--- a/translations/messages+intl-icu.nl.xlf
+++ b/translations/messages+intl-icu.nl.xlf
@@ -91,6 +91,14 @@
                 <source>title.comment_error</source>
                 <target>Er is een fout opgetreden bij het opslaan van je reactie</target>
             </trans-unit>
+            <trans-unit id="title.edit_user">
+                <source>title.edit_user</source>
+                <target>Wijzig gebruiker</target>
+            </trans-unit>
+            <trans-unit id="title.change_password">
+                <source>title.change_password</source>
+                <target>Wachtwoord wijzigen</target>
+            </trans-unit>
 
             <trans-unit id="action.show">
                 <source>action.show</source>
@@ -169,6 +177,14 @@
                 <source>action.browse_admin</source>
                 <target>Beheerpaneel bekijken</target>
             </trans-unit>
+            <trans-unit id="action.edit_user">
+                <source>action.edit_user</source>
+                <target>Wijzig gebruiker</target>
+            </trans-unit>
+            <trans-unit id="action.change_password">
+                <source>action.change_password</source>
+                <target>Wijzig wachtwoord</target>
+            </trans-unit>
 
             <trans-unit id="label.title">
                 <source>label.title</source>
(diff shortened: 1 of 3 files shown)


javiereguiluz:
Thanks for this. The `validators.nl.xlf` and `security.nl.xlf` files look the same as the default Symfony files, right?

I think we should exclude them, but let's wait for more comments about this. Thanks!

evertharmeling:
Looking at:
- [default validators.nl.xlf](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Validator/Resources/translations/validators.nl.xlf) it does not exactly match the one in this PR.
- [default security.nl.xlf](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Security/Core/Resources/translations/security.nl.xlf) also does not exactly match the one in this PR (it's 1 translation off in count, didn't diff it on identifier)

Not sure how it's exactly generated, but maybe good idea to exclude them idd.

evertharmeling:
Removed the extra generated files so we can move this PR forward, we could always add those files by running the command again ;)

javiereguiluz:
Evert, I'm sorry I didn't merge this one yet. The problem is that I'm not comfortable merging it as it is now.

Why? Because all the other translation files use the traditional format, and this PR proposes to change the format but only for NL locale:

```xml
<!-- BEFORE -->
<trans-unit id="http_error_500.description">
    <source>http_error_500.description</source>
    <target>Er is een interne serverfout opgetreden.</target>
</trans-unit>

<!-- AFTER -->
<trans-unit id="ZYXSW95" resname="http_error_500.description">
    <source>http_error_500.description</source>
    <target>Er is een interne serverfout opgetreden.</target>
</trans-unit>
```

I'm not sure the new format is better. E.g. why have a random `id` value when we already have a unique identifier for each message?

So, what should we do here? Thanks!

evertharmeling:
No need to apologize Javier! As in my opinion the 'symfony/demo' project should follow 'best practices' and 'symfony / industry standards'. Therefor I used the `bin/console translation:extract` command, this resulted in these (big) changes.
So everyone working on the translation files, would generate the same files and this would prevent / decrease future conflicts etc. It also comes with the benefit that files are easier compared if they have the same order / layout, since that was necessary to find the missing translations.

So the fact that the id's are randomised it due to the fact that the `translation:extract`-command exports them like that, which I guess would be best practice in a project Symfony wise? The id's don't have a real significant value do they?

We could do the following things (and happy to do so):
1. Drop the use of `translation:extract`-command and therefor have less changes and update the files manually
2. Continue this way and create separate commits / PR's to ease the transformation (for other locales).

javiereguiluz:
I'm not sure what to do here:

* Keep translation files using the traditional format
* Update all translation files with the new format created by `translation:extract`

Let me ask some folks about what do: @stof and @wouterj Thanks!

javiereguiluz:
We finally merged this, but reverted all the changes in the translation labels of the files:

* Here we merged and updated the main Dutch trnaslation file: https://github.com/symfony/demo/commit/a348540342e6b0e570a1aff2522410ff3a2ee20e
* And here we reverted all the other undesired changes: https://github.com/symfony/demo/commit/c9523e95c1af0d7d23e24e8ce26b3301664e54ac and https://github.com/symfony/demo/commit/920d86dc809f837543cb519d3df5b364a2c36577

Thanks Evert!

PR #1468, opened 2023-12-07
CI: Share composer cache across php versions

rosier:
Composer installs the dependencies from a composer.lock file 

The result doesn't depend on the PHP version that is used in CI


GromNaN (approved):

OskarStark (approved):
I would propose to use [`ramsey/composer-install](https://github.com/ramsey/composer-install)`, as it takes care of the cache internally

javiereguiluz:
Thanks Jan!

Oskar, let's try that in a separate PR. Thanks.

PR #1469, opened 2023-12-08
Delete yarn.lock

shyim:
There should be only one lockfile for JS


javiereguiluz:
Good catch. This is no longer needed since we moved from Yarn to npm. Thanks!

PR #1470, opened 2023-12-10
Remove broken link from the error 500 page

rosier:
Reverts https://github.com/symfony/demo/pull/909


javiereguiluz:
Jan, thanks for fixing this.

I see that this was introduced by someone 😊 in 2018 as a "temporary fix" (#909).

PR #1471, opened 2023-12-11
Add a missing command to build Sass files

javiereguiluz:
After merging #1449 I saw this error when running the app:

image


javiereguiluz:
Asking @weaverryan to see if he can verify that this is indeed needed ... or if the best solution is something different. Thanks!

PR #1472, opened 2023-12-11
Improve routes requirements

HeahDude:


src/Controller/BlogController.php
```diff
@@ -46,7 +47,7 @@ final class BlogController extends AbstractController
      */
     #[Route('/', name: 'blog_index', defaults: ['page' => '1', '_format' => 'html'], methods: ['GET'])]
     #[Route('/rss.xml', name: 'blog_rss', defaults: ['page' => '1', '_format' => 'xml'], methods: ['GET'])]
-    #[Route('/page/{page<[1-9]\d{0,8}>}', name: 'blog_index_paginated', defaults: ['_format' => 'html'], methods: ['GET'])]
+    #[Route('/page/{page}', name: 'blog_index_paginated', defaults: ['_format' => 'html'], requirements: ['page' => Requirement::POSITIVE_INT], methods: ['GET'])]
```
HeahDude, line 50: We may want to keep this one as is with a comment to explain the short syntax.

seb-jean:
I like this 👍

bocharsky-bw (approved):

94noni (approved):

javiereguiluz:
Thank you Jules.

src/Controller/BlogController.php
javiereguiluz: At the end I merged this "as is" ... so we kept the `requirements` here too.

The reason is that adding comments between the PHP attributes to explain this could be a bit confusing. So, let's use `requirements` in all cases. Thanks.

PR #1473, opened 2023-12-11
Fix return type of change password action

HeahDude:


javiereguiluz:
Thanks for fixing this bug Jules.

PR #1477, opened 2023-12-11
Fix comment about mailer config

HeahDude:
The config files have been merged and `when@dev` is now used.


javiereguiluz:
Thank you Jules.

PR #1480, opened 2023-12-15
Add a subtle cross-fade view transition

javiereguiluz:
On symfony.com we recently added some subtle [CSS view transitions](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API).

I propose to add this little change that adds a very subtle cross-fade animation between page loads. For now, it only works on Google Chrome browsers.

### DIFF
--- a/templates/base.html.twig
+++ b/templates/base.html.twig
@@ -8,6 +8,7 @@
     <head>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1"/>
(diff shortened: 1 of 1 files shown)


weaverryan:
And you also need to enable it, I believe, currently to automatically see them (thanks to Javier for pointing this out) - `chrome://flags#view-transition-on-navigation`

(not for Turbo & Turbo 8's auto view transitions, you do not need to enable this flag - just mentioning that here for clarity - it is only the automatic view transitions that requires the flag, it seems)

PR #1481, opened 2023-12-24
Replace namespace Route Annotation with Route Attribute

ker0x:


derrabus (approved):

xabbuh (approved):

javiereguiluz:
Good catch, thanks Romain.

PR #1483, opened 2023-12-27
Use MapEntity

seb-jean:
Hi,

I would like to simplify the controller by adding MapEntity but I encounter the error below:

`Controller "App\Controller\Admin\BlogController::index" requires that you provide a value for the "$authorPosts" argument. Either the argument is nullable and no null value has been provided, no default value has been provided or there is a non-optional argument after this one.`

I use Symfony docs : https://symfony.com/doc/current/doctrine.html#fetch-via-an-expression:
```php
#[Route('/product/{id}/comments')]
public function show(
    Product $product,
    #[MapEntity(expr: 'repository.findBy(["product_id" => id], null, request.query.get("limit", 10)')]
    iterable $comments
): Response {
}
```

Could you help me finalize the PR?

### EDIT
This PR is a reproducer for 
* https://github.com/symfony/symfony/issues/53236

### DIFF
--- a/src/Controller/Admin/BlogController.php
+++ b/src/Controller/Admin/BlogController.php
@@ -14,9 +14,9 @@ namespace App\Controller\Admin;
 use App\Entity\Post;
 use App\Entity\User;
 use App\Form\PostType;
-use App\Repository\PostRepository;
 use App\Security\PostVoter;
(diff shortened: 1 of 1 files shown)


xabbuh:
I guess we can close here now after symfony/symfony#53236

PR #1486, opened 2024-01-22
shortened the function in index.php

okorneliuk:
tiny refactoring of index.php


bocharsky-bw (commented):
I think it should be suggested and approved in the recipe first: https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/7.0/public/index.php

Here we want to match the recipe

okorneliuk:
Will create another PR based on recipe

PR #1491, opened 2024-02-03
More consistent code style

shuraknows:
Multiple-line code blocks is being separated with new line from single-line code blocks.

Also php-cs-fixer suggested fix: functions with default null arguments should accept null value.


javiereguiluz:
I like this. Thanks @shuraknows

PR #1492, opened 2024-02-09
Nepali Language Translations

TsongSteven:
This is a Nepali Translation


javiereguiluz:
@TsongSteven this is really great! A new translation for the app ... and one that uses a non-Latin script, which is always important to test that everything keeps working as expected 🙏 

This is now merged! Thanks a lot! _मुरी मुरी धन्यवाद_

PR #1493, opened 2024-02-11
Update tests.yaml

seb-jean:
Hi,

I updated the workflow to have the same as the one proposed in the blog article: [Upgrading Symfony Websites to AssetMapper
](https://symfony.com/blog/upgrading-symfony-websites-to-assetmapper#changes-in-ci-and-deployment)
What do you think ?


javiereguiluz:
This is what I use in all my apps ... so I'm merging it. But, if someone has ideas to fix/improve this, we can change that in future PRs. 

Thanks @seb-jean.

PR #1495, opened 2024-02-16
Add bangla translation

AminulBD:
This is "Bangla" language translation. 

**Wiki:**
1. https://en.wikipedia.org/wiki/International_Mother_Language_Day
2. https://en.wikipedia.org/wiki/Language_Movement_Day
3. https://en.wikipedia.org/wiki/Bengali_language_movement

PR #1498, opened 2024-02-18
Make `Keep me logged in` as a translatable message

llupa:
I used [Wikipedia's login page for all messages](https://es.wikipedia.org/w/index.php?title=Especial:Entrar&uselang=es). 

Some notes: 

- Wikipedia only has `pt` and this app has `pt_BR`. Generally Wikipedia's login labels are different from the demo, but I cannot really tell the cultural difference of `Keep me logged in`
- I could not find a way to trigger Wikipedia's `sr` for latin script, it always defaulted to cyrilic. I used Google's translate API to get the latin version. (It looks ok, but slavic languages are not my _forte_). 
- The demo app renders the Japanese text correctly in browser, but over GH it is just a square (test: ロ).

### DIFF
--- a/templates/security/login.html.twig
+++ b/templates/security/login.html.twig
@@ -35,7 +35,7 @@
                         <div class="form-group">
                             <label>
                                 <input type="checkbox" name="_remember_me" checked/>
-                                Keep me logged in
+                                {{ 'label.remember_me'|trans }}
                             </label>
                         </div>
                         <input type="hidden" name="_target_path" value="{{ app.request.get('redirect_to') }}"/>
--- a/translations/messages+intl-icu.ar.xlf
+++ b/translations/messages+intl-icu.ar.xlf
@@ -218,6 +218,10 @@
                 <source>label.password</source>
                 <target>باسوورد</target>
             </trans-unit>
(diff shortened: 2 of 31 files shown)


javiereguiluz:
@llupa thanks a lo for taking the time to extract this new translation. I know this takes some time, so we truly appreciate it 🙇 

In #1495 we added a Bengali translation (code `bn`). The "Keep me logged in" string should be translated as `আমাকে প্রবেশকৃত অবস্থায় রাখুন`. If you have some time for this, please add that translation too. Otherwise, I'll do it while merging. Thanks a lot!

javiereguiluz:
Thanks a lot. Yes, the tool we use to merge things squashes everything by default. If it doesn't work in this case, I'll tell you. Thanks.

gmsantos:
> Wikipedia only has pt and this app has pt_BR. Generally Wikipedia's login labels are different from the demo, but I cannot really tell the cultural difference of Keep me logged in

Native pt-br speaker here. Current translation is fine 👍

javiereguiluz:
This is now merged! Thanks a lot @llupa

PR #1501, opened 2024-02-19
get rid of $ so cut and paste from github page works.

tacman:


javiereguiluz:
Yes, this is better. Thanks Tac!

PR #1502, opened 2024-02-20
add vietnamese translation

sungvadan:
Hello,

This is "Vietnam" language translation.


llupa:
@sungvadan maybe you checked out before `label.remember_me` was added. Would you be so kind to add it too?

sungvadan:
> @sungvadan maybe you checked out before `label.remember_me` was added. Would you be so kind to add it too?

Yes of course, I added this one to my PR

PR #1508, opened 2024-03-30
explicitly configure a default value that will change

### LINKED ISSUE #1507: deprecation log message
Since doctrine/doctrine-bundle 2.12: The default value of "doctrine.orm.controller_resolver.auto_mapping" will be changed from `true` to `false`. Explicitly configure `true` to keep existing behaviour

After update, I receive this log message.
In doctrine.yaml this auto_mapping exist, not sue what is wrong.

maybe it should be in this format:
`
doctrine:
   orm:
        controller_resolver:
            auto_mapping: true
`

xabbuh:
fixes #1507


javiereguiluz:
Thank you Christian.

Issue #1505, opened 2024-03-04
Error: Symfony\Polyfill\Intl\Icu\Locale::setDefault() Not Implemented


igor-jean:
I encountered an error while trying to run the Symfony project. The error message states:

"The Symfony\Polyfill\Intl\Icu\Locale::setDefault() is not implemented. Please install the "intl" extension for full localization capabilities."

Installation Procedure Used:

I used option 2 for installation:

bash
Copy code
You can create a new project based on the Symfony Demo project...
composer create-project symfony/symfony-demo my_project

 ...or you can clone the code repository and install its dependencies
git clone https://github.com/symfony/demo.git my_project
cd my_project/
composer install
Server Launch Behavior:

When I launch the server using symfony serve, here's what happens:
![Capture d'écran 2024-03-04 165121](https://github.com/symfony/demo/assets/122306310/283564b3-79ce-4971-96c1-56d78dc7d7f7)

igor-jean:
I just tried again, it doesn't work. I thinking it was the location.I am in France and when I launch the server the URL address appears like this: http://127.0.0.1:8000/fr
So I tried to change the address to http://127.0.0.1:8000/en, and I managed to join the demo.
![Capture d'écran 2024-03-06 092356](https://github.com/symfony/demo/assets/122306310/3f6db6aa-dd7b-4a34-85bf-7cbb5f6d14d0)

derrabus:
For the best experience, you should really install the INTL extension, as the error message suggests.

--

Regarding the issue, I'm unsure how we should tackle it. I basically see two options:

* Document `ext-intl` as a requirement in composer.json of the demo app.
* Send a PR to Symfony to ignore this error in the locale switcher.

stof:
Symfony should ignore this in the LocaleSwitcher, as already done in `\Symfony\Component\HttpFoundation\Request::setLocale`

PR #1510, opened 2024-04-05
require php intl extension

shyim:
When you don't have the intl extension and change the language 💥 .


derrabus:
Duplicate of #1505?

shyim:
It was too late, didn't saw that :)

PR #1513, opened 2024-04-14
Update SourceCodeExtension.php

parijke:
Use property promotion

We use property promotion all over the code base.


javiereguiluz:
Yes, this is better. Thanks Paul!

PR #1517, opened 2024-04-29
Fix small comment hiccup

Nispeon:
Hello, I opened this tiny PR because I noticed this little error in a comment.

Despite the size, I think it's a rather confusing error, because not knowing how locales work myself, I am not sure whether it was `an` or `no` that I should remove to fix the comment. 

Hopefully I got it right 🙏🏽


xabbuh (approved):
LGTM

javiereguiluz:
Good catch, thanks Julien.

PR #1522, opened 2024-06-14
[Streamlining Entity Design] Introducing the DTOs

yceruto:
I've made some updates to our entity designs to match what the community recommends about Forms -> DTOs -> Entities.

So, the `Post` entity is now separate from input/request processing, which is handled by the new `PostDto`. I updated all Post's properties to meet business rules about nullability. For example, since the Post title is required, it’s no longer nullable. This also let me remove the phpstan exceptions about Entity <-> DB mapping mismatches.

I kept things simple on purpose, but we might think about adding required properties in the constructor later (or through semantic constructors) to avoid invalid states.

These changes are only for the `Post` entity so far. If you like them, let me know, and I can update the rest.

I'll add comments to clarify any changes that might not be immediately obvious.

Cheers!


src/Entity/Post.php
```diff
@@ -42,33 +39,28 @@ class Post
...
     #[ORM\Column(type: Types::TEXT)]
-    #[Assert\NotBlank(message: 'post.blank_content')]
-    #[Assert\Length(min: 10, minMessage: 'post.too_short_content')]
-    private ?string $content = null;
+    private string $content;
 
     #[ORM\Column]
     private \DateTimeImmutable $publishedAt;
 
-    #[ORM\ManyToOne(targetEntity: User::class)]
```
yceruto, line 64: the `targetEntity` is inferred from the property type

src/Entity/Post.php
```diff
@@ -42,33 +39,28 @@ class Post
...
+    #[ORM\ManyToOne]
     #[ORM\JoinColumn(nullable: false)]
-    private ?User $author = null;
+    private User $author;
 
     /**
      * @var Collection<int, Comment>
      */
-    #[ORM\OneToMany(targetEntity: Comment::class, mappedBy: 'post', orphanRemoval: true, cascade: ['persist'])]
+    #[ORM\OneToMany(targetEntity: Comment::class, mappedBy: 'post', cascade: ['persist'], orphanRemoval: true)]
```
yceruto, line 63: resorting based on the argument definition

src/Form/Model/PostDto.php
```diff
@@ -0,0 +1,78 @@
...
+namespace App\Form\Model;
+
+use App\Entity\Post;
+use App\Entity\Tag;
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[UniqueEntity(fields: ['slug'], message: 'post.slug_unique', entityClass: Post::class, errorPath: 'title', identifierFieldNames: ['id'])]
```
yceruto, line 21: this is now possible since Symfony 7.1

src/Form/PostType.php
```diff
@@ -52,15 +52,18 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             ->add('title', null, [
                 'attr' => ['autofocus' => true],
                 'label' => 'label.title',
+                'empty_data' => '',
```
yceruto, line 55: [phpstan] this approach allows us to avoid nullable properties in the DTOs (forcing us to check for null before setting its value into a non-nullable argument) 

If you remove the `required` HTML attribute, the submitted form field data will be an empty string `''` instead of `null`, and `NotBlank` validator will do its job.

src/Form/PostType.php
```diff
@@ -52,15 +52,18 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
             ])
             ->add('summary', TextareaType::class, [
                 'help' => 'help.post_summary',
                 'label' => 'label.summary',
+                'empty_data' => '',
             ])
-            ->add('content', null, [
+            ->add('content', TextareaType::class, [
```
yceruto, line 62: this improve the UX by adding more vertical space to work with the content input

src/Entity/Post.php
```diff
@@ -78,12 +70,11 @@ class Post
     #[ORM\ManyToMany(targetEntity: Tag::class, cascade: ['persist'])]
     #[ORM\JoinTable(name: 'symfony_demo_post_tag')]
     #[ORM\OrderBy(['name' => 'ASC'])]
-    #[Assert\Count(max: 4, maxMessage: 'post.too_many_tags')]
     private Collection $tags;
 
-    public function __construct()
+    public function __construct(User $author)
     {
-        $this->publishedAt = new \DateTimeImmutable();
```
garak, line 86: Why have you removed this? So, now the publish date is not set when the entity is constructed.

yceruto: This was initially set up to provide a default value during the Post creation form, but its final value was always set through the setter, because it's currently a user input.

Now the default value is set in the DTO instead, and the `Post::$publishedAt` value continues to be set through its `setPublishedAt()` setter method.

Note that (as before) this proposal does not follow the stricter path to guarantee that the entity’s state is always valid from its construction which I'd like to (waiting for confirmation)

src/Controller/Admin/BlogController.php
```diff
@@ -111,7 +111,6 @@ public function new(
         }
 
         return $this->render('admin/blog/new.html.twig', [
-            'post' => $post,
```
yceruto, line 114: removed as it's useless in the template

setineos:
Any news?

PR #1523, opened 2024-07-13
Grammar fix

dereuromark:
As per IDE suggestion, the of is not needed or rather not good here.


javiereguiluz:
Thanks Mark and sorry it took us so long to merge this.

PR #1529, opened 2024-09-30
Update to Symfony 7.2

javiereguiluz:


stof:
@javiereguiluz annotating the User entity to tell phpstan that the identifier property contains a `non-empty-string` and not just any `string`

nicolas-grekas:
A non empty kernel.secret is required for remember_me, so we cannot empty it, unless we also comment out the remember me feature in security.yaml

stof:
@javiereguiluz aren't you seeing a deprecation at least ?

javiereguiluz:
No. Now I also removed `APP_SECRET` from `.env.test` and I don't see a deprecation when running tests either.

nicolas-grekas:
That's a bug, fixed by https://github.com/symfony/symfony/pull/58419
If we want to have an empty APP_SECRET but still use remember me, we could configure a persistent remember strategy.

nicolas-grekas:
> If we want to have an empty APP_SECRET but still use remember me, we could configure a persistent remember strategy.

Actually that's not enough because ESI/fragments need the secret also.

PR #1531, opened 2024-09-30
Generate APP_SECRET in .env.local when missing

nicolas-grekas:
The demo app needs an APP_SECRET because it uses ESI, fragments and remember-me, which all need a secret.

Instead of hardcoding a secret in `.env`, I propose here to generate one in `.env.local` on `composer i/u` when it's missing.

Inspired by the discussion on https://github.com/symfony/demo/pull/1529#issuecomment-2383545961

### DIFF
--- a/.env
+++ b/.env
@@ -16,7 +16,7 @@
 
 ###> symfony/framework-bundle ###
 APP_ENV=dev
-APP_SECRET=2ca64f8d83b9e89f5f19d672841d6bb8
+APP_SECRET=%generate(secret)%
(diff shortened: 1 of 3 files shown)


javiereguiluz:
This proposal is smart, but I think it's convoluted for the Demo application. Also, it introduces a new "moving part" that could fail.

Is it really wrong to ship this Demo application with a predefined value in `APP_SECRET` as we've done since day one?

nicolas-grekas:
Closing in favor of #1532

PR #1538, opened 2024-11-10
[CI] Adjust composer cache key hash to account for composer.lock removal

rosier:


javiereguiluz:
Let's close this in favor of #1551, which removes the GitHub cache entirely because it's not that useful to us. Thanks!

PR #1540, opened 2024-11-13
Move the twbs/bootstrap dependency from require-dev to require

### LINKED ISSUE #1509: saas:build fails on `--no-dev`
saas:build fails right now when `composer install --no-dev` has been done. it cannot find bootstrap, as it's right now in require-dev 🤔 
https://github.com/symfony/demo/blob/main/composer.json#L66

rosier:
I don't know if the demo is intended to work in production mode. However since the assets are no longer bundled, but generated during `composer install` with `sass:build` i think it makes sense to ensure that bootstrap is always available by moving it to the required dependencies.

Fixes #1509

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,6 +47,7 @@
         "symfony/validator": "^7",
         "symfony/yaml": "^7",
         "symfonycasts/sass-bundle": "^0.7",
+        "twbs/bootstrap": "^5",
         "twig/extra-bundle": "^3.3",
         "twig/intl-extra": "^3.3",
         "twig/markdown-extra": "^3.3"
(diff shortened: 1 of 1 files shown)


rosier:
BTW to make the demo work in production mode I also had to run:

`APP_ENV=prod bin/console asset-map:compile`

javiereguiluz:
Jan, sorry I forgot about this PR. Merged!

About the asset error, in the past (when we used Webpack Encore) we included the compiled web assets in this repository. Maybe we should do the same now with AssetMapper? This app has to work perfectly without running any command or configuring anything. Let's think a bit about this. Thanks.

PR #1543, opened 2024-12-08
Use `app.` prefix for `parameters`

seb-jean:
I changed the `locale` setting to put the `app.` prefix in order to follow best practices.

https://symfony.com/doc/current/best_practices.html#use-short-and-prefixed-parameter-names :

> Consider using app. as the prefix of your [parameters](https://symfony.com/doc/current/configuration.html#configuration-parameters) to avoid collisions with Symfony and third-party bundles/libraries parameters. Then, use just one or two words to describe the purpose of the parameter

### DIFF
--- a/config/packages/translation.yaml
+++ b/config/packages/translation.yaml
@@ -1,12 +1,12 @@
 # See https://symfony.com/doc/current/translation.html
 framework:
-    default_locale: '%locale%'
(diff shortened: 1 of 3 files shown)


javiereguiluz:
Yes, let's follow our own best practices 🙏 

Thanks @seb-jean!

PR #1544, opened 2024-12-08
Remove importmap:install command from devcontainer.json

### LINKED ISSUE #1542: Running in GitHub Codespaces
[GitHub Codespaces](https://github.com/features/codespaces) is a good solution to run and try modifying the application without installing anything the desktop. Part of the support was added by https://github.com/symfony/demo/pull/1369

But it doesn't  work anymore, we get the following message when starting a codespace.
> This codespace is currently running in recovery mode due to a configuration error. Please review the creation logs, update your dev container configuration as needed, and run the "Rebuild Container" command to rectify.

What we need to achieve with codespace:

- [ ] No error on start
- [ ] PHP installed with required extensions
- [ ] Composer packages installed
- [ ] Composer script run
- [ ] Server started (HTTP and HTTPS)
- [ ] Server port exposed
- [ ] Bonus: PHP extensions enabled in VScode
- [ ] Bonus: CI that ensure this doesn't break

maxgrundnig:
If it is executed as part of the updateContentCommand step, the container creation fails.

importmap:install is executed by composer install though (part of the auto-scripts), which executes just fine in Codespaces.

Part of #1542

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,11 +4,10 @@
     "ghcr.io/shyim/devcontainers-features/symfony-cli:0": {},
     "ghcr.io/shyim/devcontainers-features/php:0": {
       "version": "8.2"
-    },
+    }
   },
   "updateContentCommand": {
-    "composer install": ["composer", "install"],
-    "importmap:install": ["symfony", "console", "importmap:install"]
+    "composer install": ["composer", "install"]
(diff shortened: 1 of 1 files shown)


GromNaN (approved):
Great! Thanks a lot.

javiereguiluz:
Thanks for fixing this bug Max.

PR #1545, opened 2024-12-09
Remove bin/console ignore from php-cs-fixer configuration

codedmonkey:
Remove redundant ignored path in .php-cs-fixer.dist.php since the Finder only searches for .php files


javiereguiluz:
Good catch, thanks Tim.

PR #1547, opened 2024-12-11
Upgrade to PHPStan 2

rosier:
This PR:
 - Upgrades to PHPStan 2
 - Adds a separate test job for PHPStan
 - Turns off the doctrine rules
 - Merges the config files phpstan.dist.neon and phpstan.neon.dist

I disabled the doctrine rules because all the errors coming from this were basically ignored permanently anyway.

I keep the config file phpstan.dist.neon from the recipe https://github.com/symfony/recipes-contrib/blob/main/phpstan/phpstan/1.0/phpstan.dist.neon and added the rest of the config into that file.


xabbuh (approved):
LGTM

javiereguiluz:
A nice upgrade and perfectly explained. Yes, disabling Doctrine rules for now seems the best decision. Jan, as always, thanks a lot for your help in this project 🙏

PR #1549, opened 2024-12-12
Upgrade to PHPUnit 11

wouterj:
This PR upgrades to PHPUnit 11, including using it's deprecation handling features. Starting with PHPUnit 11, all deprecations features from the bridge are implemented in PHPUnit.

This PR also stops using the PhpunitBridge wrapper, which is recommended for applications in the documentation for at least the past 2 years.

I've added the configuration to make PHPUnit not report indirect deprecations (those caused by 3rd party code), and to fail on a test that triggers a direct deprecation. I think this is the most sensible default configuration that we should promote. If we agree, I'll create a recipe to do this automatically.

---

This PR has mostly been created as a test to see what we need to fix in recipes/code and add to the documentation about PHPUnit 11.


.github/workflows/tests.yaml
```diff
@@ -8,9 +8,6 @@ on:
 
 env:
     fail-fast: true
-    PHPUNIT_FLAGS: "-v"
```
xabbuh, line 11: The PHPUnit 11 equivalent would be `--debug`, right?

wouterj: No, debug is much more verbose than the previous verbose flag (up to a point where using it by default wouldn't make much sense). Verbose is removed without replacement (https://github.com/sebastianbergmann/phpunit/issues/5647#issuecomment-1890483283)

xabbuh: ah okay, thanks for the clarification

phpunit.xml.dist
```diff
@@ -4,17 +4,15 @@
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
          backupGlobals="false"
+         failOnDeprecation="true"
```
stof, line 7: if you fail on deprecations, you should also enable the `displayDeprecations` flag. Otherwise, the failure will be impossible to understand.

.github/workflows/tests.yaml
stof: The replacement is actually the set of separate `display-*` flags

.github/workflows/tests.yaml
```diff
@@ -68,11 +65,5 @@ jobs:
...
 
-            - name: "Install PHPUnit"
-              run: vendor/bin/simple-phpunit install
-
-            - name: "PHPUnit version"
-              run: vendor/bin/simple-phpunit --version
-
             - name: "Run tests"
-              run: vendor/bin/simple-phpunit ${{ env.PHPUNIT_FLAGS }}
+              run: bin/phpunit
```
stof, line 69: shouldn't this be `vendor/bin/phpunit` ?

config/packages/csrf.yaml
```diff
@@ -0,0 +1,11 @@
+# Enable stateless CSRF protection for forms and logins/logouts
+framework:
+    form:
+        csrf_protection:
+            token_id: submit
+
+    csrf_protection:
```
stof, line 7: this looks like a separate change

phpunit.xml.dist
stof: based on our existing usage (and the behavior of older versions of PHPUnit), we should probably also fail on warnings and notices. Our debug mode generally breaks for them, so we want the testsuite to fail for them as well.

config/packages/csrf.yaml
wouterj: Yes, this was just to get it working. I'm not really understanding the set-up of this project, because it has a symfony.lock but no composer.lock (although composer.lock is not gitignored). So I couldn't install phpunit/phpunit without also getting these recipe updates.

.github/workflows/tests.yaml
wouterj: The phpunit/phpunit recipe also installs `bin/phpunit`. This is done to make migrating away from our simple-phpunit wrapper easy and without impacting documentation (as both use the same helper script).

I'm not sure why this project used the vendor binary in the CI (as the `bin/phpunit` file is mentioned in the README).

phpunit.xml.dist
wouterj: > if you fail on deprecations, you should also enable the `displayDeprecations` flag. Otherwise, the failure will be impossible to understand.

From a local test, I think this is not needed. `failOn*` automatically enables `display*` (at least, I get not more output with the config value). The `display*` options seem to be useful when you don't want to fail on them, but still see the information.

stof: oh, I missed that https://github.com/sebastianbergmann/phpunit/issues/5856 has been fixed.

.github/workflows/tests.yaml
javiereguiluz: I'm upgrading some Symfony apps to use PHPUnit 11 and I think Christophe is right.

I have the following `bin/phpunit` file in my apps now:

```php
#!/usr/bin/env php
<?php

if (!ini_get('date.timezone')) {
    ini_set('date.timezone', 'UTC');
}

if (is_file(dirname(__DIR__).'/vendor/phpunit/phpunit/phpunit')) {
    if (PHP_VERSION_ID >= 80000) {
        require dirname(__DIR__).'/vendor/phpunit/phpunit/phpunit';
    } else {
        define('PHPUNIT_COMPOSER_INSTALL', dirname(__DIR__).'/vendor/autoload.php');
        require PHPUNIT_COMPOSER_INSTALL;
        PHPUnit\TextUI\Command::main();
    }
} else {
    if (!is_file(dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit.php')) {
        echo "Unable to find the `simple-phpunit.php` script in `vendor/symfony/phpunit-bridge/bin/`.\n";
        exit(1);
    }

    require dirname(__DIR__).'/vendor/symfony/phpunit-bridge/bin/simple-phpunit.php';
}
```

stof: @wouterj the `phpunit/phpunit` recipe does **not** create this `bin/phpunit` file: https://github.com/symfony/recipes/tree/main/phpunit/phpunit/9.6

wouterj:
Rebased.

The test seem to be broken in main though.

rosier:
The lines https://github.com/symfony/demo/blob/main/phpstan.dist.neon#L8-L10 can be removed.

This should fix the Lint / PHPStan test

javiereguiluz:
Thanks Wouter 🙇🙇

Jan, I removed the lines while merging. Thanks.

alexander-schranz:
As we do this in the demo I think we should adjust also the `test-pack`: https://github.com/symfony/test-pack/pull/24

coldic3:
This is no longer valid https://github.com/symfony/demo#tests

The `bin/phpunit` together with the `symfony/phpunit-bridge` package has been removed in this PR.

PR #1551, opened 2025-01-02
Remove the composer cache from the CI workflows

rosier:
I think there is little benefit anymore for this repository to have a composer cache in the CI workflows. With composer v2 the dependencies are downloaded in parallel which is a lot faster than composer v1. This makes the performance gains from the cache small since it also takes time to setup and save or restore the cache.

Also the chance for a cache hit is low:
- This repository has low traffic and the cache expires after a week which reduces the chance for a cache hit. 
- Every branch / PR get it's own cache which again reduces the chance for a cache hit.

So I purpose to remove the composer cache to simplify the workflows.

Replaces: #1538

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -44,17 +44,6 @@ jobs:
                   php-version: ${{ matrix.php-version }}
                   tools: composer:v2
 
-            - name: "Set composer cache directory"
-              id: composer-cache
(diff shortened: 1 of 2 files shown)


nicolas-grekas (approved):

javiereguiluz:
I like this. I recently found out that the GitHub cache is per PR, but we'd need a global cache for `vendor/` dir that it's shared by everyone and that only changes the the `composer.lock` file changes. It's not possible to do that, so let's remove the cache entirely. Thanks!

PR #1555, opened 2025-01-17
Don't store the password in the session after logging in

nicolas-grekas:
Related to https://github.com/symfony/symfony/pull/59106

I'm wondering why we don't have the User object this way: aka why do we need a serialize function? And why don't we erase the password? Yes it's hashed, but removing it is still better to me.

Anything I'm missing?
@chalasr @stof maybe?

### DIFF
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -145,23 +145,6 @@ class User implements UserInterface, PasswordAuthenticatedUserInterface
      */
     public function eraseCredentials(): void
     {
-        // if you had a plainPassword property, you'd nullify it here
-        // $this->plainPassword = null;
-    }
-
-    /**
-     * @return array{int|null, string|null, string|null}
-     */
-    public function __serialize(): array
-    {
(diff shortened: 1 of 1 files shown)


chalasr:
I think https://github.com/symfony/symfony/pull/59106#issuecomment-2598541584 invalidates this

nicolas-grekas:
Nah https://github.com/symfony/symfony/pull/59106#issuecomment-2598578195

stof:
Another valid reason to use `__serialize` is that a real-world project might have additional relations in their User entity (for other reasons that the authentication) and you would not want to have all of them included in the serialized payload (which might even trigger initialization of proxies depending on whether serialization triggers it or no).
And if serialization does not automatically triggers the initialization of proxies, you would need this `__serialize` performing property access that would ensure that Doctrine initializes the proxy.

nicolas-grekas (commented):
This would require https://github.com/symfony/symfony/pull/59539

PR #1556, opened 2025-01-19
feat: MetaStorm integration

xepozz:
MetaStorm integration.
Start integrating MetaStorm: 
- [xepozz/meta-storm-idea-plugin](http://github.com/xepozz/meta-storm-idea-plugin)
- [PHPStorm Marketplace](https://plugins.jetbrains.com/plugin/26121-metastorm?noRedirect=true)

Found a few more additional features that must be integrated to make Symfony more developer friendly. It's not replacement for Symfony PHPStorm Plugin, at least for now.

Let me know if you're interested in further MetaStorm integration. 
Thanks for any feedback. 

Here are a few screenshots what the `.meta-storm.xml` already made with the demo project.

image
image
image
image
image
image
image
image
image

### DIFF
--- a/.meta-storm.xml
+++ b/.meta-storm.xml
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<meta-storm xmlns="meta-storm">
+    <definitions>
+        <classMethod class="\Symfony\Bundle\FrameworkBundle\Controller\AbstractController" method="render" argument="0">
+            <files xpath="$project/templates" extension="" />
+        </classMethod>
+        <classMethod class="\Symfony\Bundle\FrameworkBundle\Controller\AbstractController" method="redirectToRoute" argument="0">
+            <collection name="routes" />
+        </classMethod>
+        <classMethod class="\Symfony\Component\Routing\Generator\UrlGeneratorInterface" method="generate" argument="0">
+            <collection name="routes" />
+        </classMethod>
+        <classMethod class="\Symfony\Component\String\UnicodeString" method="replaceMatches" argument="0">
+            <languageInjection language="RegExp" />
+        </classMethod>
+        <classMethod class="\Symfony\Component\DomCrawler\Crawler" method="filter" argument="0">
+            <languageInjection language="CSS" />
+        </classMethod>
+        <classMethod class="\Twig\Environment" method="render" argument="0">
+            <files xpath="$project/templates" extension="" />
+        </classMethod>
+        <classMethod class="\Doctrine\ORM\EntityRepository" method="resolveMagicCall" argument="0">
+            <methods xpath="$containingClass" />
+        </classMethod>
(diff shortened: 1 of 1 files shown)


xabbuh:
Can you elaborate a bit on the advantage of using your plugin that requires a custom config compared to the Symfony plugin?

xepozz:
Hi @xabbuh 
Thanks for the question. I've added a few points I recon important:

https://github.com/xepozz/meta-storm-idea-plugin?tab=readme-ov-file#advantages
> Advantages
>
> No need dedicated IDE plugin to support your own framework
> Anyone can extend or override plugin functionality
> So if a feature is not ready by the dedicated IDE plugin, MetaStorm is ready to covers with its possibilities
> No need to support legacy codebase in the dedicated IDE plugin
> Written component config might be used with any other projects, like Symfony Console and almost the whole PHP world
> Life save for single developers, any library configuration might improve developer experience with less knowledge
> Totally free

stof:
This explains the benefits for projects that don't have a plugin specific to their framework. But the Symfony plugin exists since years. So what are the benefits for a Symfony project compared to using the Symfony plugin (which does not require any config) ?

xepozz:
image

Seems like the author isn't ready to support the plugin further, but I don't want to talk about it. 

> So what are the benefits for a Symfony project compared to using the Symfony plugin (which does not require any config) ?

MetaStorm is not a replacement for any dedicated software plugins because it serves a general-purpose role.

Many Symfony components are used outside of the Symfony skeleton, demo, or template applications. As a result, not all projects may use the Symfony Plugin, even though they rely heavily on Symfony components.

The main goal of MetaStorm is to provide library developers with a way to integrate their software, making it reusable within different codebases.

MetaStorm can be used alongside the Symfony Plugin or independently. My goal is to improve the UX/DX as much as possible by integrating helpful functions into projects and enabling maintainers to define how it should be properly configured.

MetaStorm works with any configuration files it can locate. For example, I could create the xepozz/metastorm-symfony package with just one .meta-storm.xml file, but that would likely confuse users, wouldn’t it?

Additionally, I addressed the feature request with a few lines of code: https://github.com/Haehnchen/idea-php-symfony2-plugin/issues/2410.

If you disagree with this proposal, I’m happy to close it.

This version is more polished and professional. Let me know if you’d like further refinements!

seb-jean:
I'm not sure this repository is good for showing this project.

stof:
Adding files in the symfony/demo repository benefits only people contributing to the demo project. It does not help at all devs working on their own Symfony projects. To me, this seems less useful than a repository providing a reusable configuration that they can use.

xepozz:
@stof you're right. The first step is to "boost" demo, after than the files might be commited to the specific packages.

xabbuh:
👎 from me because this doesn't provide any improvement for Symfony users

javiereguiluz:
Let's close this without merging because of the reasons explained above.

In any case, we wish you good luck with your JetBrains plugin. Cheers!

PR #1562, opened 2025-02-21
refactor: Use app.locale

seb-jean:
I updated the code to use the new global app variable; `app.locale`.


javiereguiluz:
Another nice one. Thanks @seb-jean

PR #1564, opened 2025-02-21
Use `IsCsrfTokenValid` Attribute

seb-jean:
This will allow the use of `IsCsrfTokenValid` Attribute but we lose the redirection when the csrk token is not valid.
What do you think?


94noni:
not commenting for the PR changes but cant this redirect feature be implemented upstream ?
for ex a code with:

```
#[Route('/entity/{id}')]
public function show(Entity $entity): Response
```

returns a 404, can this attribute propose a way to return a specific response status code ?

javiereguiluz:
I was going to merge this ... but I tested it again and now I'm not sure.

With the new code, when there's a CSRF token exception the user is redirected to the public blog index instead of the blog admin page. It's pretty confusing 😐

seb-jean:
Yes, indeed. We have no control over this behavior.

stof:
There is a discussion about it in https://github.com/symfony/symfony/issues/57343

PR #1568, opened 2025-04-12
Update DefaultControllerTest.php

94noni:
I thinks its worth mentioning it, a web test, kernel test etc


javiereguiluz:
Thanks Antoine!

PR #1569, opened 2025-05-13
Update RedirectToPreferredLocaleSubscriber.php

flovntp:
It needs to be an array for this eventListener to be triggered 
(just get the case where this eventListener is not triggered because of that)


94noni:
are you sure its not because of another event prior to this one that redirects or stoped the flow?
because its legit to have `event => method` as well as what you propose as far as I know

stof:
To me, the description of the change is what causes confusion in the discussion. The actual change being done here is **changing the priority of the listener**. Saying "it must be an array" does not describe what is done (indeed, to use a non-0 priority, it needs to be an array, but you could make it an array without changing the priority, so using an array is not the meaningful change).

and then, what needs to be explained is *why* the priority needs to be so high.

PR #1570, opened 2025-05-20
Fix composer lock not up to date

symfonyaml:
## Problem
When I follow [the instructions to install the demo project](https://github.com/symfony/demo?tab=readme-ov-file#installation) with the `git clone` command, I have the following error
```
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.
- Required package "twbs/bootstrap" is not present in the lock file.
This usually happens when composer files are incorrectly merged or the composer.json file is manually edited.
Read more about correctly resolving merge conflicts https://getcomposer.org/doc/articles/resolving-merge-conflicts.md
and prefer using the "require" command over editing the composer.json file directly
```

## Reproduce
Script taken from README Installation guide
```
git clone https://github.com/symfony/demo.git my_project
cd my_project/
composer install
```
## Solution
I run a `composer update` and pushed the composer lock file

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -44,6 +44,9 @@ jobs:
                   php-version: ${{ matrix.php-version }}
                   tools: composer:v2
 
+            - name: Lint Composer config
+              run: composer validate --no-check-publish --strict
+
             - name: "Install dependencies"
               id: install
               run: composer install --ansi --no-interaction --no-progress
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Thanks @symfonyaml and sorry for this issue.

Your proposed solution is correct but, we just merged #1572 to update the project dependencies to fix a security vulnerability. So, that also fixed (indirectly) the problem reported in this pull request.

That's why we are closing this one without merging it. Thanks.

xabbuh:
I think if @symfonyaml rebases these changes and amends the commit to something like "update dependencies" it's still worth to merge it.

javiereguiluz:
Let's reopen then!

symfonyaml:
> I think if @symfonyaml rebases these changes and amends the commit to something like "update dependencies" it's still worth to merge it.

@xabbuh done in https://github.com/symfony/demo/pull/1573

PR #1571, opened 2025-05-20
run composer validate before installing dependencies

xabbuh:
in #1540 it was forgotten to sync the `composer.lock` file


rosier:
Moving up the `composer validate` check makes sense since the current check is to late to detect the problem described in #1570

stof (approved):

mtarld (approved):

PR #1576, opened 2025-05-26
Upgrade to Symfony 7.3

rosier:
This updates all dependencies and upgrade to Symfony v7.3.0-RC1

Also fixes the following deprecations:
- The "errors.xml" routing configuration file is deprecated, import "errors.php" instead.
- The "profiler.xml" routing configuration file is deprecated, import "profile.php" instead.
- The "xdt.xml" routing configuration file is deprecated, import "xdt.php" instead.
- Not setting the "property_info.with_constructor_extractor" option explicitly is deprecated because its default value will change in version 8.0.
- The "App\Security\PostVoter::voteOnAttribute()" method will require a new "Vote|null $vote" argument in the next major version of its parent class "Symfony\Component\Security\Core\Authorization\Voter\Voter", not defining it is deprecated.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,19 +1,9 @@
-; top-most EditorConfig file
 root = true
 
-; Unix-style newlines
 [*]
 charset = utf-8
-end_of_line = LF
+end_of_line = lf
+indent_size = 4
+indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
-
-[*.{php,html,twig}]
-indent_style = space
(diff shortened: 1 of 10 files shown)


javiereguiluz:
Symfony 7.3 is now available (https://symfony.com/blog/symfony-7-3-0-released) so we can update to the final stable version. Thanks!

javiereguiluz:
Merged! I updated dependencies to the stable version while merging. Thanks!

PR #1577, opened 2025-05-29
Readd the APP_SECRET value

javiereguiluz:
Context:

* The promise of the "Symfony Demo" app since day one is that it works out of the box. You clone the repo or create an app based on it using Composer ... and it just works. You don't have to run any command, change any config or do anything at all. It just works.
* "Symfony Demo" is a very important application also outside the Symfony context. Folks working on PHP source code routinely use it to benchmark changes in PHP.

As explained in #1574, we're not fulfilling that promise right now because of the change made in #1532. So, let's fix that. The change proposed in #1574 is correct ... but it could be not enough. If you run Composer with the `--no-scripts` option, the proposed script won't run and the error will happen again.

So, in this PR I propose to go back to what worked before ... and add a message explaining that folks don't need to do that in their apps; this is just for the Symfony Demo. There are other parts of the app where we say _"we do this because this is a Demo app and bla bla, don't do this in your own app"_. So, this should be fine.

### DIFF
--- a/.env
+++ b/.env
@@ -16,7 +16,10 @@
 
 ###> symfony/framework-bundle ###
 APP_ENV=dev
-APP_SECRET=
+# in your own Symfony applications, you can leave this secret empty,
+# but in this example application, we need it so the application runs out of the box.
+# see https://symfony.com/blog/new-in-symfony-7-2-optional-secret
+APP_SECRET=2ca64f8d83b9e89f5f19d672841d6bb8
 ###< symfony/framework-bundle ###
 
 ###> doctrine/doctrine-bundle ###
--- a/.env
+++ b/.env
@@ -16,10 +16,7 @@
 
 ###> symfony/framework-bundle ###
 APP_ENV=dev
-# in your own Symfony applications, you can leave this secret empty,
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
It should be added to .env.local instead

nicolas-grekas:
My bad, this is added to `.env.dev` when using recipes, so let's follow this lead instead!

javiereguiluz:
If I add it to `.env.dev`, whenever a user sets `APP_ENV` to `prod` in `.env` *(a very common scenario for this app when using it in benchmarks, etc.)* a 500 error is triggered. 😐

Folks, this is not a normal Symfony app. It's a bit special and its main feature is that it must work under all circumstances and for all users (`git clone` or `composer create-project`; using Composer with or without `--no-scripts`; in `dev` and `prod` envs; etc.)

So, is there any alternative to just defining the secret in the main `.env` file?

nicolas-grekas:
This is also a showcase of best practices, that's why we should find a better option.

> using Composer without --no-scripts

I don't think this is valid objection. If you run without composer scripts, that's your issue.

> git clone or composer create-project

That's the issue with #1574 to me: not dealing with git clone.

PR #1585, opened 2025-07-20
Feature/rtl arabic support

mohssineAboutaj:
I added proper RTL support for Arabic and other RTL languages, plus fixed a bunch of issues in the Arabic translations.


javiereguiluz:
@mohssineAboutaj thanks a lot for this contribution!

Note that while merging we removed two of the new proposed Twig extensions (`rtl_class()` and `rtl_dir()`). This is because Twig extensions are mostly used to store logic that is reused in different templates and these extensions were very simple and only used once or twice.

Thanks!

PR #1588, opened 2025-08-04
Add ext-ctype and ext-iconv as required dependency

rosier:
Sync with https://github.com/symfony/skeleton/blob/7.3/composer.json


javiereguiluz:
Thank you Jan.

PR #1590, opened 2025-08-07
[CI] Fixes and tweaks for the demo install test

rosier:
I did some fixes and tweaks for the demo install test.

I did multiple commit to make it easier to review

### DIFF
--- a/.github/workflows/end_to_end_tests.yaml
+++ b/.github/workflows/end_to_end_tests.yaml
@@ -4,8 +4,12 @@ on:
     schedule:
         # run daily at 2:00 AM UTC
         - cron: '0 2 * * *'
-        # allow manual triggering for testing purposes
-        workflow_dispatch:
+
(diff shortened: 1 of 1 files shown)


.github/workflows/end_to_end_tests.yaml
```diff
@@ -40,21 +50,22 @@ jobs:
             - name: Setup PHP
               uses: shivammathur/setup-php@v2
               with:
-                  php-version: '8.4'
-                  extensions: none, ctype, iconv, intl, mbstring, pdo_sqlite, xml
-                  tools: composer
                   coverage: none
+                  extensions: none, ctype, dom, iconv, intl, mbstring, pdo_sqlite, simplexml, tokenizer, xml, xmlwriter
+                  php-version: latest
```
rosier, line 55: Since the above test with `symfony new` also does use `composer create-project` internally I opted for a different PHP version for this job

PR #1595, opened 2025-08-16
Use bootstrap 5 form theme

rosier:
When bootstrap was updated to v5 the form theme was overlooked


config/packages/twig.yaml
```diff
@@ -1,9 +1,8 @@
 twig:
-    default_path: '%kernel.project_dir%/templates'
```
rosier, line 2: I removed this line as it was also remove from the recipe

https://github.com/symfony/recipes/blob/main/symfony/twig-bundle/6.4/config/packages/twig.yaml

PR #1599, opened 2025-08-31
[Bootstrap v5] Fix the active links in the navigation

rosier:
For the navbar the .active class can no longer be applied to .nav-items, it must be applied directly on .nav-links.


javiereguiluz:
Another one! Thanks!

PR #1604, opened 2025-09-03
Not enabling native lazy objects on PHP 8.4+ is deprecated since Doctrine ORM v3.5

rosier:
This PR adds the config `enable_native_lazy_objects: true` to fix the deprecation on PHP 8.4+ and removes the config `enable_lazy_ghost_objects` as it's no longer needed.

With Doctrine ORM v3 installed the DoctrineBundle enables the lazy ghosts by default (https://github.com/doctrine/DoctrineBundle/pull/1722) and in Doctrine ORM v3 the lazy ghosts are enabled unconditionally (https://github.com/doctrine/orm/pull/10969).

The config `enable_native_lazy_objects: true` is only enabled on PHP 8.4+


composer.json
```diff
@@ -11,9 +11,9 @@
         "ext-iconv": "*",
         "ext-pdo_sqlite": "*",
         "doctrine/dbal": "^4.0",
-        "doctrine/doctrine-bundle": "^2.11",
+        "doctrine/doctrine-bundle": "^2.15",
```
xabbuh, line 14: I think we should update to 2.16 because of doctrine/DoctrineBundle#1905

xabbuh (approved):

PR #1606, opened 2025-09-20
[flatpickr] Fix flatpickr calendar buttons and localization

rosier:
Also fixes the alignment of the input fields and the add-ons


stof:
I think the intent was that the button would also trigger the picker (see the `aria-label` for instance), but the implementation was broken.
I think it would make sense to update `assets/js/flatpicker.js` to also use the button as trigger.

PR #1610, opened 2025-10-05
[CI] Remove the redundant step "Build and compile assets"

rosier:
The step "Install dependencies" with `comoser install` also runs the commands `importmap:install` and `sass:build`.

Also I think the tests should be able to run without `asset-map:compile`.


javiereguiluz:
I like this, because less config is less maintenance for us. Thanks Jan!

My guess is that tests will be (slightly) faster after running `asset-map:compile` ... but the difference will be really small in CI tests, so let's merge this.

PR #1611, opened 2025-10-09
[PHP CS Fixer] Update the config

rosier:
This PR simplifies the Finder config and removes the rules that are already part of the Symfony ruleset


javiereguiluz:
This is one of those things that often gets neglected as "unimportant", so it never gets updated. But as you've shown here, it actually matters ... for example, to stay aligned with Symfony's best practices.

Thanks a lot, Jan, for taking the time to update this!

PR #1616, opened 2025-11-04
Update to Symfony 7.4

javiereguiluz:


yceruto:
Should we also upgrade the min php version to 8.4? Like this we can update also the codebase with latest php features

rosier:
> Should we also upgrade the min php version to 8.4? Like this we can update also the codebase with latest php features

I think we should first do a release with Symfony 7.4 and bump the php version for the release with Symfony 8.0.

This way we stay in sync with the required php versions of Symfony. It also gives us a bit more time to update the codebase with latest features

PR #1622, opened 2025-11-28
Use the new Symfony 7.4 features

javiereguiluz:


yceruto:
now that we are using 8.4, what about validating the values using property hook approach?

yceruto:
>now that we are using 8.4, what about validating the values using property hook approach?

Never mind, I thought the 8.0 upgrade had been merged at this point. We can start introducing these new features now, but with PHP 8.4 as the min requirement it will be much easier to implement

PR #1625, opened 2025-12-29
Update the namespace of AppExtensionTest

rosier:
To reproduce run: `composer dump-autoload -o`

Fixes:
```
Class Twig\AppExtensionTest located in ./tests/Twig/AppExtensionTest.php does not comply with psr-4 autoloading standard (rule: App\Tests\ => ./tests). Skipping.
```


javiereguiluz:
Very nice catch 👏 Thanks Jan!

PR #1658, opened 2026-05-27
Update to Symfony 8.1

javiereguiluz:
~Not finished yet. I still need to update the Symfony recipes.~


javiereguiluz:
This is now ready for review. Thanks!

config/reference.php
94noni, line 1: i find that this is better to avoid committing for personal project
moreover for a discovery repo demo, I would not commit it

src/Kernel.php
```diff
@@ -17,4 +17,12 @@
 class Kernel extends BaseKernel
 {
     use MicroKernelTrait;
+
+    /**
+     * @return list<string> An array of allowed values for APP_ENV
+     */
```
94noni, line 25: perhaps add a doc to https://symfony.com/doc/current/configuration.html#configuration-environments ?

templates/base.html.twig
```diff
@@ -21,6 +21,13 @@
         {% block javascripts %}
             {% block importmap %}{{ importmap('app') }}{% endblock %}
         {% endblock %}
+
+        {% set frankenphpHotReload = app.request.server.get('FRANKENPHP_HOT_RELOAD') %}
```
94noni, line 27: seems weird to have this but having the runtime described in the README to be `symfony serve`
I would remove it ^^

---

if kept:

```suggestion
        {% set frankenphp_hot_reload = app.request.server.get('FRANKENPHP_HOT_RELOAD') %}
```

other twig var are named with this style

src/Kernel.php
javiereguiluz: Good idea! But I changed the link to https://symfony.com/doc/current/configuration/micro_kernel_trait.html#restricting-allowed-environments

templates/base.html.twig
javiereguiluz: This variable name comes from the official recipe (https://github.com/symfony/recipes/blob/main/symfony/twig-bundle/6.4/templates/base.html.twig) and we never make any change to the official recipe contents (to not confuse users) so we cannot change it. But, I added a comment because it's probably a bit odd to newcomers. Thanks.

stof: I think we should modify the recipe to make it use a Twig variable name that follows the recommended Twig coding standards

PR #1669, opened 2026-07-02
Fix the E2E tests workflow

### LINKED ISSUE #1662: E2E Test Failure: 2026-06-26
The daily end-to-end test workflow has failed.

This may indicate users are experiencing issues installing the Symfony Demo application.

**Run**: https://github.com/symfony/demo/actions/runs/28214827138

### Failed Jobs
- **Test Composer Create Project**: failed

Please investigate and fix the installation issues as soon as possible.

javiereguiluz:
Fixes #1662

The daily E2E workflow has been failing every day since June 26. This PR fixes the root cause and makes two improvements to the workflow.

**Why the tests were failing**

Only the Test Composer Create Project job failed: the install worked, bin/console about worked, the server started, but the first request to the homepage returned 502 Bad Gateway. The root cause is a PHP-FPM segfault caused by the opcache JIT:

```
php-fpm8.5[5018]: segfault at 7f273e598ffc ip 00005581c66e35a6 ... in php-fpm8.5
```

The conditions that had to combine:

* This job is the only one using php-version: latest (currently PHP 8.5.7); the other jobs pin PHP 8.4 and kept passing.
* setup-php enables the opcache JIT by default (opcache.jit=1235, 256M buffer). Stock PHP builds (official Docker images, sury packages, Homebrew) ship with the JIT disabled, which is why this was not reproducible outside GitHub runners
* PHP 8.5.7's JIT crashes on a hot code path in Symfony 8.1. Failures started exactly when demo v3.1.0 (the Symfony 8.1 upgrade) was released on June 25: the last green run installed v3.0.2/Symfony 8.0 on the same PHP 8.5.7. PHP 8.4's JIT is not affected.

The crashed FPM worker made the Symfony CLI proxy return 502; an immediate retry succeeded (fresh worker), which is how the crash was narrowed down.

**Changes**

* Disable the opcache JIT in the composer job (ini-values: opcache.jit=disable)
* Add a "Show diagnostics on failure" step to each test job: PHP version and key ini values, a retry of the request (to tell first-request-only failures from persistent ones), the application logs and the kernel messages (dmesg, which is what exposed the segfault). Until now the workflow captured nothing on failure, making issues like this one invisible.
* Create at most one E2E failure issue at a time: previously the workflow created one issue per day (see #1663–#1668, all closed manually). Now, if any issue whose title starts with E2E Test Failure is still open, the workflow doesn't create a new issue.

### DIFF
--- a/.github/workflows/end_to_end_tests.yaml
+++ b/.github/workflows/end_to_end_tests.yaml
@@ -42,6 +42,23 @@ jobs:
                   curl --fail --max-time 5 --no-progress-meter --output home.html -v http://127.0.0.1:8000/
                   symfony server:stop
 
+            - name: Show diagnostics on failure
+              if: ${{ failure() }}
+              working-directory: ${{ github.job }}
+              run: |
+                  php -v
+                  php -i | grep -E "memory_limit|opcache.enable|opcache.jit" || true
+                  # retry without --fail to see if the failure is first-request-only
+                  curl -sS --max-time 15 -o /tmp/retry.html -w 'retry: %{http_code} in %{time_total}s\n' http://127.0.0.1:8000/ || true
+                  head -c 300 /tmp/retry.html || true
+                  echo '===== app logs (filtered) ====='
+                  grep -hvE "security.DEBUG|http_client.INFO|cache.INFO" var/log/*.log | tail -n 60 || true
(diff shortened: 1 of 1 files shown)


javiereguiluz:
The possible PHP bug was reported here: https://github.com/php/php-src/issues/22558

PR #1672, opened 2026-07-23
Return a 404 when filtering the blog by a non-existent tag

### LINKED ISSUE #1454: Remove unnecessary SQL queries
https://github.com/symfony/demo/blob/206f0c9aab07cdc9f7437640b12947d35e9d5e72/src/Controller/BlogController.php#L55
Hi,
I think a reference application should not promote unnecessary SQL queries. You fetch a tag object from DB then execute another SQL call based on last call result, why not just passing a tag as a string to the PostRepository instead? Can I offer PR to fix this line to use just one SQL call?

Amoifr:
Fixes #1454.

On the blog index, requesting a tag that does not exist (`/blog/?tag=this-tag-does-not-exist`) silently lists every latest post instead of telling the user the tag is unknown, because `findOneBy()` returns `null` and `findLatest()` then applies no tag filter.

As @GromNaN suggested on the issue, this returns a `404` when a tag is requested but not found, which is both more helpful for users and the expected behavior for SEO. Requesting an existing tag, or no tag at all, is unchanged.

Added two functional tests to `BlogControllerTest`: one for an existing tag (still `200`) and one for an unknown tag (now `404`).

Thanks @mindaugasvcs for the report and @GromNaN for the direction! 🙏

### DIFF
--- a/src/Controller/BlogController.php
+++ b/src/Controller/BlogController.php
@@ -17,7 +17,6 @@ use App\Entity\User;
 use App\Event\CommentCreatedEvent;
 use App\Form\CommentType;
 use App\Repository\PostRepository;
-use App\Repository\TagRepository;
 use Doctrine\ORM\EntityManagerInterface;
 use Symfony\Bridge\Doctrine\Attribute\MapEntity;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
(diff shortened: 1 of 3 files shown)


src/Controller/BlogController.php
```diff
@@ -55,6 +55,12 @@ public function index(Request $request, int $page, string $_format, PostReposito
 
         if ($request->query->has('tag')) {
             $tag = $tags->findOneBy(['name' => $request->query->get('tag')]);
```
GromNaN: The initial ticket suggested removing this unnecessary DB query. The tag slug can be used to find articles directly; then return the 404 when no articles are found.

GromNaN (changes requested):

Amoifr:
Good point, thanks! I reworked the PR accordingly: the Tag lookup is gone, `findLatest()` now filters directly on the tag name (through a dedicated join so posts keep all their tags), and the controller returns a 404 when no articles are found for the requested tag. The existing tests cover both cases and the full suite is green.

src/Controller/BlogController.php
Amoifr: Done, and thanks for the steer, it came out simpler than what I had. `TagRepository` is gone from the controller, `findLatest()` takes the tag name and filters on it directly, and the 404 now comes from an empty result rather than from a missing `Tag` entity. Sorry for the silence here, I pushed the rework the same day but never said so in this thread.

One detail worth flagging since it is not obvious from the diff: the filter goes through a dedicated `filterTag` join rather than reusing the `t` alias from `addSelect()`. Reusing `t` would have hydrated each post with only the tag being filtered on instead of all of its tags.

Amoifr:
Gentle ping on this one, @GromNaN. I reworked it the day you reviewed: the `Tag` lookup is gone, `findLatest()` filters on the tag name through a dedicated join so posts keep all their tags, and the controller returns a 404 when a tag matches nothing. The seven checks are green.

The review still shows as "changes requested", which I think is only because nobody has looked since. No rush, just making sure it is not waiting on me.

